PR #345, opened 2026-02-09
fix: preserve model settings and Ollama endpoint across provider switches

safvanatzack:
## Description

Fixes multiple issues with model configuration persistence when switching between LLM providers:

1. **Ollama endpoint was hardcoded to `localhost:11434`** — Replaced the raw `fetch()` call with the Tauri `get_ollama_models` command, which uses the user's saved endpoint from config instead of a hardcoded URL.

2. **Ollama endpoint was cleared when saving with a non-Ollama provider** — The save handler now preserves `ollamaEndpoint` when the active provider is not Ollama, instead of setting it to `null`.

3. **Auto-select effect overrode saved model choice** — Removed the `useEffect` that unconditionally selected the first Ollama model whenever the model list loaded, which was overwriting the user's saved selection from the database.

4. **Model selection lost when switching providers** — Added a `providerModelMap` localStorage cache that remembers the last selected model per provider. When switching back to a provider, the previous model choice is restored. The cache is seeded from the database on initial config load.

## Issues:
- Addressed #346 

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [ ] All tests pass

## Documentation
- [ ] Documentation updated
- [ ] No documentation needed

## Checklist
- [ ] Code follows project style
- [ ] Self-reviewed the code
- [ ] No merge conflicts

### Files Changed
- `frontend/src/contexts/ConfigContext.tsx` — Replaced hardcoded Ollama fetch, removed auto-select override, seeds localStorage cache from DB
- `frontend/src/components/ModelSettingsModal.tsx` — Preserves endpoint on save, caches/restores model per provider on switch and on async model list load

### DIFF
--- a/frontend/src/components/ModelSettingsModal.tsx
+++ b/frontend/src/components/ModelSettingsModal.tsx
@@ -596,6 +596,22 @@ export function ModelSettingsModal({
     }
   }, [modelConfig.provider, apiKey]);
 
+  // Restore cached model when async model lists become available
+  useEffect(() => {
+    const providerModels = modelOptions[modelConfig.provider];
+    if (!providerModels || providerModels.length === 0) return;
+
+    // If current model is already valid, nothing to do
+    if (modelConfig.model && providerModels.includes(modelConfig.model)) return;
+
+    // Try to restore from localStorage cache
+    const map = JSON.parse(localStorage.getItem('providerModelMap') || '{}');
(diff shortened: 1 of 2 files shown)


sujithatzackriya:
Reviewed

PR #360, opened 2026-02-19
docs: add Linux prerequisites and troubleshooting to building_in_linux.md

timotif:
## Description

Fixes gaps in the Linux build guide discovered while following it on a clean Ubuntu system with an NVIDIA GPU and CUDA installed via apt.

## Related Issue

N/A — discovered during manual testing of the build guide.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other

## Changes

- **Step 1 dependencies**: expanded the apt/dnf/pacman commands to include all Tauri system dev libraries (`libglib2.0-dev`, `libgtk-3-dev`, `libwebkit2gtk-4.1-dev`, etc.). The original commands only listed `build-essential cmake git`, which caused a `glib-2.0 not found` error at compile time. Runtime packages alone are not sufficient — the `-dev` variants with headers and `.pc` files are required.

- **Step 1a (pnpm/nvm)**: added a new step explaining that `corepack enable pnpm` is required when managing Node.js via nvm, since pnpm is not automatically on PATH in that setup.

- **Step 1a (CUDA linker path)**: added a step under the NVIDIA CUDA section explaining that Ubuntu's apt CUDA packages install static libraries to `/usr/lib/x86_64-linux-gnu/` rather than the `/usr/local/cuda/lib64/` layout the build system expects. Without this, the build fails with `could not find native static library 'cudart_static'`.

- **Troubleshooting**: added entries for `cudart_static` not found, `glib-2.0` not found, and `pnpm` not found.

## Testing

- [x] Manual testing performed — followed the updated guide from a clean Ubuntu system and confirmed a successful build producing both `.deb` and `.AppImage` bundles.
- [ ] Unit tests added/updated — N/A (documentation only)
- [ ] All tests pass — N/A

## Documentation

- [x] Documentation updated

## Checklist

- [x] Self-reviewed the code
- [ ] Added comments for complex code — N/A
- [ ] Updated README if needed — README already links to this guide

### DIFF
--- a/docs/building_in_linux.md
+++ b/docs/building_in_linux.md
@@ -13,13 +13,41 @@ If you're new to building on Linux, start here. These simple commands work for m
 ```bash
 # Ubuntu/Debian
 sudo apt update
-sudo apt install build-essential cmake git
+sudo apt install build-essential cmake git \
+  libglib2.0-dev libgtk-3-dev libwebkit2gtk-4.1-dev \
+  libsoup-3.0-dev libjavascriptcoregtk-4.1-dev \
+  librsvg2-dev libssl-dev libayatana-appindicator3-dev \
+  patchelf pkg-config
 
 # Fedora/RHEL
-sudo dnf install gcc-c++ cmake git
+sudo dnf install gcc-c++ cmake git \
+  glib2-devel gtk3-devel webkit2gtk4.1-devel \
+  libsoup3-devel javascriptcoregtk4.1-devel \
+  librsvg2-devel openssl-devel libappindicator-gtk3-devel \
+  patchelf pkg-config
 
 # Arch Linux
-sudo pacman -S base-devel cmake git
+sudo pacman -S base-devel cmake git \
+  glib2 gtk3 webkit2gtk-4.1 \
+  libsoup3 librsvg openssl libappindicator-gtk3 \
+  patchelf pkg-config
+```
+
+> **Why so many libraries?** Tauri requires GTK and WebKit to render the app window. The packages above are the *development* versions (headers + `.pc` files) needed at compile time — the runtime-only packages are not sufficient.
+
+### 1a. Enable pnpm (if using nvm)
+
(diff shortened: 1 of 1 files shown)


passabilities:
@shuibui — thanks for putting this together. I hit nearly the exact same gaps while following the guide on a clean Ubuntu 24.04 box with apt-installed CUDA, and arrived at essentially the same fixes (your `.cargo/config.toml` approach for the CUDA linker path is cleaner than what I had).

A few additions you might want to fold in before this merges — these are the only things still missing for me to get a clean build from a fresh checkout:

1. **`clang` and `libclang-dev` in the apt list (and equivalents for dnf/pacman).** `llama-cpp-sys` (transitively, `whisper-rs`) uses `bindgen` to parse C headers, and without `libclang-dev` it fails with `fatal error: 'stdbool.h' file not found`. Worth a matching troubleshooting entry too:

   > ### \"fatal error: 'stdbool.h' file not found\" (bindgen)
   > - **Cause:** `clang` / `libclang-dev` is not installed. `bindgen` (used by `llama-cpp-sys`) needs them to parse C headers.
   > - **Fix:** `sudo apt install clang libclang-dev` (or your distro's equivalent).

2. **`libasound2-dev` in the apt list (and `alsa-lib-devel` / `alsa-lib`).** `cpal` uses the ALSA backend on Linux for audio capture; without the dev headers the audio crate fails to compile.

3. **Build output path correction.** The existing snippet shows `src-tauri/target/release/bundle/appimage/Meetily_<version>_amd64.AppImage`, but Tauri actually writes to the workspace `target/` at the repo root with the lowercase product name:
   ```
   target/release/bundle/appimage/meetily_<version>_amd64.AppImage
   target/release/bundle/deb/meetily_<version>_amd64.deb
   ```
   (This is the same casing fix as #458 / #464.)

I have these locally as a superset commit on a branch; happy to either push a follow-up PR with just these deltas after this lands, or open one tomorrow as backup if you'd prefer not to touch this PR again. Let me know what's easiest for you.

PR #397, opened 2026-03-10
fix: disable usage analytics by default

wdbetts:
## Summary
- Analytics is now **opt-in** rather than opt-out for new users
- New installs start with the analytics toggle **off**
- Existing users who previously had analytics enabled are unaffected (their stored preference is preserved)

## Changes
- `AnalyticsProvider.tsx`: Default context, state, and store values changed from `true` to `false`
- `AnalyticsConsentSwitch.tsx`: Store default changed from `true` to `false`

## Test plan
- [ ] Fresh install: verify analytics toggle is off by default
- [ ] Existing install with analytics enabled: verify it remains enabled
- [ ] Toggle analytics on, restart app: verify it stays on

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ target/
 
 frontend/src-tauri/binaries/
 Cargo.lock
+backend/.tool-versions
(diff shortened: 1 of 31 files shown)


wdbetts:
Opened against wrong repo, re-creating against fork

PR #430, opened 2026-04-15
feat: add Apple Speech Recognition as transcription provider

ichbinderbar:
## Summary

Adds macOS-native `SFSpeechRecognizer` as a third local transcription provider alongside Whisper and Parakeet.

- Uses Apple's on-device speech recognition, optimized for Apple Silicon
- Leverages the existing `TranscriptionProvider` trait — no changes to `worker.rs` or existing providers
- All new code gated behind `#[cfg(target_os = "macos")]` — zero impact on Windows/Linux builds
- No database migration needed — provider is stored as a string in existing `transcript_settings` table

## Changes

**New files (4):**
- `apple_speech_engine/engine.rs` — Wraps `SFSpeechRecognizer` via `objc2-speech` crate with `Arc<RwLock<>>` pattern
- `apple_speech_engine/commands.rs` — Global engine state + init command (matches Whisper/Parakeet pattern)
- `apple_speech_engine/mod.rs` — Module with platform-conditional exports
- `audio/transcription/apple_speech_provider.rs` — `TranscriptionProvider` trait implementation

**Modified files (7):**
- `Cargo.toml` — Added `objc2-speech`, `objc2-avf-audio`, `objc2`, `objc2-foundation` (macOS only)
- `lib.rs` — Module declaration (+2 lines)
- `audio/transcription/mod.rs` — Export (+4 lines)
- `audio/transcription/engine.rs` — Match arms for validation + initialization (+47 lines)
- `database/repositories/setting.rs` — Match arms for API key handling (+2 lines)
- `config.rs` — Default model constant (+4 lines)
- `TranscriptSettings.tsx` — UI dropdown option + type update (+11 lines)

**Untouched:**
- `whisper_provider.rs`, `parakeet_provider.rs`, `provider.rs`, `worker.rs` — no changes

## How it works

1. User selects "🍎 Apple Speech (macOS Native)" in Transcript Settings
2. Engine requests speech recognition authorization on first use
3. Audio chunks (16kHz mono f32) are converted to `AVAudioPCMBuffer` and fed to `SFSpeechAudioBufferRecognitionRequest`
4. Recognition runs on-device when supported (`supportsOnDeviceRecognition`)
5. Results include confidence scores averaged across segments

## Test plan

- [ ] Build on macOS (Apple Silicon + Intel)
- [ ] Select Apple Speech from provider dropdown
- [ ] Record meeting and verify transcription appears
- [ ] Verify authorization prompt appears on first use
- [ ] Switch between providers (Parakeet → Apple Speech → Whisper) without crashes
- [ ] Verify Windows/Linux builds are unaffected (`#[cfg]` gating)

### DIFF
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -170,22 +170,25 @@ futures-channel = "0.3.31"
 
 # Apple Speech Recognition (SFSpeechRecognizer) for native macOS transcription
 objc2 = "0.6"
-objc2-foundation = { version = "0.3", features = ["NSString", "NSLocale", "NSError", "block2"] }
+block2 = "0.6"
+objc2-foundation = { version = "0.3", features = ["NSString", "NSLocale", "NSError", "NSArray", "alloc", "block2"] }
 objc2-speech = { version = "0.3", features = [
     "SFSpeechRecognizer",
     "SFSpeechRecognitionRequest",
     "SFSpeechRecognitionResult",
     "SFSpeechRecognitionTask",
-    "SFSpeechAudioBufferRecognitionRequest",
     "SFTranscription",
+    "SFTranscriptionSegment",
+    "alloc",
     "block2",
+    "objc2-avf-audio",
 ] }
 objc2-avf-audio = { version = "0.3", features = [
     "AVAudioBuffer",
-    "AVAudioPCMBuffer",
     "AVAudioFormat",
-    "AVAudioCommonFormat",
+    "AVAudioTypes",
     "AVAudioChannelLayout",
+    "alloc",
 ] }
 
 # PERFORMANCE: Enable Metal GPU + CoreML acceleration automatically on macOS
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -168,6 +168,26 @@ cidre = { git = "https://github.com/yury/cidre", rev = "a9587fa", features = ["a
 dasp = "0.11.0"
 futures-channel = "0.3.31"
 
+# Apple Speech Recognition (SFSpeechRecognizer) for native macOS transcription
+objc2 = "0.6"
+objc2-foundation = { version = "0.3", features = ["NSString", "NSLocale", "NSError", "block2"] }
+objc2-speech = { version = "0.3", features = [
+    "SFSpeechRecognizer",
+    "SFSpeechRecognitionRequest",
+    "SFSpeechRecognitionResult",
+    "SFSpeechRecognitionTask",
+    "SFSpeechAudioBufferRecognitionRequest",
+    "SFTranscription",
+    "block2",
+] }
+objc2-avf-audio = { version = "0.3", features = [
+    "AVAudioBuffer",
+    "AVAudioPCMBuffer",
+    "AVAudioFormat",
+    "AVAudioCommonFormat",
+    "AVAudioChannelLayout",
+] }
+
 # PERFORMANCE: Enable Metal GPU + CoreML acceleration automatically on macOS
 # CoreML provides additional acceleration for Apple Silicon chips
 whisper-rs = { version = "0.13.2", features = ["raw-api", "metal", "coreml"] }
--- a/frontend/src-tauri/src/apple_speech_engine/commands.rs
+++ b/frontend/src-tauri/src/apple_speech_engine/commands.rs
@@ -1,0 +1,79 @@
+//! Tauri commands for Apple Speech engine initialization and status.
+
+use log::{info, error};
+use std::sync::{Arc, Mutex};
+
+/// Stub engine for non-macOS platforms.
+#[cfg(not(target_os = "macos"))]
+pub struct AppleSpeechEngine;
+
+#[cfg(not(target_os = "macos"))]
+impl AppleSpeechEngine {
+    pub async fn is_model_loaded(&self) -> bool {
+        false
+    }
+    pub async fn get_current_model(&self) -> Option<String> {
+        None
+    }
+    pub async fn transcribe_audio(&self, _audio: Vec<f32>) -> anyhow::Result<(String, Option<f32>, bool)> {
+        Err(anyhow::anyhow!("Apple Speech is only available on macOS"))
+    }
+}
+
+/// Global Apple Speech engine instance (matches WHISPER_ENGINE / PARAKEET_ENGINE pattern).
+pub static APPLE_SPEECH_ENGINE: Mutex<Option<Arc<super::AppleSpeechEngine>>> = Mutex::new(None);
+
+/// Initialize the Apple Speech engine.
+/// Checks availability and requests authorization if needed.
+#[cfg(target_os = "macos")]
+pub async fn apple_speech_init() -> Result<(), String> {
+    // Check if already initialized
+    {
+        let guard = APPLE_SPEECH_ENGINE.lock().unwrap();
+        if let Some(ref engine) = *guard {
+            if futures_util::FutureExt::now_or_never(engine.is_model_loaded()).unwrap_or(false) {
+                info!("🍎 Apple Speech engine already initialized");
+                return Ok(());
+            }
+        }
+    }
+
+    // Request authorization
+    super::AppleSpeechEngine::request_authorization()
+        .await
+        .map_err(|e| format!("Authorization failed: {}", e))?;
+
+    // Create engine
+    let engine = super::AppleSpeechEngine::new()
+        .map_err(|e| format!("Failed to create Apple Speech engine: {}", e))?;
+
+    if !engine.is_model_loaded().await {
+        return Err("Apple Speech recognizer is not available on this device".to_string());
+    }
+
+    let engine = Arc::new(engine);
+    {
+        let mut guard = APPLE_SPEECH_ENGINE.lock().unwrap();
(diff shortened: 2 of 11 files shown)


sujithatzackriya:
This is an excelent PR. We are reviewing this internally. thank you for the contribution @ichbinderbar

sujithatzackriya:
Hey @ichbinderbar — thanks for putting this together. The architectural shape is genuinely good: routing Apple Speech through the existing `TranscriptionProvider` trait without touching `worker.rs` or `pipeline.rs`, mirroring the Whisper/Parakeet singleton convention, and gating everything behind `#[cfg(target_os = "macos")]` is exactly the right approach. On-device `SFSpeechRecognizer` would be a great addition for Apple Silicon users — zero model download, fast cold start, low latency.

I gave it a thorough read and tried to build/run it locally on `aarch64-apple-darwin` (M-series Mac). Sharing findings below in case it helps you iterate. Happy to chat about any of these.

## Critical (blocks merge)

### 1. The PR does not currently compile on Apple Silicon (`cargo build` produces 31 errors)

Reproduced on commit `4a33b2a1` with `tauri 2.9.5`, `tokio 1.48`, `objc2 0.6.3`. All errors are variants of:

```
error[E0277]: `*const UnsafeCell<()>` cannot be sent between threads safely
   --> frontend/src-tauri/src/audio/transcription/apple_speech_provider.rs:23:32
    |
23 | impl TranscriptionProvider for AppleSpeechProvider {
    |                                ^^^^^^^^^^^^^^^^^^^
    = help: within `SFSpeechRecognizer`, the trait `Send` is not implemented...
    = note: required for `Arc<AppleSpeechEngine>` to implement `Sync`
note: required by a bound in `TranscriptionProvider`
    --> frontend/src-tauri/src/audio/transcription/provider.rs:50:41
     |
  50 | pub trait TranscriptionProvider: Send + Sync {
```

…and `future cannot be sent between threads safely` on every `#[tauri::command] async fn start_recording*` plus the `tokio::spawn` in `audio/transcription/worker.rs:49`.

**Root cause.** `objc2::rc::Retained<SFSpeechRecognizer>` and `Retained<SFSpeechRecognitionTask>` are intentionally `!Send + !Sync` in `objc2` because Apple's Speech framework is main-thread bound (see [the framework docs](https://developer.apple.com/documentation/speech/sfspeechrecognizer)). The PR currently:

- wraps the recognizer in `tokio::sync::RwLock<Option<Retained<SFSpeechRecognizer>>>` — `RwLock<T>: Sync` requires `T: Send`,
- stores `Arc<AppleSpeechEngine>` in a `pub static Mutex<…>` — statics require `Sync`,
- awaits across `RcBlock<dyn Fn(...)>` and `Retained<SFSpeechRecognitionTask>` inside `request_authorization` (`engine.rs:82-91`) and `transcribe_audio` (`engine.rs:213-218`) — `tauri::generate_handler!` requires `Future + Send`.

**Suggested fix.** The cleanest pattern is a dedicated **main-thread executor**: spawn one native thread that runs `dispatch_main()` or wraps a `CFRunLoop`, keep all `SFSpeechRecognizer` / task / block objects on that thread, and communicate via `tokio::sync::mpsc::Sender<Cmd>` / `oneshot::Sender<Result>` where `Cmd` and `Result` are `Send` plain data:

```rust
enum SpeechCmd {
    Authorize { reply: oneshot::Sender<bool> },
    Transcribe { samples: Vec<f32>, locale: String, reply: oneshot::Sender<TranscriptResult> },
    // …
}
```

The async Rust side never touches Obj-C objects directly. This matches how `objc2` recommends integrating with async runtimes and avoids the unsoundness of `unsafe impl Send`.

Alternatively, `objc2::MainThreadMarker` + `dispatch::Queue::main().sync(|| ...)` works for simpler call sites but blocks the calling tokio worker on every transcription, which is bad for the parallel chunk pipeline in `worker.rs`.

I'd avoid `unsafe impl Send for AppleSpeechEngine {}` — it compiles, but Apple's docs forbid cross-thread use of `SFSpeechRecognizer`, and you'd see non-deterministic crashes deep in `Speech.framework` under load.

### 2. Missing `NSSpeechRecognitionUsageDescription` in `Info.plist`

Once the build issue is resolved, the first call to `SFSpeechRecognizer::requestAuthorization` will crash with `NSInternalInconsistencyException` because macOS TCC requires a usage description. Add to `frontend/src-tauri/Info.plist`:

```xml
<key>NSSpeechRecognitionUsageDescription</key>
<string>Meetily uses Apple Speech Recognition to transcribe your meetings on-device.</string>
```

You may also want `NSMicrophoneUsageDescription` review if it's not already present in a form that mentions Speech Recognition.

## High

### 3. Hardcoded 30 s timeout in `transcribe_audio` vs VAD-segmented chunks

`engine.rs:218`:
```rust
match tokio::time::timeout(Duration::from_secs(30), rx).await { … }
```

The pipeline emits VAD chunks of roughly 1-10 seconds. A 30 s ceiling means a stalled SF callback freezes that worker for up to 30 s before yielding, and with `NUM_WORKERS = 1` (default) that visibly halts live transcription. Consider a 5-8 s ceiling that matches the chunk-duration envelope, or make it proportional to chunk length (e.g., `chunk_secs * 3 + 2`).

### 4. `now_or_never` on `tokio::sync::RwLock::read()` as an "already initialized" fast-path

`commands.rs:35`:
```rust
if futures_util::FutureExt::now_or_never(engine.is_model_loaded()).unwrap_or(false) {
    info!("🍎 Apple Speech engine already initialized");
    return Ok(());
}
```

Tokio's `RwLock::read()` isn't guaranteed to complete in a single poll even uncontended. When it returns `None`, the function falls through and rebuilds the engine + re-runs `request_authorization`. macOS dedupes the auth dialog so the user doesn't see anything, but each invocation allocates a new `Retained<SFSpeechRecognizer>` and replaces the singleton — wasteful and leaks the old recognizer until drop.

A simpler pattern: the `Option` in `APPLE_SPEECH_ENGINE: Mutex<Option<Arc<…>>>` already tells you whether the engine exists. This matches `parakeet_engine/commands.rs`:

```rust
{
    let guard = APPLE_SPEECH_ENGINE.lock().unwrap();
    if guard.is_some() {
        info!("🍎 Apple Speech engine already initialized");
        return Ok(());
    }
}
```

### 5. `tx.take()` called before checking `is_final` in the result handler

`engine.rs` recognition handler: the `oneshot::Sender` is consumed (`tx.take()`) before the `if result.isFinal()` branch. With `setShouldReportPartialResults(false)` the recognizer is supposed to deliver only one final, so in practice this rarely fires — but any spec deviation (or a future toggle to enable partials) sends the first partial through the channel, drops the sender, and every subsequent delivery hits the `Ok(Err(_))` "channel closed" arm. Safer pattern: only `tx.take()` inside the `is_final == true` branch.

## Medium

### 6. UI panel is empty when Apple Speech is the selected provider

`TranscriptSettings.tsx` adds the dropdown option, but when selected the model-options array is empty and no status / locale / availability indicator renders. Users have no way to confirm the provider initialised, see which locale is in use, or learn that authorization is pending. Even a single read-only "Locale: en-US · Status: ready" row would close the loop.

### 7. `provider` union in `LanguageSelection.tsx` (and any other components that key off it) still reads `'whisper' | 'parakeet'`

Apple Speech selections may fall through to a default branch in language-related UI, with no visible indication that the choice did anything.

### 8. Language hint is silently ignored

`apple_speech_provider.rs:transcribe(…)` ignores the `language` parameter entirely (the locale is fixed at engine init). Parakeet emits a warning when an unsupported language is requested; Apple Speech is silent. If a user picks "Japanese" in the language selector while Apple Speech is selected with `en-US`, they get English transcripts with no warning. Either honour the hint (re-init the recognizer with `with_locale`) or surface a toast.

### 9. No explicit `NUM_WORKERS` consideration for Apple Speech

The shared worker pipeline assumes providers can run in parallel. `SFSpeechRecognizer` is single-instance-per-locale on the main thread. If `NUM_WORKERS` ever grows beyond 1 for Apple Speech the recognizer becomes a contention point. Worth documenting in a comment or asserting `NUM_WORKERS == 1` for this provider.

## Low / nits

- `commands.rs:3` — unused `error` import (`use log::{info, error}`)
- `apple_speech_provider.rs:8` — unused `use log::info`
- `config.rs:DEFAULT_APPLE_SPEECH_MODEL` — declared but never read inside the crate
- `TranscriptSettings.tsx` — `isLocalProvider` local variable computed but never consumed
- The `objc2-*` crates in `Cargo.toml` are pinned only to `0.6` / `0.3` minor; consider pinning patch versions to reduce future breakage from objc2 semver drift

## What I'd suggest as a path forward

1. Restructure around a main-thread executor so the engine + provider become `Send + Sync` at the boundary (#1).
2. Add the Info.plist key (#2).
3. Tune the timeout (#3), swap the `now_or_never` for an `Option::is_some` check (#4), and gate `tx.take()` on `is_final` (#5).
4. Render a minimal status panel + wire the language hint warning (#6, #7, #8).
5. Polish the nits.

Once #1 lands the rest is mostly mechanical. Really appreciate the work here — Apple Speech as a first-class provider would land well with users on Apple Silicon, and the integration shape you've chosen is the right one. Let me know if you want a second pair of eyes on the main-thread-executor refactor; happy to help.

If you're planning to tackle these yourself, kindly let us know so we can hold the slot for this PR. If not, no worries — just say so and we can plan accordingly.

Issue #233, opened 2025-11-02
[FEATURE] Multi-Language Support for Summaries


hubfc:
## Feature Description
As a user I would like to be abl to set the language for the AI summary (ex. "German").

## Problem Statement
As of now all Summaries created (also when I add further explanations) are made in english although ollama models are also capable of other languages.

## Proposed Solution
Preprompt/or Setting where I can set the language.

## User Story
As a meeting host,
I want to be able to get my meeting summaries in the language spoken in the meeting,
So that I dont have to translate it with another model outside of meetily.

## Acceptance Criteria
- [ ] Summary is written in the language of the meeting or as I have requested it.

## Technical Considerations
No, I think this might work with a preprompt?

## Alternatives Considered
RN I copy the meeting summary to Ollama and translate it there

## Additional Context
Im from Germany and my customers don't speak any english from time to time

## Checklist
- [x] I have searched for similar feature requests
- [x] I have provided all required information
- [x] I have included any relevant screenshots/mockups --> not necessary
- [x] I have described the problem and proposed solution clearly (hope so)

BTW: I love your solution! Very good work!

JackThursby252:
Any news? Imho this should be easy to implement, and the potential customer base for meetily PRO could grow exponentialy over night ;)

sujithatzackriya:
Absolutely. The feature is implemented and is in internal testing. We'll be releasing this within 1 week for pro users first.

<img width="1470" height="923" alt="Image" src="https://github.com/user-attachments/assets/798b7e95-0f6a-49de-9d73-ebd5d6bf9391" />

despinozap:
Hello guys! Any ETA for this amazing feature to be released for everyone 🙏 ? Thanks in advance!

rgerr:
I’ve given up hope and started building my own project for this, including recording, transcribing, meeting summaries and follow-up.

despinozap:
> I’ve given up hope and started building my own project for this, including recording, transcribing, meeting summaries and follow-up.

I ended up solving it by creating a custom template for summary and at the end of every instruction added something like "Your response must be always in [LANG], just avoid the language used to provide the prompt".
Not as elegant as I wish, but works.

sujithatzackriya:
> Hello guys! Any ETA for this amazing feature to be released for everyone 🙏 ? Thanks in advance!

Can be expected to be available in 2 weeks. We were researching Speaker identification ans transcription quality improvement internally hence the delay.

[...]

PR #447, opened 2026-04-22
feat/multi language summary

### LINKED ISSUE #413: [FEATURE] Change summary output language
## Feature Description
Add an option to configure the language in which meeting summaries are generated, independent of the spoken language in the meeting.

## Problem Statement
Currently, summaries are generated in a fixed or auto-detected language with no user control. Users who conduct meetings in one language but require documentation/summaries in another (e.g., English meetings → Polish summaries, or vice versa) have no way to enforce this. This limits usability for multilingual teams and non-English speakers.

## Proposed Solution
Add a language selector in the app settings (and optionally per-meeting) that is passed as an instruction to the LLM prompt used for summarization. The prompt should explicitly instruct the model to output the summary in the selected language regardless of the meeting's spoken language.

Example prompt injection: `"Generate the summary in [language]."` or locale-based: `"Respond in [locale]."` 

A reasonable default would be to auto-detect from system locale or default to English.

## User Story
As a non-English speaker using Meetily for meetings conducted in a foreign language,
I want to set the summary output language independently of the meeting language,
So that I receive summaries in my preferred language without manual translation.

## Acceptance Criteria
- [ ] User can set a preferred summary language in app settings
- [ ] The selected language is applied to all subsequent summaries
- [ ] Optionally: language can be overridden per meeting session
- [ ] The setting persists across app restarts
- [ ] Default behavior (no language set) remains unchanged for existing users

## Technical Considerations
- The language preference should be injected into the LLM summarization prompt
- Implementation is likely a one-line prompt change — low engineering effort
- Language list can be a static dropdown (ISO 639-1 codes or display names) — no need for dynamic detection
- If the app supports multiple LLM backends, ensure the prompt injection works across all of them

## Alternatives Considered
- **Auto-detect from system locale** — could work as a default but doesn't cover cases where preferred summary language differs from OS language
- **Post-processing translation** — adds latency, cost, and a second LLM call; solving it at the prompt level is cleaner

## Additional Context
N/A

## Checklist

p-s-vishnu:
## Description

Enable users to generate summaries in a selected output language independent of transcript language.

**Changes**
- Settings page consists of a summary selection section with an option to pin default language for all summaries.
- Using a two pass translation
  - **Pass 1**: summarise transcript in English using the chosen template.
  - **Pass 2**: translate the finished Markdown into the user's target language via a dedicated LLM call. Skipped for English or when no language is set.
- All summary-language state lives in `localStorage`.
- When no summary "Generate Summary" button else "Regenerate Summary".

**Why two-pass?** 
The smallest default model ship gemma3:1b cannot reliably produce non-English summaries in a single call it summarises well in English and translates finished Markdown well, but not both at once. Single-call prompt approaches worked on qwen2.5:3b+ and cloud providers but regressed on 1B local models. Smoke-tested on gemma3:1b for Hindi and German: body and headings both land in the target language. 

**Note / Question**
1. Might have to default to other similar sized model for better summarisation.
1. *Explicit Temperature setting:* At the moment it would be server default but for a pure translate-preserving-structure task isn't determinism preferred? (temperature 0 or closer), is it skipped on purpose?
1. *Caching the English intermediate.* Every regenerate currently runs both passes from scratch. If the user only changes the target language (not transcript/template/model), pass 1 is wasted work. worth caching, it would improve speed but should that be done now or current latency is acceptable?

## Related Issue
Closes #413

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [ ] All tests pass

## Documentation
- [ ] Documentation updated
- [ ] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

1. Settings page
image

3. Meeting summary page (new language picker besides Generate summary)
image


sujithatzackriya:
Hey @p-s-vishnu — thanks for this contribution! Really solid work here. A few things I want to call out before the change requests:

## What's done well

**The two-pass pipeline was the right call.** Your reasoning in the PR description — that gemma3:1b can't reliably produce non-English summaries in a single call but handles English summarization + translation independently — shows good real-world testing instinct. You didn't just go with the "obvious" single-prompt approach; you tested against the smallest model we ship and found the practical limit. That kind of thinking matters.

**The resolution cascade is well-designed.** Per-meeting → pinned default → transcription language → Auto is exactly the right UX hierarchy. The `AUTO_VALUE = '__auto__'` sentinel to distinguish "user explicitly chose Auto" from "user hasn't chosen anything" is a subtle but important distinction — it means a user can override a global pin on a per-meeting basis, which is the right mental model.

**localStorage over DB was the right pivot.** Language preference is a UI concern, not a data model concern. Keeping it out of the database avoids migration overhead and keeps the feature self-contained on the frontend. Good judgment.

**The `language_name_from_code` whitelist pattern** is exactly right for security — only static strings reach the LLM prompt, never arbitrary user input. Clean design.

**Error rollback in `handleLangChange`** — the `setSummaryLang(previous)` on localStorage failure is a nice touch that most contributors skip.

## Requested Changes

### HIGH — Must fix before merge

**H1: Translation failure discards the completed English summary**

In `processor.rs:442`, the `.await?` on `translate_markdown` propagates the error, and the already-completed English summary is lost. The service layer marks the entire process as failed. If the LLM times out or errors on the translation pass, the user gets nothing — even though the (expensive) summarization pass succeeded.

Fix — catch the error and fall back:
```rust
let final_markdown = match summary_language.and_then(language_name_from_code) {
    Some(name) if name != "English" => {
        match translate_markdown(
            client, provider, model_name, api_key,
            &english_markdown, name, ollama_endpoint,
            custom_openai_endpoint, max_tokens, temperature,
            top_p, app_data_dir, cancellation_token,
        ).await {
            Ok(translated) => translated,
            Err(e) => {
                error!("Translation to {} failed, returning English summary: {}", name, e);
                english_markdown
            }
        }
    }
    _ => english_markdown,
};
```

### MEDIUM — Should fix

**M2: Full summary logged at INFO level**
`service.rs:270` — `info!("final markdown is {}", &final_markdown)` logs the entire meeting summary (could contain sensitive content) to application logs. For a privacy-first app, this should be `info!("Final markdown generated ({} chars)", final_markdown.len());`

**M3: Settings page picker clipped by scrollable parent**
`SummaryLanguageSettings.tsx:80` — the `absolute z-10 mt-2` positioned picker will get clipped by the parent `overflow-y-auto` when the card is near the bottom of the viewport. Using Radix `<Popover>` with Portal rendering (same pattern you used in `SummaryPanel.tsx`) would fix this.

**M4: Language list gap**
`summary-languages.ts` supports 20 languages but Whisper transcription (`constants/languages.ts`) supports 40. Languages like Malayalam, Hebrew, Finnish, Ukrainian, Greek, Czech, etc. are missing from the summary translation list. If a user's transcription language is one of these, the summary silently falls back to English with no feedback. Either expand the list or show a toast when the resolved language isn't supported.

**M5: No tooltip explaining what "Auto" resolves to**
The language button shows "Auto" but gives no indication of what it actually resolves to — is it the pinned default? The transcription language? English? A subtitle under "Auto" in the picker (e.g. "Uses default (French)" or "Matches transcription language") would make the cascade transparent to users.

---

Overall this is a well-thought-out feature with good user-perspective decisions. The H1 fix is the only blocker — everything else can be iterative. Looking forward to the update!

sujithatzackriya (changes requested):
Requesting changes — one HIGH blocker (H1: translation failure discards English summary) needs to be addressed before merge. See detailed comment above for all findings. Great work overall!

sujithatzackriya:
One more comment

image


The default summary language pin - I struggled a bit to find out how to set the primary language. I had to click exactly on the pin to set it. most users might miss it. Maybe make default language selection a bit more intuitive

Overall, this feature is working really well and good work with this UX.

p-s-vishnu:
Hey thanks @sujithatzackriya happy to help and amazing points, as you suggested I have made the following changes

---

Summary of changes:

### Main

**H1 (blocker) - Translation failure no longer discards the English summary**
`processor.rs` - replaced `.await?` on `translate_markdown` with a `match` that catches errors, logs them at error level, and returns the already-completed English summary. Users always get a result even if the translation LLM call fails.

**UX - Pin interaction made more discoverable**
`SummaryLanguageSettings.tsx` - the language pill is now split into two click zones: the pin icon + label together form a single wide button that toggles the pin **_(no need to click exactly on the small icon)_**, and `×` remains a separate button on the right edge to remove the language. Footer hint updated to match: "Click any language to set it as your default."

image



### Extras

**M2 - Meeting content removed from INFO logs**
`service.rs` - replaced `info!("final markdown is {}", &final_markdown)` with `info!("Final markdown generated ({} chars)", final_markdown.len())`. No meeting content written to application logs.

**M3 - Settings picker no longer clipped by scroll containers**
`SummaryLanguageSettings.tsx` - replaced the `absolute z-10 mt-2` div with the Radix `<Popover>` + Portal pattern already used in `SummaryPanel.tsx`. Picker now renders outside the overflow hierarchy regardless of where the settings card sits on the page.

**M4 - Language list gap closed + unsupported language feedback**
`summary-languages.ts` + `processor.rs` - added Czech, Danish, Finnish, Greek, Hebrew, Hungarian, Norwegian, Romanian, Ukrainian. List goes from 20 to 29, covering all Whisper-supported languages that major LLMs (Claude, GPT-4, Ollama qwen2.5+) handle reliably.
`useSummaryGeneration.ts` - when the transcription language is set but still not in the supported list (e.g. Catalan), a toast now fires on generate/regenerate: _"Transcription language 'ca' is not supported for summary translation — summary will be in English."_ No more silent fallback.

**M5 - "Auto" resolution now visible in the picker**
`LanguagePickerPopover.tsx` - added optional `autoSubtitle?: string` prop; rendered as a small grey subtitle below the "Auto" row when provided.
`SummaryPanel.tsx` - computes and passes the subtitle based on the resolution cascade:
- Pinned default set → `"Uses default (French)"`
- Transcription language resolves → `"Matches transcription (Finnish)"`
- Neither → `"Falls back to English"`

Note: the subtitle also correctly skips the pinned default when the user has explicitly picked Auto for a meeting (the `AUTO_VALUE` override case), so it always reflects what the summary will actually use.

---

Please let me know if any changes are needed. Thanks again for your review.

sujithatzackriya:
Thank you Vishnu for the careful work on this one.

The two-pass pipeline is the right call for small local models, and the helper functions you extracted (resolve_cached_english, extract_english_from_result_json, strip_title_if_present) are clean and well covered by unit tests. The response to the first round of review feedback across d40493b, f2c312b, bc55b80, and cec75d9 was thoughtful, and the cache resolver in particular shows good thinking about the failure modes (locale variants, malformed JSON, empty fields).

Follow-up items from the second-round review are captured in docs/PR-447-Fix-Plan.md

safvanatzack:
closes #233

Zaincasper69 (commented):
```
```
```
```
```

```
```
```
```
```
```
```
```
```
```
```
- [ ] ```
```
```
```
```
```
```

```
```
```
```
```
```
```
```
```
```
```
```
```

Zaincasper69 (commented):
image

PR #455, opened 2026-05-06
Fix/review 447 [skip ci]

safvanatzack:
## Description
This PR includes the latest two commits on `fix/review-447`:

- Hardens multilingual summary regeneration by validating cached English summary provenance before reuse, preventing stale translated summaries when transcript/template/prompt/provider/model inputs change.
- Keeps English summary cache metadata safer and preserves the cached title source used during multilingual regeneration.
- Serializes transcription engine startup and batch unload paths to avoid racing live recording startup.
- Improves invalid stored summary language fallback behavior.
- Moves per-meeting summary language override storage from `localStorage` into each meeting folder’s `metadata.json` as top-level `summary_language`.
- Adds Tauri commands to read/save the meeting summary language while preserving existing metadata fields and using atomic metadata writes.

## Related Issue
Addresses PR #447 review feedback. Related to summary language work in #233 / #413.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [ ] All tests pass

Validation performed:
- `cmd /c frontend\build.bat check cpu` passed
- `bun run build` passed
- `git diff --check` passed

Note: direct `cargo test -p meetily summary_language --lib` was blocked by the existing `whisper-rs-sys` CMake/MSBuild Vulkan build path before tests executed.

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [ ] No merge conflicts

N/A

## Additional Notes
Per-meeting summary language is now stored in `metadata.json` as:

```json
{
  "summary_language": "fr"
}
```

When set to Auto, `summary_language` is removed from metadata. Global defaults and recents remain in `localStorage`.


sujithatzackriya:
## Two follow-ups worth folding in

Both are cheap and close real correctness gaps. Happy to land them in this PR or in a quick follow-up — your call.

### M1 — Custom template content not in cache fingerprint

`frontend/src-tauri/src/summary/service.rs:54-67, 87-112`

`SummaryCacheSource` keys on `template_id` but not template content. `loader.rs:95-118` resolves a custom template under the same `template_id` whether the on-disk JSON has been edited or not.

**Repro:** edit a custom template's JSON to change the section list, regenerate the summary in a non-English language — the cached English pass-1 is reused from the OLD template, only translation re-runs. The displayed translated summary reflects the previous structure, not the edited one.

**Suggested fix:** include a stable hash of the rendered template in `SummaryCacheSource` after `get_template` succeeds — e.g., a hash of `template.to_markdown_structure() + template.to_section_instructions()`. ~10 lines.

### M2 — Metadata.json write race (one-line hardening)

`frontend/src-tauri/src/summary/metadata.rs:9,36`

Temp file is a fixed `.metadata.json.tmp` — no per-call suffix. In a bad interleave (T1 reads → T2 reads → T1 writes temp → T2 overwrites temp → T1 renames → T2's rename fails with ENOENT), at least one save is lost. Unlikely with the current picker UI (saves don't fire within ~10 ms of each other), but the fix is one line:

```rust
const METADATA_TEMP_FILE_PREFIX: &str = ".metadata.json.";
let temp_path = folder.join(format!("{METADATA_TEMP_FILE_PREFIX}{}.tmp", uuid::Uuid::new_v4()));
```

### Other notes (LOW, non-blocking)

- **`zh-tw` asymmetry**: Rust preserves `zh-tw` in `metadata.rs:84-99`, but frontend `LANGUAGE_OPTIONS` only has `zh` and `normaliseLanguageCode('zh-tw') → 'zh'`. No code path currently emits `zh-tw`, so latent only — consider either adding Traditional Chinese to the UI or trimming the Rust handling until UI parity exists.
- **Silent fallback on metadata read error**: `useSummaryGeneration.ts:23` and `SummaryPanel.tsx:114-117` log to console only. A user with corrupt metadata silently loses their pinned language with no toast.
- **N1**: `resolveSummaryLanguage` serializes all transcript texts via IPC every Generate click. For long meetings (1–5 MB JSON), worth caching the resolved language in metadata once after recording stops.
- **N2**: `language_detection.rs:38-56` returns `None` on 50/50 ties → silently defaults to English. Consider a toast: "Bilingual transcript detected — pick a language manually."

Cross-verified against `a85e2e3`.

sujithatzackriya:
Thanks @safvanatzack, sat with this one for a while because of the surface area (cache provenance, storage migration, and a lock primitive all in one PR). Overall it's solid. The cache fingerprinting reads well and the metadata-atomic-write pattern is the right call.

Before fixing anything: please re-verify each finding against the actual code yourself first. A few of these are narrow-reachability and a couple are design calls rather than bugs. If you disagree on any, I'll happily go deeper. Comment on the specific items and I'll respond with the trace.

### Quick glance

| ID | Sev | Area | One-line |
|---|---|---|---|
| M1 | Medium | `summary/metadata.rs:56-96` | Concurrent `api_save_meeting_summary_language` + `api_save_meeting_detected_summary_language` clobber each other (RMW race on metadata.json) |
| M2 | Medium | `summary/service.rs:55-68, 336` | `token_threshold` not in `SummaryCacheSource`, so an Ollama context change reuses stale chunked English for translation |
| M3 | Medium | `audio/recording_commands.rs:83-250` | `ENGINE_LIFECYCLE_LOCK` held across `validate_transcription_model_ready`, so a first-run model download blocks batch unload for minutes |
| M4 | Medium | `MeetingDetails/SummaryPanel.tsx:111-135` | useEffect `cancelled` flag only flips on unmount, so user picking a language during load gets clobbered by the late-resolving fetch |
| M5 | Medium | `useSummaryGeneration.ts:621-631` | Regen passes joined `[MM:SS]`-prefixed transcript instead of per-segment array to detector (first-gen uses per-segment) |
| M7 | Medium | `useRecordingStop.ts`, `useImportAudio.ts`, `useTranscriptRecovery.ts` | `applyPinnedSummaryLanguageToMeeting` failures swallowed with `console.warn`, silent loss of pinned language |
| L1 | Low | `useSummaryGeneration.ts` (deletion) | `primaryLanguage` transcription hint and `summaryLanguageDefault` no longer consulted at generation time. Design call, please confirm intentional |
| L2 | Low | `audio/recording_commands.rs:483+` | `stop_recording` doesn't hold lifecycle lock. Verified `unload_model` is idempotent, so only side-effect is benign `"Recording already in progress"` error |
| L3 | Low | `audio/common.rs:219-232` | Lock-serialization test uses wall-clock `timeout(25ms)`, flaky on loaded CI |
| L4 | Low | `LanguagePickerPopover.tsx` | Missing `aria-pressed` / `aria-selected` on per-language buttons (screen-reader gap) |
| L5 | Low | `tests/lib/summary-language-preferences.test.js` | No test for localStorage quota-exceeded silent-catch path |
| L7 | Low | `summary/service.rs:53` | `english_markdown` to `english_cache` rename forces one extra LLM call per pre-existing meeting on first regen (intentional, worth mentioning in release notes) |
| I1 | Info | `audio/retranscription.rs:742` | `obj.remove("detected_summary_language")` on retranscribe is correct, but unrelated to engine-lifecycle commit message. Consider splitting |
| I2 | Info | `summary/language_detection.rs:267` | Dual reliability gate admits reliable-but-low-confidence segments. Behavior is reasonable, just worth a `debug!` log for tuning |

Full doc with reproduction steps, call-path traces, falsification verdicts, and steel-manning per finding: [`docs/PR-455-Code-Review.md`](../blob/fix/review-447/docs/PR-455-Code-Review.md).

### Top three to look at first

1. **M1** is the only "this can lose user data" one, but it needs two writes within ~100ms on the same meeting, which happens naturally at recording-stop when the user is in SummaryPanel. Easy fix: collapse the two commands into one `api_save_meeting_languages(language, detected)` written from a single site. Removes the surface entirely instead of locking around it.

2. **M2** is subtle. `token_threshold` from Ollama isn't just chunk metadata. `processor.rs:278` uses it as the *strategy gate* (single-pass vs. chunked) and `:292` as the *chunk size*. So an Ollama context change between two regenerations produces a different English summary, but the fingerprint won't notice and the cache reuse will translate stale content. One field on `SummaryCacheSource` closes it.

3. **L1** is the one I'd most like you to confirm intentional. Old `resolveSummaryLanguage` had a `primaryLanguage` (transcription setting) fallback and read `summaryLanguageDefault` at generation time. New chain replaces both with whatlang content detection plus applies the global default only at meeting creation via the three new hooks. For meetings recorded before this PR (no `summary_language` in their metadata), the global default never kicks in. They'll detect and fall back to English. Either restore `primaryLanguage` as a final tier when detection returns `None`, or just call it out in the release notes so users know to pick languages manually for older meetings. Your call.

### Things specifically verified as safe (no action needed)

- Tauri command registration plus camelCase / snake_case mapping for all 5 new commands
- `unload_model` idempotency on both Whisper and Parakeet (`Option::take()` semantics)
- Atomic temp-file-then-rename pattern in `metadata.rs` (same-FS, UUID-suffixed)
- Auto sentinel handling: `null` from JS becomes `Option<String>::None` which calls `object.remove(field)`, key absent
- Serde mapping for `MeetingSummaryLanguagePreference.storage` round-trips correctly
- Bun resolves the `@/*` tsconfig path alias natively. New test suite runs (4/4 pass locally)
- Recording-start entry points both acquire the new lock. Both batch callers route through `unload_engine_after_batch`

Take what's useful, push back on what isn't. Happy to dig deeper on anything that doesn't match what you see.

PR #459, opened 2026-05-07
fix: use TDT duration head in Parakeet greedy decoder

lorenzojb:
## Problem

Live transcription with the Parakeet TDT models (`parakeet-tdt-0.6b-v2/v3`)
produces runs of the same word repeated 8–11 times, e.g.

    "if if if if if if if if if if if for example"
    "uh uh uh uh uh uh uh uh uh uh a a a a a a a a a a headshot"

## Root cause

`decode_sequence` in `parakeet_engine/model.rs` splits the joint-network
output into `[vocab_logits | duration_logits]` but never reads the
duration logits. It always advances the time index by 1 frame on
blank/cap, treating the model as plain RNN-T. For a TDT model the
duration head says how many encoder frames to skip after each emission
(bins `[0,1,2,3,4]`); ignoring it means the decoder revisits the same
audio 3–4× and re-emits the token until `MAX_TOKENS_PER_STEP` (10) is
hit — hence the 10-11× repeats.

## Fix

- Argmax the duration logits and advance `t += skip` per the model's
  prediction; force `skip = 1` on blank-with-zero-duration or after the
  per-frame emission cap to guarantee forward progress.
- Lower `MAX_TOKENS_PER_STEP` to 3 (matches NVIDIA NeMo / other
  reference TDT decoders).
- Non-TDT (plain RNN-T) path keeps the original step-by-1 behaviour.

## Result

Re-transcribing a 15-minute recording at the same segment boundaries:

| metric | before | after |
|---|---|---|
| max consecutive word repeat | 11 | 2 |
| segments with repeat ≥ 5 | 10 | 0 |

Decode is also ~2-3× faster since it visits roughly a third of the
encoder frames.


sujithatzackriya:
@lorenzojb  Thank you for the PR. Will review this. @safvanatzack please check this out

safvanatzack:
Manually tested, the intuition is correct and fix is valid @sujithatzackriya

PR #462, opened 2026-05-09
feat(summary): add Qwen 3.5 built-in models [skip ci]

### LINKED ISSUE #395: [FEATURE] Add support for a small new Qwen 3.5 models
## Feature Description
Please consider adding a new small model from the Qwen 3.5 family as an available option.
Qwen’s smaller models have shown very strong capabilities relative to their size, especially in instruction-following, reasoning efficiency, and overall quality per parameter. A compact Qwen 3.5 model could provide users with a better lightweight alternative for everyday use cases where speed and cost matter.

## Problem Statement
Existing lightweight models, such as Gemma, are useful but may not always offer the best balance between performance, responsiveness, and quality with respect to new models.

## Proposed Solution
Add at least one small model from the Qwen 3.5 family to Meetily’s supported model list.

## User Story
As a user who needs a lightweight but capable model,
I want Meetily to support a Qwen 3.5 model,
So I can achieve better performance and quality with a compact model option.

## Acceptance Criteria
- [ ] Small model family from the Qwen 3.5 family is added as a selectable model
- [ ] The model is available wherever other lightweight models are currently offered
- [ ]Users can easily distinguish it as a compact/high-efficiency option

## Additional Context
The main reason for this request is that Qwen 3.5 small models seem exceptionally strong for their class. They could improve the quality of Meetily’s lightweight model offering without requiring a shift to much larger, more expensive models.

This would be particularly valuable for users who prioritize efficiency but still want high-quality outputs.

## Checklist

safvanatzack:
## Description
This PR adds Qwen 3.5 2B and 4B to the built-in summary model registry, including Qwen-specific non-thinking prompt formatting and safer restrained sampling presets.

It also updates built-in summary model downloads to use Hugging Face URLs, changes the default recommendation policy to prefer Qwen 4B except on low-RAM non-macOS systems, and keeps Gemma models available as legacy/manual alternatives.

The onboarding and download progress UI now tracks the selected summary model generically instead of assuming Gemma-specific model names.

## Related Issue
Fixes #395 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [x] Other: dependency upgrade for Qwen 3.5 architecture support

## Testing
- [x] All tests pass

Tested with:
- `cargo test -p meetily qwen35 --lib`
- `cargo test -p meetily recommended_summary_model --lib`
- `cargo test -p meetily summary_engine --lib`
- `cargo check -p meetily`
- `npm run build`

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [ ] No merge conflicts

N/A

## Additional Notes
Gemma models remain supported for existing users and as manual legacy alternatives. The llama-cpp-2 upgrade is included so llama-helper can support the Qwen 3.5 architecture.


sujithatzackriya:
## Review findings — please validate

Two HIGH-severity issues surfaced during review that I'd like you to verify against the code:

### 1. Broken HuggingFace URL for Gemma 1B (HTTP 404)

`frontend/src-tauri/src/summary/summary_engine/models.rs:155-158`

```bash
$ curl -I https://huggingface.co/bartowski/google_gemma-3-1b-it-GGUF/resolve/main/gemma-3-1b-it-Q8_0.gguf
HTTP/2 404
x-error-code: EntryNotFound

$ curl -I https://huggingface.co/bartowski/google_gemma-3-1b-it-GGUF/resolve/main/google_gemma-3-1b-it-Q8_0.gguf
HTTP/2 302
x-linked-size: 1069306624   # 1.07 GB
```

The 4B URL on line 144 already has the `google_` prefix correctly; only the 1B variant is missing it. Users who pick "Gemma 3 1B" from the Built-in AI model list will get `Download failed with status: 404 Not Found`.

**Suggested fix:**
- `models.rs:155` `gguf_file`: `"gemma-3-1b-it-Q8_0.gguf"` → `"google_gemma-3-1b-it-Q8_0.gguf"`
- `models.rs:157` `download_url`: prepend `google_` to the filename portion
- `models.rs:303` test `gemma_models_use_huggingface_urls_and_tight_structured_sampling` — currently this test **asserts the broken URL string**, so it will keep passing as-is. Update it alongside the URL fix.

### 2. Onboarding regression: Parakeet stuck if summary recommendation IPC fails/is slow

`frontend/src/components/onboarding/steps/DownloadProgressStep.tsx:207-213` + `frontend/src/contexts/OnboardingContext.tsx:86`

Pre-PR `selectedSummaryModel` defaulted to `'gemma3:1b'`, so the download `useEffect` ran on mount and started Parakeet (the blocking model) immediately. Post-PR the default is `''` and the new gate `if (!selectedSummaryModel) return;` prevents `startDownloads()` (which is also the Parakeet trigger via `startBackgroundDownloads`) until the recommendation IPC resolves. If `builtin_ai_get_recommended_model` errors or is slow, Parakeet never downloads, "Continue" stays disabled, and the user gets only a `console.error`.

**Suggested fix:** restore an initial default in `OnboardingContext.tsx:86`, matching the `unwrap_or` fallback in `database/commands.rs:194`:

```ts
const [selectedSummaryModel, setSelectedSummaryModel] = useState<string>('qwen3.5:4b');
```

Single-line change; no logic in `DownloadProgressStep` needs to change.

### Other notes (non-blocking)

- **L3**: `onboarding.rs:23` doc comment still says "(gemma3:1b or gemma3:4b)" — refresh to mention qwen3.5 variants.
- **M4 (release notes)**: Gemma sampling preset changed from `temperature=1.0, top_k=64, top_p=0.95` → `tight_structured` (`0.1, 20, 0.88`). Sampling is read fresh from the static registry on every generation, so existing users with Gemma in their DB will silently get the new preset on next summary. Worth flagging in the release notes.
- **L8 nuance**: `llama-cpp-2 = "0.1.139"` is a caret constraint; with no checked-in `Cargo.lock` it resolves to `0.1.146` on fresh builds. `cargo check` passes (4 deprecation warnings, 0 errors).

Cross-verified against the actual code at `d187f1c`. Happy to review again once H1+H2 land.

sujithatzackriya:
Hey, thanks for the work on this. Spent some time going through the changes end to end and re-verifying everything. Wanted to share the findings so you can take a look.

Initial pass flagged two mediums but on re-verification both turned out to be intentional design decisions documented in your fix commit and an inline code comment. Steel-manning defeated both. Final result is clean to merge with a few optional polish items.

### Quick-glance findings

| ID | Severity | Where | Note |
|----|----------|-------|------|
| LOW-1 | Low | `models.rs:240` (`format_prompt`) | `"gemma3n"` template alias is unreachable. No `ModelDef` uses `template: "gemma3n"`. Either remove or add a comment that it is reserved for a future model. |
| LOW-2 | Low | `SetupOverviewStep.tsx:27` | Size label uses `model === 'qwen3.5:2b' ? '~1.2 GB' : '~2.6 GB'`. Works today, but fragile if a non-Qwen model is ever recommended. `DownloadProgressStep` already has a proper lookup. |
| LOW-3 | Low | `DownloadProgressStep.tsx` + `DownloadProgressToast.tsx` | `SUMMARY_MODEL_SIZES_MB` and `getSummaryModelSizeMb` are duplicated across both files. Worth lifting into `onboarding-summary-model.ts`. |
| INFO-1 | Info | `vad.rs` + `device_detection.rs` | These look unrelated to the Qwen scope. The VAD test was weakened from `assert!(segments.len() >= 6)` to `assert!(!segments.is_empty())`, which removes a real invariant. Worth splitting into its own PR or strengthening the test (use a speech-like signal that Silero will segment, not a sine wave). |
| INFO-2 | Info | `greedy_strict` preset | `top_k: 1` with `temperature: 0.0` is redundant in greedy mode (top_k/top_p ignored by llama.cpp), but `sanitize_for_llama_helper` clamps it safely. Style only. |
| INFO-3 | Info | Release notes | Gemma sampling preset changed (`temperature: 1.0 -> 0.1`, `top_k: 64 -> 20`, `top_p: 0.95 -> 0.88`). The `gemma_models_use_huggingface_urls_and_tight_structured_sampling` test makes the intent explicit. Existing Gemma users will see more deterministic summaries after this update. Probably worth a one-line mention in the release notes. |

### What got invalidated on re-verification

Two findings I initially flagged as medium turned out to be intentional. Calling them out so you know they were considered and dropped:

1. The hardcoded `selectedModel: ''` in `verifyModelStatus`. Your commit `ce499ab` message documents this as the new policy: *"Treat onboarding summary readiness as selected/recommended model readiness instead of any available built-in summary model."* Makes sense as the Gemma to Qwen migration path.
2. The removal of the 3-second Parakeet-priority delay in `startBackgroundDownloads`. Your inline comment (`Start selected Summary Model download immediately so completion cannot race the request`) plus the same commit message make the correctness-over-bandwidth tradeoff clear.

### Asks

- Could you re-verify the findings above against your own reading first? Several of mine reversed once I had more context, so a second pair of eyes would be good.
- If anything looks wrong, please push back with details and I will dig further.
- If they hold up, the LOW items are optional polish, but INFO-1 (the unrelated audio/VAD scope drift) is the one most worth acting on, either split it out or strengthen the test.
- If you choose not to act on any of them, a quick response comment with the reasoning would help close the loop.

Otherwise this looks clean to merge. Nice work on the migration story for existing Gemma users staying functional, and on the test coverage for the new resolver.

safvanatzack:
@sujithatzackriya 
Deferred:
- INFO-2: leaving `top_k: 1` as-is for now. It is redundant in greedy mode, but harmless, sanitized, and keeps the preset explicit.
- INFO-3: release-note coverage for the Gemma sampling change is deferred to release/QA notes rather than this code commit.

sujithatzackriya:
## Fresh re-review at `c0c973d` (deep, root-cause + regression pass)

Thanks for the follow-up commits. I re-reviewed the whole change end to end against `devtest` at the current tip (`c0c973d`, +591/-160), not the older revision, and re-verified every finding against the actual code on the PR branch. Posting the full picture so you can validate.

### What is confirmed fixed (re-verified, no action needed)

| Prior finding | Status | Evidence |
|---|---|---|
| H1: Gemma 1B 404 | Fixed | URL now has the `google_` prefix; `curl -IL` returns `302 -> 200`, `x-linked-size 1069306624` (1.07 GB). Test asserts the corrected URL. |
| All 4 model URLs | Verified | Qwen 2B/4B and Gemma 1B/4B all resolve `200` with `Content-Length` matching the registry within rounding. |
| LOW-1: `gemma3n` template alias | Resolved | Absent from `format_prompt` at `c0c973d`. |
| LOW-2: fragile size ternary | Resolved | `SetupOverviewStep` model/size state was never rendered even on `devtest`; removing it is a clean no-op for UX. |
| LOW-3: duplicated size map | Resolved | Lifted into `src/lib/onboarding-summary-model.ts`. |
| Existing Gemma users re-download | Not an issue | `model_manager` keys the local path off `gguf_file`, not the URL basename, so the HF vs old-CDN basename difference is inert. |

### Remaining findings

| ID | Sev | Area | Summary |
|---|---|---|---|
| MED-1 | Medium | UX / recording lifecycle | Parakeet (the only download gating `Continue`) is now coupled to summary-model resolution. If model-manager init fails, onboarding is hard-stuck with only a `console.error`. Regression vs `devtest`. |
| MED-2 | Medium | Merge readiness / LLM KPI | Qwen 3.5 becomes the default summary engine for every fresh install with greedy decoding and a hand-rolled non-thinking template, and "Manual testing performed" is unchecked. Code looks correct, but summary quality is unvalidated. |
| LOW-1 | Low | UX / model mgmt | First-run download volume is roughly 2.3x larger (Qwen 4B 2.6 GB default) and the Parakeet bandwidth-priority delay was removed. Intentional, but worth a release-note expectation. |
| LOW-2 | Low | Scope drift | `lib.rs` rewrites 14 command-registration paths. Functionally equivalent (`summary/mod.rs` re-exports unchanged), but unrelated to Qwen and enlarges the merge-conflict surface. |
| INFO-1 | Info | Scope drift | `vad.rs` large-file test rewrite, unrelated to Qwen. Net stronger and the rationale is sound, but consider splitting. |
| INFO-2 | Info | Data accuracy | `qwen3.5:2b` `size_mb = 1270` vs actual ~1221 MB. Display-only estimate, no functional impact. |
| INFO-3 | Info | Behavior change | Gemma sampling preset changes silently for existing users (intentional). Needs a release note. |
| INFO-4 | Info | Regression surface | `llama-cpp-2` 0.1.128 -> 0.1.139 affects the shared built-in LLM path (Gemma too), not just Qwen. No tracked Cargo.lock, so the resolved 0.1.x is non-deterministic across builds. |
| INFO-5 | Info | Scope drift | `templates/mod.rs` doc-comment edit, unrelated to Qwen. Harmless. |

### MED-1 detail (root cause)

`Continue` is enabled only when `parakeetDownloaded` is true. `parakeetDownloaded` flips only after `startDownloads()` runs, which is gated by `useEffect(... if (!selectedSummaryModel) return; ..., [selectedSummaryModel])` in `DownloadProgressStep.tsx`. `selectedSummaryModel` (default `''`) is set in all three setter paths (context mount, `verifyModelStatus`, `fetchRecommendation`) only after `await invoke('builtin_ai_is_model_ready', ...)` resolves. `builtin_ai_is_model_ready` rejects if `init_model_manager` fails (cannot create the app data dir, permission, disk, sandbox). When that happens the transcription engine download never starts, so onboarding cannot proceed and there is no retry affordance (Parakeet stays `waiting`, not `error`).

On `devtest` the default was `gemma3:1b` and the download effect ran unconditionally on mount, so transcription onboarding survived a broken summary-model directory and the user could still finish and configure an external LLM. This PR removes that escape path.

Suggested minimal fix: restore an initial default in `OnboardingContext.tsx` matching the `database/commands.rs` `unwrap_or("qwen3.5:4b")` fallback:

```ts
const [selectedSummaryModel, setSelectedSummaryModel] = useState<string>('qwen3.5:4b');
```

One line, no logic change, re-decouples Parakeet from summary-model resolution.

### Decision gate (please action before merge)

1. MED-1: apply the one-line default restore, or explicitly accept the failure-path regression with a reason. Blocking until decided.
2. MED-2: run a manual summary-quality smoke test of Qwen 3.5 2B and 4B, plus a Gemma regression check on the bumped `llama-cpp-2`, since this becomes the global default. Blocking until done or explicitly waived with rationale.
3. LOW-2, INFO-1, INFO-5: either revert the scope drift to keep the PR focused, or note it in the PR description. Non-blocking.
4. INFO-3: add the Gemma sampling change to release notes. Non-blocking.

### Ask

Could you re-verify all of the above against your own reading first? Several findings reversed across earlier rounds once more context was in hand, so a second pair of eyes from you would help. If anything is wrong, push back with the file and line and I will dig further. If a finding holds and you choose not to act, a short reasoning comment will close the loop. Happy to re-review once MED-1 and MED-2 are addressed.

safvanatzack:
- MED-1: fixed by decoupling Parakeet startup from summary model readiness. Parakeet now starts independently, while summary download still waits for the backend-selected model, so low-RAM systems do not eagerly fall back to Qwen 4B.
- MED-2: PR body has been updated after manual smoke testing.
- LOW-1: expected behavior for the Qwen rollout. If needed, this will be covered in the main release notes.
- LOW-2: keeping the `lib.rs` command-registration path updates because they were needed for `cargo check` / Tauri macro resolution.
- INFO-1: keeping the VAD test update as broader application test hardening.
- INFO-2: fixed. Qwen 3.5 2B metadata now uses the verified `1221 MiB` size.
- INFO-3: Gemma sampling note will be added to final release notes if needed.
- INFO-4: `Cargo.lock` is already being staged on `devtest`; resolving the conflict and rebuilding should settle the dependency pinning concern.
- INFO-5: keeping the template doc edit, as it is harmless

sujithatzackriya:
## Code Review PR #462 (deep / agent / root-cause)

First, thanks for the iteration on this. Seven commits across nine days, each responding to prior review feedback, is real work and the migration story for existing Gemma users is thoughtful (`gguf_file` basenames preserved, local detection intact). The intent here is to give you the architectural picture under the surface findings so the next round of fixes lands at the level the recurrence class actually lives at, not one symptom at a time.

**Verdict: 🔴 NEEDS CHANGES.** Reviewed head `feat/qwen3-5-model` @ `ee92b4d` vs `devtest` (merge-base `b7903bf`), 18 files / +639 / -161. Deep mode: dedicated architectural agent ran `/think` (five-whys-plus, archetypes, systems-thinking, inversion, map-territory), call-path sweep over 15 read/write sites for the "selected summary model" signal, per-finding re-verification by an independent reverify pass (HIGH-3 from the prior round downgraded to LOW after confirming a redundant stable listener mitigates user impact; 21 confirmed, 1 retracted). No independent codex cross-check available for this pass, so the P0s carry full weight rather than being softened.

> ✅ Branch is **clean / `MERGEABLE`** vs `devtest` (no rebase needed).

### 1. Overview

| ID | Pri | Sev | Title | Area | Introduced by | File |
|----|-----|-----|-------|------|---------------|------|
| **H-1** | **P0 must fix (merge blocker)** | High | Qwen 3.5 runs in pure-greedy regime; vendor warns of endless repetition without presence penalty | LLM sampling | `c84595f` *"feat(summary): add Qwen 3.5 built-in models"* | `frontend/src-tauri/src/summary/summary_engine/models.rs:30-37, 122, 135` + `llama-helper/src/main.rs:341-357` |
| **H-2** | **P0 must fix (merge blocker)** | High | macOS recommendation drops the 16 GB ceiling; 8 GB / 4 GB Macs forced from 1 GB Gemma to 2.6 GB Qwen 4B | Recommendation policy | `c84595f` | `frontend/src-tauri/src/summary/summary_engine/commands.rs:21-27` + test `:413-415` |
| **M-1** | **P1 should fix** | Med | Gemma sampling silently shifted from Google defaults (1.0/64/0.95) to `tight_structured` (0.1/20/0.88) with no rationale | LLM sampling (shared Gemma path) | `c84595f` | `models.rs:148, 161` |
| **M-2** | **P1 should fix** | Med | Three concurrent `builtin_ai_get_recommended_model` + `is_model_ready(refresh:true)` invocations on every onboarding mount | Onboarding state / IPC | `c84595f, ce499ab, c0c973d` (three commits, three fetches) | `OnboardingContext.tsx:150, :147→:377-378` + `DownloadProgressStep.tsx:191→:158-181` |
| **M-3** | **P1 should fix** | Med | `verifyModelStatus` hard-codes `selectedModel: ''`, silently replacing the user's prior model with the new Qwen recommendation | Onboarding state | `ce499ab` *"verify selected summary model readiness"*. **The fix commit itself; commit name claims a guarantee the code does not hold.** | `OnboardingContext.tsx:382-386` |
| **M-4** | **P1 should fix** | Med | `loadOnboardingStatus` runs `verifyModelStatus` even for completed users; overwrites their finished selection on every app start | Onboarding state | `ce499ab` (load-bearing on this regression; surrounding scaffold predates) | `OnboardingContext.tsx:332-353` |
| L-1 | P2 follow-up | Low | `Special::Tokenize` / `token_to_bytes` deprecated in `llama-cpp-2 0.1.133`; breaks on next minor bump | Dep hygiene | `c84595f` | `llama-helper/src/main.rs:15, :372` |
| L-2 | P2 follow-up | Low | `format_prompt` naive replace + `parse_special=true` allows transcript-supplied `<\|im_end\|>` / `<end_of_turn>` to tokenize as special tokens | Tokenization safety | pre-existing (PR multiplies surface) | `models.rs:244-247` + `llama-helper str_to_token` |
| L-3 | P2 follow-up | Low | `top_k.max(1)` overwrites `top_k=0` (llama.cpp "disabled") with greedy-equivalent 1; dormant footgun | Sampling sanitize | `c84595f` | `models.rs:56` |
| L-4 | P2 follow-up | Low | `database/commands.rs:194` `unwrap_or("qwen3.5:4b")` contradicts `models::get_default_model()` returning `qwen3.5:2b`; also misaligned with `f4fd19d`'s stated "avoid eager Qwen 4B fallback" intent | Default-value alignment | `c84595f` | `database/commands.rs:193-194` |
| L-5 | P2 follow-up | Low | Toast briefly flashes `0.0 MB / 0.0 MB` on completion: backend emits `total_mb: 0`, `total_mb ?? fallback` does not fall through `0` | UX | pre-existing pattern | `DownloadProgressToast.tsx:319` |
| L-6 | P2 follow-up | Low | New `onboarding-summary-model.test.mjs` not wired into any `pnpm` script or CI workflow | Test infra | `ce499ab` | `frontend/package.json`, `.github/workflows/*.yml` |
| L-7 | P2 follow-up | Low | Parakeet listener `useEffect` dep array `[selectedSummaryModel]` causes teardown churn on initial transition; redundant stable listener in `DownloadProgressStep` mitigates current user impact | Onboarding listener | pre-existing + `c84595f` changed initial state from `'gemma3:1b'` to `''` (turned dormant defect into near-certain trigger) | `OnboardingContext.tsx:286` |
| L-8 | P2 follow-up | Low | VAD large-file test scope drift: segment-count assert `>= 6` removed, monotonic-progress + 100 % asserts added; production code untouched | Scope hygiene | `c0c973d` *"strengthen VAD large-file test assertions"* | `frontend/src-tauri/src/audio/vad.rs:474-516` |
| L-9 | P2 follow-up | Low | `llama-cpp-2` Cargo.toml bump `0.1.128 → 0.1.139` is symbolic only; `Cargo.lock` pins `0.1.146` on both branches | Dep hygiene | `c84595f` | `llama-helper/Cargo.toml:10` |
| L-10 | P2 follow-up | Low | `size_mb` field stores MiB but UI labels as "MB"; 4-5 % under-statement of decimal download size | UX labelling | pre-existing (PR adds 2 more entries) | `models.rs:121, 133, 145, 159` + UI |
| ~~F-X~~ | retracted | | ~~Parakeet/summary coupling locks Continue~~ retracted as false positive: `DownloadProgressStep.tsx:225-279` has a stable redundant Parakeet listener with `[]` deps that flips `setParakeetDownloaded` independently. Re-verified at HEAD. | | | |

All commits authored by **Mohammed Safvan**. Verified clean (steel-manned, NOT findings): all 4 HuggingFace download URLs return HTTP 200 with `Content-Length` matching `size_mb` within ±1 MiB; existing Gemma installs auto-detect on disk (`gguf_file` basenames unchanged); Qwen non-thinking chat template byte-exact match with official Qwen tokenizer `apply_chat_template(enable_thinking=False)`; stop tokens correct per model; sampling read fresh per generation (no stale cache); Tauri IPC compatibility (no commands renamed/moved); `cargo check -p meetily` PASS (0 errors, 7 pre-existing warnings); security surface clean (no user input → URL, no path traversal, TLS verified).

### 2. The structural picture under all six findings

These are not six independent bugs. They are six symptoms of **two missing primitives** in the codebase.

**Primitive A absent: `Request::Generate` is a sampling-feature-poor IPC contract.** The JSON wire format in `llama-helper/src/main.rs:25-35` carries only `temperature / top_k / top_p / stop_tokens`. There is no field for `presence_penalty`, `frequency_penalty`, or `repeat_penalty`. The sampler chain at `llama-helper/src/main.rs:341-357` never composes `LlamaSampler::penalties(...)`. **The codebase cannot express the Qwen 3 vendor-recommended sampling regime even if a developer wanted to** the IPC ceiling sits below the vendor floor. When you reached Qwen, the only safe lever left was `temperature = 0.0`, decorated with a `top_k=1, top_p=1.0` that the very next branch in the sampler ignores. That is `greedy_strict`. **H-1 is not a sampling-preset bug; it is a forced workaround for a missing IPC field.** The name `greedy_strict` announces the workaround in the type system. The companion `SamplingParams::sanitize_for_llama_helper` (`models.rs:50-67`) is the further evidence: it is a downcast from a richer in-process type to the impoverished IPC type, a wrapper around the constraint rather than a fix for it. That is the **Shifting the Burden** archetype: the symptomatic solution (preset enums + sanitizer) suppresses pressure to widen the wire schema.

**Primitive B absent: "selected summary model" is ambient state re-derived everywhere, not owned data persisted once.** The "selected summary model" signal touches **15 sites across 2 persistence stores with 0 source-of-truth invariants enforced**:

| Site | Role | File:line |
|---|---|---|
| `recommend_summary_model(is_macos, ram)` | pure-function derivation | `summary_engine/commands.rs:21-27` |
| `builtin_ai_get_recommended_model` Tauri command | IPC wrapper | `summary_engine/commands.rs:386-391` |
| `database/commands.rs:194 unwrap_or` | fresh-install fallback (hard-coded) | `database/commands.rs:194` |
| `SettingsRepository::save_model_config` | canonical persistence write | `database/repositories/setting.rs:41` |
| `SettingsRepository::get_model_config` | canonical persistence read | `database/repositories/setting.rs:32` |
| `complete_onboarding` | writes to SettingsRepository | `onboarding.rs:168-198` |
| `OnboardingStatus.model_status: ModelStatus` | second store schema | `onboarding.rs:11-24` (**no model-name field**) |
| `onboarding-status.json` Tauri-store file | persists ModelStatus | `onboarding.rs:122-138` |
| `OnboardingContext.selectedSummaryModel` React state | UI mirror, initialized `''` | `OnboardingContext.tsx:94` |
| `initializeSummaryModelSelection` mount effect | reader | `OnboardingContext.tsx:109-134` |
| `loadOnboardingStatus → verifyModelStatus` | second reader, passes `selectedModel:''` | `OnboardingContext.tsx:332, 377-388` |
| `DownloadProgressStep.fetchRecommendation` | third reader | `DownloadProgressStep.tsx:158-181` |
| `resolveOnboardingSummaryModelStatus` | reducer (collapses `''` to recommendation) | `lib/onboarding-summary-model.ts:20-30` |
| `api_get_model_config` | canonical UI-side read | `services/configService.ts:61-62` (used elsewhere, **not by OnboardingContext**) |
| `completeOnboarding` (frontend) | writer back to SettingsRepository | `OnboardingContext.tsx:447-475` |

The user's actual choice lands in SQLite (`settings` row). Every onboarding-side reader re-asks the recommendation function instead. `OnboardingContext` never invokes `api_get_model_config`. The `selectedModel: ''` literal at `OnboardingContext.tsx:382-386` is honest: the call site does not own the data it would need. **M-3 and M-4 are not coding mistakes. They are inevitable when the writer's persistence target (SQLite) and the reader's persistence target (Tauri JSON store) differ, and the reader was never taught to consult the writer's store.** M-2 is the same defect expressed through React lifecycle: three components each "own" the recommendation because no single owner exists.

**The systems archetype: Fixes That Fail** (with a Shifting-the-Burden overlay). Walk the seven commits:

| Commit | Symptom it patched | Symptom it created or left |
|---|---|---|
| `c84595f` *Add Qwen 3.5 + recommendation flip + restrained sampling* | introduces Qwen | `greedy_strict` (H-1), `is_macos OR ram≥8` (H-2), Gemma silently re-sampled (M-1) |
| `d187f1c` *Cap built-in model list scroll inside dialog* | dialog UX | unrelated, but is itself a symptom of churn |
| `384cd40` *Address review findings* | reviewer caught fallout | further patches the same path |
| `4c5a250` *Preserve summary model compatibility, prevent low-RAM Qwen 4B download* | low-RAM visible regression | "prevent" applied at frontend save path only; Rust `recommend_summary_model` untouched (H-2 still live) |
| `ce499ab` *Verify selected summary model readiness* | onboarding marks Gemma-only installs as ready when Qwen is the recommendation | introduces the `selectedModel: ''` literal (M-3); makes `loadOnboardingStatus` run verify for completed users (M-4) |
| `c0c973d` *Centralize size helpers, remove stale fetches* | duplicated state | reduces 4 fetches to 3, does not collapse to 1 (M-2) |
| `f4fd19d` *Decouple parakeet from summary readiness, avoid eager Qwen 4B fallback* | Parakeet-block-on-Qwen symptom from `ce499ab` | "avoid eager Qwen 4B fallback for low-RAM systems" applied at frontend save path again; H-2 still live in Rust |

Each fix targeted the previous fix's side effect, not the missing primitive. The commits with `[skip ci]` and the rapid cadence indicate the iteration loop has been productive at the symptom level but has not yet reached the structural lever. Every round the symptomatic fix is cheaper than the structural fix; every round the structural fix gets harder (more call sites to migrate).

### 3. 🔴 P0 H-1: Qwen runs in vendor-warned greedy regime

**Who/which commit/why.** Introduced by `c84595f` *"feat(summary): add Qwen 3.5 built-in models"*. The author's reasoning is in the commit message itself: *"restrained sampling defaults"*. The intent was conservatism, but the chosen instrument (greedy + decorative top_k/top_p) hits the exact failure mode Qwen's vendor documents.

**Root cause (architectural, not surface).** `models.rs:30-37` defines `SamplingParams::greedy_strict(stop)` with `temperature=0.0`. `models.rs:122` and `:135` attach it to both Qwen models. `llama-helper/src/main.rs:341-357` is the actual generation path:
```rust
let sampler = if temperature <= 0.0 {
    LlamaSampler::chain_simple([LlamaSampler::greedy()])     // ← Qwen lands here
} else {
    LlamaSampler::chain_simple([
        LlamaSampler::top_k(top_k), LlamaSampler::top_p(top_p, 1),
        LlamaSampler::temp(temperature), LlamaSampler::dist(seed),
    ])
};
```
The `top_k=1` and `top_p=1.0` in the `greedy_strict` literal are decorative; the sampler chain is bypassed entirely. The wire schema (`Request::Generate`, `llama-helper/src/main.rs:25-35`) has no `presence_penalty` / `frequency_penalty` / `repeat_penalty` fields, and `LlamaSampler::penalties` is never composed. **The structural defect is the IPC ceiling, not the preset choice.**

**Systems archetype.** Shifting the Burden: every time a new model family needs a regime the IPC cannot carry, the team adds a preset enum value and a sanitizer instead of widening the contract. The fundamental fix becomes harder each round.

**Five-whys to the structural cause.** 1) Why does Qwen risk repetition? Vendor requires presence penalty; greedy bypasses penalties entirely. 2) Why greedy? `greedy_strict` chosen as "safest". 3) Why was the "safest" preset still unsafe? Qwen's safety lever is presence penalty, not low temperature; low temperature is less safe on long inputs. 4) Why couldn't the author pick a Qwen-safe preset? `Request::Generate` schema has no penalty fields; sanitize downcasts away anything richer. 5) Why is the schema this narrow? Built when Gemma 3 was the only family. Gemma tolerates low temperature; widening was never forced.

**Reproducibility.**
1. Pre-state: fresh install on any macOS or non-macOS ≥ 8 GB RAM (full default user base after this PR).
2. Action: record a 20-30 minute meeting, click "Generate summary" using built-in Qwen 4B.
3. Expected: coherent summary respecting the prompt.
4. Actual: non-zero probability of degenerating into single-token / single-phrase repetition loop, especially for transcripts with repeated speaker names or boilerplate phrases. Indistinguishable from a pipeline freeze to the user.
5. Code path: `summary/commands.rs::generate_summary_request → client.rs:178 sanitize_for_llama_helper → JSON Request::Generate → llama-helper/main.rs:341 greedy() → llama.cpp`.
6. Affected users: everyone running the default summary model. The PR ships Qwen 4B as the default for the entire user base after merge.

**Tactical fix (≈ 4 LOC, unblocks this PR, low risk).** In `models.rs` lines 122 and 135, replace both Qwen `SamplingParams::greedy_strict(...)` calls with `SamplingParams::tight_structured(vec!["<|im_end|>".to_string()])`. Update the two test assertions at `models.rs:286, :294` to match. Grep-verified: `greedy_strict` has zero other callers.

**Why this does NOT reopen a known bug.** `tight_structured` = `temperature=0.1, top_k=20, top_p=0.88`. The vendor warning is specifically about **greedy** decoding producing repetition; with `temperature > 0` the chain composes `top_k(20) → top_p(0.88) → temp(0.1) → dist(seed)`, introducing non-zero token diversity that breaks the deterministic loop attractor. Gemma 4B has been running on `tight_structured` across the same path on this branch since `c84595f` without a repetition regression report (empirical evidence is the Gemma users in M-1's blast radius). It is still below vendor's `temp=0.7, top_p=0.8, top_k=20` but materially safer than greedy.

**Strategic fix (follow-up PR, the real fix).** Widen `Request::Generate` and the sampler chain. Add optional `presence_penalty / frequency_penalty / repeat_penalty / repeat_last_n` fields at `llama-helper/src/main.rs:25-35`. Compose `LlamaSampler::penalties(repeat_last_n, repeat_penalty, frequency_penalty, presence_penalty)` into the chain at `:347-356`, including in the `temperature <= 0.0` branch so low-temperature runs can still penalize repetition. Mirror the fields onto `SamplingParams`. Delete `greedy_strict` (the workaround type goes away). Then express Qwen as `temp=0.7, top_p=0.8, top_k=20, presence_penalty=1.5` per vendor. Wire-compatible because new fields are optional; ship as a stand-alone PR after this one; follow-up PR then flips Qwen presets.

### 4. 🔴 P0 H-2: macOS recommendation drops the 16 GB ceiling

**Who/which commit/why.** Introduced by `c84595f`. Layered on by `4c5a250` and `f4fd19d` whose commit messages claim *"prevent low-RAM onboarding from downloading Qwen 4B"* and *"avoid eager Qwen 4B fallback for low-RAM systems"*. Both fixes were applied at the **frontend save path** only, leaving the Rust recommendation function untouched. The Rust function continues to return `qwen3.5:4b` for any macOS regardless of RAM. **The fix commits' stated guarantee is not held by the code.**

**Root cause.** `summary_engine/commands.rs:21-27`:
```rust
pub(crate) fn recommend_summary_model(is_macos: bool, system_ram_gb: u64) -> &'static str {
    if is_macos || system_ram_gb >= 8 {
        "qwen3.5:4b"
    } else {
        "qwen3.5:2b"
    }
}
```
The boolean `||` rather than `&&` is the proximate defect, and the test at `commands.rs:413-415` codifies the regression in CI:
```rust
assert_eq!(recommend_summary_model(true, 4), "qwen3.5:4b");   // 4 GB Mac, asserted to get 4 B
```
Devtest baseline at `commands.rs:372-376` had a deliberate 16 GB ceiling: `if is_macos && system_ram_gb > 16 { "gemma3:4b" } else { "gemma3:1b" }`. The new policy removes that ceiling and inverts the operator.

**Systems archetype.** Same Fixes-That-Fail loop as H-1. The frontend-only patches at `4c5a250` and `f4fd19d` are downstream of an unchanged Rust source, so they cannot enforce the policy globally; SQLite fresh-install fallback at `database/commands.rs:194` (`unwrap_or("qwen3.5:4b")`) and downstream consumers (DownloadProgressStep) re-derive from the same compromised function.

**Reproducibility.**
1. Pre-state: 8 GB MacBook Air M1 / M2 / M3 (most common Apple-Silicon configuration); browser + meeting app + Meetily running.
2. Action: fresh install or onboarding re-entry.
3. Expected: a model footprint matching the user's hardware tier; never larger than the previous release's gated ceiling.
4. Actual: onboarding downloads 2.6 GiB; first summary loads ~3-4 GB resident (model + 32 k KV cache) into a system the codebase previously gated at 16 GB. Swap thrash and likely OOM during summary generation.
5. Code path: `OnboardingContext.tsx:109 → builtin_ai_get_recommended_model → recommend_summary_model(true, 8) → "qwen3.5:4b" → download → model load`.
6. Affected users: every fresh-install Mac, every fresh-install non-macOS device ≥ 8 GB RAM. The 4 GB case is encoded in the test.

**Tactical fix (≈ 8 LOC, unblocks this PR, low risk).**
```rust
pub(crate) fn recommend_summary_model(is_macos: bool, system_ram_gb: u64) -> &'static str {
    match (is_macos, system_ram_gb) {
        (true,  ram) if ram >= 16 => "qwen3.5:4b",
        (true,  _)                => "qwen3.5:2b",
        (false, ram) if ram >= 16 => "qwen3.5:4b",
        (false, _)                => "qwen3.5:2b",
    }
}
```
Update the test matrix at `commands.rs:413-415` (4 GB Mac → 2 B, 16 GB Mac → 4 B, 32 GB → 4 B, 4 GB non-mac → 2 B). Change `database/commands.rs:194` `unwrap_or("qwen3.5:4b")` to `unwrap_or("qwen3.5:2b")` for the safe-floor alignment. Leadership may want to keep `gemma3:1b` as the floor for ≤ 8 GB devices if footprint matters; that decision should be explicit in the PR description.

**Strategic fix (follow-up).** Lift `recommend_summary_model` into the model registry itself. Each `ModelDef` declares `min_ram_gb` and `recommended_platforms`. The recommendation function becomes a fold over the registry. Adding a model in future cannot regress recommendation for existing models.

### 5. 🟠 P1 M-1, M-2, M-3, M-4

**M-1 Gemma sampling silently shifted from Google defaults.** `c84595f`. `models.rs:148` and `:161` were inline `SamplingParams { temperature: 1.0, top_k: 64, top_p: 0.95, ... }` (Google's published defaults for Gemma 3 IT) on devtest; now both call `SamplingParams::tight_structured(...)`. No commit message or PR description explains the change. Every existing Gemma user sees materially flatter, more deterministic, likely more terse summaries on first generation after upgrade. **Fix:** either revert `models.rs:148, :161` to the inline Google defaults (and rename the related test), or document the change in the PR description and release notes; staying silent on a vendor-default deviation across a release boundary violates user expectation.

**M-2 Three concurrent recommendation+ready fetches on every mount.** `c84595f, ce499ab, c0c973d` (each commit added or perpetuated a fetch site). `OnboardingContext.tsx:150` → `initializeSummaryModelSelection` (recommend + ready at `:109, :113`); `OnboardingContext.tsx:147` → `loadOnboardingStatus → verifyModelStatus` (recommend + ready at `:377, :378`); `DownloadProgressStep.tsx:191` → `fetchRecommendation` (recommend + ready at `:158, :166`). All three pass `refresh: true`, triggering three concurrent `scan_models` filesystem walks on cold start; all three call `setSelectedSummaryModel` and `setSummaryModelDownloaded`. The race resolves to a consistent final value (recommendation is pure of platform+RAM) but cold-start cost is real and the three setter races can produce visible "card flicker" during the first second. **Fix:** make `OnboardingContext` the single source of truth; have `DownloadProgressStep` and `verifyModelStatus` read context state instead of independently fetching. Behind that, share an in-flight promise so even an explicit refresh collapses to one IPC + one scan.

**M-3 `verifyModelStatus` hard-codes `selectedModel: ''`.** `ce499ab` *"verify selected summary model readiness"* is the introducing commit. **The commit name claims a guarantee the code does not hold:** at `OnboardingContext.tsx:382-386` the call passes the literal `''` for `selectedModel`, ignoring `savedStatus.model_status.summary` entirely. Inside `lib/onboarding-summary-model.ts:24`, `selectedModel || recommendedModel` then collapses to the recommendation. `saveOnboardingStatus` (`OnboardingContext.tsx:426-437`) does not persist `selectedSummaryModel`; the Rust schema at `onboarding.rs:11-24` has no field for it. **Fix:** add `selected_summary_model: Option<String>` to `ModelStatus` (Rust); include it in `saveOnboardingStatus` payload; pass `savedStatus.model_status.selected_summary_model.unwrap_or_default()` at `:383` instead of `''`. Best paired with M-4 and the strategic Primitive-B fix; together this closes the recurrence class.

**M-4 `loadOnboardingStatus` runs `verifyModelStatus` even for completed users.** `ce499ab` (the load-bearing change for this regression; the surrounding scaffold predates). `OnboardingContext.tsx:332-353` calls `verifyModelStatus(status)` unconditionally including when `status.completed === true`. For users who completed onboarding with Gemma before this PR landed, the next session start silently flips `selectedSummaryModel` to the Qwen recommendation and marks `summaryModelDownloaded = false`. **Fix:** short-circuit when completed:
```ts
if (status.completed) {
  setCompleted(true);
  setCurrentStep(status.current_step);
  return;
}
const verifiedStatus = await verifyModelStatus(status);
```
This change is independently shippable from M-3; together they close the recurrence class.

### 6. 🟡 P2 follow-up

- **L-1** `llama-helper/src/main.rs:15, :372` use `Special::Tokenize` and `token_to_bytes`, both deprecated in `llama-cpp-2 0.1.133+`. `c84595f`. Rename to `token_to_piece_bytes` while the file is being touched.
- **L-2** `models.rs:244-247` does naive `replace("{user_prompt}", user_prompt)`; downstream `str_to_token` uses `parse_special=true` (8th C arg hardcoded in `llama-cpp-2-0.1.146/src/model.rs`). A transcript containing literal `<|im_end|>` or `<end_of_turn>` tokenizes as the actual special token. Pre-existing, but PR multiplies the surface area. Sanitize `user_prompt` against known template markers before substitution.
- **L-3** `models.rs:56` `top_k.max(1)` overwrites llama.cpp's "disabled" sentinel `0` with greedy-equivalent `1`. Dormant (no preset uses `0`). Document the contract or use `.max(0)`.
- **L-4** `database/commands.rs:194 unwrap_or("qwen3.5:4b")` contradicts `models::get_default_model().name == "qwen3.5:2b"`, and is misaligned with `f4fd19d`'s stated "avoid eager Qwen 4B fallback" intent. Dead path today (RAM probe never errors). Align via `models::get_default_model().name` or change to `"qwen3.5:2b"`.
- **L-5** `DownloadProgressToast.tsx:319` uses `total_mb ?? getSummaryModelSizeMb(model)`; JS `??` does not fall through `0`, and backend `summary_engine/commands.rs:175-187` emits literal `total_mb: 0` on completion. Toast flashes `0.0 MB / 0.0 MB` briefly. Use `||` instead.
- **L-6** `frontend/tests/lib/onboarding-summary-model.test.mjs` not invoked by any `pnpm` script or CI workflow. `ce499ab`. Add `"test": "node tests/lib/*.test.mjs"` to `frontend/package.json`; invoke from CI.
- **L-7** `OnboardingContext.tsx:286` Parakeet listener dep `[selectedSummaryModel]` does not reference the variable in the body. Stable redundant listener in `DownloadProgressStep.tsx:225-279` currently mitigates, so user impact is bounded. Change dep to `[]`.
- **L-8** VAD large-file test scope drift in `c0c973d`: segment-count assert weakened, monotonic-progress + 100 % asserts added. Production code untouched. Acceptable as documented; ideally split into its own PR.
- **L-9** `llama-helper/Cargo.toml:10` bump `0.1.128 → 0.1.139` is symbolic: `Cargo.lock` pins `0.1.146` on both branches. Either pin `=0.1.146` for reproducibility or document the symbolic floor.
- **L-10** `size_mb` stores MiB but UI shows "MB"; 4-5 % under-statement of decimal-MB download size. Pre-existing convention; PR perpetuates rather than introduces. Rename or convert at the boundary.

> Note: L-4, L-7, and L-9 are in-the-small echoes of the same macro root cause as H-2 and M-2 (shared mutable defaults / multiple stale "selected model" interpretations / dependency drift). Fixing the pattern via Primitive-A and Primitive-B closes most of them by construction.

### 7. Architectural relevance / blast radius

- **LLM processing KPI direct hit.** H-1 hits every default install on long transcripts; M-1 silently re-samples every existing Gemma install. Both reach the shared `LlamaSampler` chain.
- **Recording lifecycle KPI indirect hit.** Summary readiness gates onboarding completion; M-4 makes already-completed users re-enter a download flow on app start.
- **Model management KPI direct hit.** M-2 causes three concurrent `scan_models` walks per onboarding mount; H-2 causes 8 GB Macs to attempt loading 2.6 GB models into a configuration the codebase previously gated at 16 GB.
- **Tauri IPC** is safe today (no commands renamed or moved). The Strategic widening of `Request::Generate` and the bump to `OnboardingStatus.model_status` are both forward-compatible (new optional fields), so the strategic fix is independently shippable from this PR.
- **Sampling path reaches Gemma too.** The `temperature <= 0.0` branch in `llama-helper` is shared. Any future model adopting `greedy_strict` hits the same defect. The strategic widening preserves Gemma's existing behavior on `temp=0.1` because penalty defaults of `0.0` are a no-op in the sampler chain (verify against Gemma summary fixtures before shipping).
- **Pre-emptive non-mitigations.** Setting `NUM_WORKERS=1` or single-threading the recommendation will not fix M-2 because the race is across React component lifecycles, not threads. Patching `recommend_summary_model` alone will not fix M-3 / M-4 because those readers do not consult the canonical SQLite store.

### 8. Why H-1 and H-2 cannot be deferred

**Invariant 1.** *Qwen 3.5 MUST NOT be invoked in a regime its vendor documents as producing endless repetition.* The current PR head invokes exactly that regime as the default for the entire user base after merge. The blast surface is every install. Deferring means shipping a model that on long meetings can degenerate into single-token loops, indistinguishable from a recording pipeline freeze to the user, and costing support hours diagnosing a class of bug whose root cause is two lines in `models.rs`. The tactical fix is one-line per model plus a test update, and the existing Gemma user base running the same preset gives empirical evidence the fix is safe.

**Invariant 2.** *An 8 GB Mac MUST NOT be silently asked to load a 2.6 GB model into a configuration the codebase previously gated at 16 GB.* The PR not only ships that policy, it codifies it in a test (`commands.rs:413-415`) so the regression is immortalized in CI. Deferring means OOM crashes during the most critical UX moment in the product (onboarding download then first summary). The tactical fix is a 4-arm match statement.

The `[skip ci]` chain on every fix commit in this branch is itself a signal: the team has been responsive at the symptom level but each round has deferred the type-level corrections (IPC widening, status schema bump, single-reader collapse). That is the Shifting-the-Burden loop. The explicit ask: **land the two H tactical fixes in this PR, land M-3 + M-4 alongside (they share one diff: persist `selected_summary_model` on `ModelStatus` and short-circuit `verifyModelStatus` for completed users), file the strategic widening of `Request::Generate` and the single-source-of-truth refactor as tracked follow-ups, and at minimum disclose M-1 in the release notes.** That sequence closes the recurrence class for the next model family instead of narrowing it again.

Thanks again for the work and for absorbing earlier review rounds so quickly. The decisions visible in the commit chain (preserving Gemma local filenames, switching to the canonical HuggingFace mirror, the per-meeting summary language follow-through from #455 not regressing here) reflect care for users on the upgrade path. The structural feedback above is meant in the same spirit.

<sub>Full reviewer report retained locally at `PR-462-REVIEW.md` (not committed). Line numbers are PR-head blobs at `ee92b4d`.</sub>

safvanatzack:
@sujithatzackriya Rechecked the comments and added the fixes

Deferring the remaining two P2 items for this PR:

- **L-6**: test script/CI wiring only; no runtime impact. Better handled as separate test-infra cleanup.
- **L-8**: VAD test-scope hygiene only; production code is untouched and unrelated to this Qwen/model-management PR.

I think both are safe to defer since they do not affect this PR’s runtime behavior.

sujithatzackriya:
Great work on this, really cleanly executed. Thanks for the careful Qwen 3.5 rollout. Merging now.

PR #470, opened 2026-05-16
fix(summary): prevent BlockNote rendering crashes with patch + boundary

passabilities:
## Description

The summary editor crashes for some meetings with a `Cannot read properties of undefined (reading 'localsInner')` or `dom is undefined` error originating in prosemirror-view, taking down the entire meeting page. This PR adds three independent layers of defense:

1. **Patch `prosemirror-view` to honor `{dom, contentDOM}` toDOM specs.** BlockNote 0.36's node specs sometimes return a `{dom, contentDOM}` shape from `toDOM`, but upstream prosemirror-view unconditionally forwards the result to `DOMSerializer.renderSpec`, which throws when given that shape. The patch checks for the shape and uses it directly when present.

2. **Disable BlockNote's `placeholder` plugin.** BlockNote 0.36's placeholder plugin produces a `DecorationSource` that's missing `localsInner`, which crashes `DecorationGroup.locals()` in prosemirror-view during the initial editor mount. We have no UX dependency on the placeholder, so disabling it is the cleanest fix.

3. **Wrap the editor in `SummaryErrorBoundary`.** Even with (1) and (2), any future regression in the editor stack would otherwise take down the whole page. The boundary catches render failures and shows a fallback with the raw markdown plus a retry button, so a single bad summary can't break navigation.

## Files

- `frontend/patches/prosemirror-view@1.41.8.patch` — new pnpm patch (also patches `dist/index.cjs` and `dist/index.js`)
- `frontend/pnpm-workspace.yaml` — registers the patch via `patchedDependencies`
- `frontend/src/components/AISummary/BlockNoteSummaryView.tsx` — `disableExtensions: ['placeholder']` and wraps both render paths in the boundary
- `frontend/src/components/AISummary/SummaryErrorBoundary.tsx` — new class component with fallback UI

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Testing

- [x] Manual testing performed — opened previously-crashing meetings and verified the editor renders; intentionally threw inside `Editor` to confirm the boundary catches and the rest of the UI stays interactive.
- [ ] Unit tests added — N/A (boundary is exercised through manual fault injection; patch is a third-party dependency hot-fix).

## Checklist

- [x] Self-reviewed the code
- [x] Added comments for the non-obvious parts (placeholder disable + patch rationale)
- [x] No documentation needed

### DIFF
--- a/frontend/patches/prosemirror-view@1.41.8.patch
+++ b/frontend/patches/prosemirror-view@1.41.8.patch
@@ -1,0 +1,73 @@
+diff --git a/dist/index.cjs b/dist/index.cjs
+index ce89be47cebd61a79c6039950f6dcd10df0e3c7e..d1baf2b64f4b54c7f7abc0cc3653debd71503562 100644
+--- a/dist/index.cjs
++++ b/dist/index.cjs
+@@ -1263,7 +1263,12 @@ var MarkViewDesc = function (_ViewDesc3) {
+     value: function create(parent, mark, inline, view) {
+       var custom = view.nodeViews[mark.type.name];
+       var spec = custom && custom(mark, view, inline);
+-      if (!spec || !spec.dom) spec = prosemirrorModel.DOMSerializer.renderSpec(document, mark.type.spec.toDOM(mark, inline), null, mark.attrs);
++      if (!spec || !spec.dom) {
++        var toDOMResult_pm_fix_m = mark.type.spec.toDOM(mark, inline);
++        spec = toDOMResult_pm_fix_m && toDOMResult_pm_fix_m.dom
++          ? toDOMResult_pm_fix_m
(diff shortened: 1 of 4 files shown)


sujithatzackriya:
@safvanatzack please address

safvanatzack:
@passabilities Thanks for the contribution

I think the ProseMirror patch/placeholder workaround here is mostly superseded by #472, which pins the BlockNote ProseMirror stack and aliases ProseMirror packages to single instances.

The useful follow-up from this PR may be the `SummaryErrorBoundary`, but I’d keep that separate from the dependency/runtime fix.

Thank you

Issue #343, opened 2026-02-09
[FEATURE] Optional tray icon visibility and safer minimize-to-tray behavior


silvioprog:
## Feature Description
Add two optional behaviors:
1) Allow disabling the tray/system tray icon.
2) Allow "Close to tray" (minimize instead of exiting), but only when the tray icon is enabled.

This keeps privacy guarantees intact while giving users better control over UX and screen-sharing scenarios.

## Problem Statement
Currently:
- The tray icon is always visible.
- Clicking "Close" fully exits the app instead of minimizing to tray.

This causes two UX issues:
1) The always-visible tray icon can appear during professional screen sharing or video recordings (e.g., demos, presentations, trainings), which may be distracting or undesirable.
2) Users who expect "close" to behave as "minimize to tray" may accidentally stop recordings or workflows.

At the same time, hiding the tray icon while allowing background recording could raise valid privacy concerns.

## Proposed Solution
Introduce two settings:
- **Show tray icon** (default: enabled)
- **Close to tray instead of exiting** (only available if tray icon is enabled)

Behavior rules:
- If tray icon is disabled → app must fully exit on close.
- If tray icon is enabled → user can choose to minimize to tray on close.
- Recording should never continue invisibly without a visible tray icon.

This ensures:
- No hidden background recording.
- Clear user awareness.
- Better UX flexibility.

## User Story
As a professional user who records meetings or runs Meetily while screen sharing,
I want to control tray icon visibility and close behavior,
So that I can avoid visual distractions during presentations without compromising privacy or safety.

## Acceptance Criteria
- [ ] Tray icon visibility can be toggled via settings
- [ ] "Close to tray" option is only available when tray icon is enabled
- [ ] App always exits if tray icon is disabled
- [ ] Recording cannot continue without a visible tray icon
- [ ] Default behavior remains unchanged for existing users

## Technical Considerations
- Tray icon acts as the privacy indicator and safety guard
- Prevent background recording when tray icon is disabled
- Clear state handling between exit vs minimize
- Settings should be explicit and reversible

## Alternatives Considered
- Always forcing tray icon visibility (rejected due to screen sharing use cases)
- Allowing hidden background recording (rejected due to privacy concerns)

## Additional Context
Common scenarios where hiding the tray icon is useful:
- Corporate demos and sales presentations
- Training videos and tutorials
- Screen recordings for documentation
- Live presentations where minimal UI is expected

[...]

Issue #239, opened 2025-11-06
[BUG] Windows app sometimes crashing while recording


DPBattaglia:
## Bug Description
Sometimes when I press exit button on Meetily while recording it exits fully and is not minimised in the system tray.

## Current Behavior
Crashing

## Expected Behavior
No crashing

## Steps to Reproduce
1. During meeting exit - only sometimes this happens

## Environment
- OS: Windows App
- Browser: N/A
- Version: 0.1.1

## Screenshots/Videos
None

## Error Messages
None

## Additional Context
Only happens sometimes and would be good to review logs... unsure how though and where to go?

## Possible Solution
Unsure at this stage - requires some logs and troubleshooting I imagine

## Checklist
- [x] I have searched for similar issues
- [ ] I have provided all required information
- [ ] I have included screenshots/videos if applicable
- [ ] I have included error messages if applicable

safvanatzack:
@sujithatzackriya closable

PR #471, opened 2026-05-16
feat(tray): hide main window to tray on close instead of exiting

passabilities:
## Description

Meetily already ships a fully-wired system tray (start/stop recording, open window, quit), but closing the main window currently terminates the entire app, including any active recording. That's a footgun: a single accidental close ends the meeting.

This PR intercepts the main window's `CloseRequested` event and hides the window instead of letting it close. The tray icon persists, so users can re-open the window or quit explicitly from the tray menu.

## Behavior

- **Close button (X)** → window hides; app keeps running in tray; recording continues
- **Tray → Quit** → still exits cleanly (unchanged)
- **Other windows** → unaffected; only the `main` window is intercepted

## Files

- `frontend/src-tauri/src/lib.rs` — adds a single `on_window_event` handler (12 lines)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Testing

- [x] Manual testing on Linux — closed window with active recording; verified recording continued and window reopened via tray; verified Quit still exits.
- [ ] macOS / Windows verification welcome from maintainers.

## Follow-ups (not in this PR)

- A user preference toggle for `close = exit` vs `close = hide` would let people who prefer the old behavior opt back in. Happy to do this in a follow-up if maintainers want it.

## Checklist

- [x] Self-reviewed the code
- [x] No documentation needed (behavior change is intuitive and matches the existing tray menu)

### DIFF
--- a/frontend/src-tauri/src/lib.rs
+++ b/frontend/src-tauri/src/lib.rs
@@ -496,6 +496,18 @@ pub fn run() {
 
             Ok(())
         })
+        .on_window_event(|window, event| {
+            if let tauri::WindowEvent::CloseRequested { api, .. } = event {
+                if window.label() == "main" {
+                    api.prevent_close();
+                    if let Err(e) = window.hide() {
+                        log::error!("Failed to hide main window on close request: {}", e);
+                    } else {
+                        log::info!("Main window hidden to tray on close request");
+                    }
+                }
+            }
+        })
         .invoke_handler(tauri::generate_handler![
             start_recording,
             stop_recording,


safvanatzack:
Addresses #239 and #343

sujithatzackriya:
Thanks for this PR, @passabilities! This is a real UX win -- closing the window during an active recording was a genuine footgun, and the fix is minimal and correct. The `api.prevent_close()` + `window.hide()` pattern is exactly right for Tauri 2.

Reviewed . Merging as-is and tracking the following items for a follow-up:

**Finding 1 (Medium) -- Linux without AppIndicator**
On GNOME without `gnome-shell-extension-appindicator`, hiding the window leaves the app as an invisible orphan process with no recovery path. Before this PR, closing the window exited cleanly. A follow-up should either fall through to a clean exit when `window.hide()` fails, or document the AppIndicator requirement.

**Finding 2 (Low) -- `prevent_close()` ordering**
`api.prevent_close()` is called before `window.hide()`. If `hide()` ever fails, the close is already prevented and the window becomes uncloseable without killing the process. Fix: move `api.prevent_close()` inside the `Ok` branch of the `hide()` result match. This also partially mitigates Finding 1.

**Finding 3 (Low) -- Analytics session length**
`AnalyticsProvider.tsx` ends analytics sessions in `beforeunload`, which does not fire on window hide (only on actual destroy). Session-length metrics will be inflated when users adopt the hide-to-tray pattern. Acceptable for now; needs a dedicated event if session accuracy matters.

**Finding 4 (Low / UX) -- No first-close hint**
The window disappears silently. A one-time system notification ("Meetily is still running in the menu bar") would help discoverability, particularly on Windows.

@safvanatzack please note

PR #472, opened 2026-05-20
fix(frontend): stabilize BlockNote ProseMirror resolution [skip ci]

safvanatzack:
## Description
Fixes the BlockNote editor runtime crashes caused by duplicated or mismatched ProseMirror dependencies after a clean install.

This PR pins the ProseMirror packages used by BlockNote to known-good versions and adds webpack aliases so BlockNote and ProseMirror resolve through single module instances. It also replaces render-time BlockNote `require()` calls with static imports and ensures `blocksToMarkdownLossy` is wrapped only once to avoid recursive wrapper buildup during editor reuse.

## Related Issue
N/A

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [x] All tests pass

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [ ] No merge conflicts

N/A

## Additional Notes
Verified with a clean dependency install and production build. The lockfile now resolves BlockNote’s ProseMirror dependencies consistently, which should make the fix reproducible across developer machines

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -1,3 +1,8 @@
+const path = require('path');
+const tiptapPmResolveBase = path.dirname(require.resolve('@tiptap/pm/model'));
+const resolveFromTiptapPm = (pkg) =>
+  require.resolve(pkg, { paths: [tiptapPmResolveBase] });
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: false, // Disabled for BlockNote compatibility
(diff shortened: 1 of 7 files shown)


sujithatzackriya:
Thanks for this fix, @safvanatzack. The core approach is correct: `pnpm.overrides` + webpack `resolve.alias` together is the right way to eliminate duplicate ProseMirror registration crashes, and the static-import conversion in `Editor.tsx` is a clean improvement.

**Two things worth knowing before/after merge:**

**`prosemirror-transform` is a minor-version rollback, not just a patch (1.12.0 to 1.10.4).** That is the largest version gap in the override set. A production build passes because there is no runtime type check, but a regression would only surface when a specific editor operation triggers a code path that uses something added in 1.11 or 1.12. After merge, please do a manual run of table editing, list nesting, and block drag-and-drop to confirm nothing is broken.

**The `blocksToMarkdownLossy` wrapper in `Editor.tsx` does not guard the calls that actually matter.** `BlockNoteSummaryView` creates its own editor via `useCreateBlockNote` (line 84) and calls `blocksToMarkdownLossy` on that instance directly (lines 143, 172, 181). The wrapper in `Editor.tsx` applies to a different editor object that no external code can reach. The infinite-recursion bug fixed here was real, but unreachable in practice. The calls in `BlockNoteSummaryView` have no error handling. Not a blocker for this PR, but worth a follow-up.

**Merging.** The dependency fix is the substance of this PR and it is solid. The two points above are noted for awareness, not blocking.

safvanatzack:
> prosemirror-transform is a minor-version rollback, not just a patch (1.12.0 to 1.10.4). That is the largest version gap in the override set. A production build passes because there is no runtime type check, but a regression would only surface when a specific editor operation triggers a code path that uses something added in 1.11 or 1.12. After merge, please do a manual run of table editing, list nesting, and block drag-and-drop to confirm nothing is broken.

The manual testinng is done, no regression noticed and the second issue is addressed with changes

PR #498, opened 2026-06-03
hotfix(release): Fix summary base language and macOS Dock reopen behavior [skip ci]

safvanatzack:
## Description
This PR includes two fixes from the branch:

1. Stabilizes English summary output for non-English transcripts.
   - Enforces English as the base summary language in chunk, combine, and final report prompts.
   - Adds a soft English normalization pass when the selected summary language is English and the detected transcript language is non-English or unknown.
   - Skips the normalization pass when the transcript is detected as English.
   - Preserves the existing non-English target flow: English base summary first, then translation to the selected target language.
   - Saves the normalized English markdown as the English cache.
   - Falls back to the original pass-1 markdown if normalization fails, while still respecting cancellation.
   - Adds unit coverage for prompt enforcement, final-language branching, normalization fallback, and cache behavior.
   - Refactors the shared markdown transform path used by translation and English normalization to keep cancellation, cleanup, and error handling consistent.
 
2. Restores the main window when clicking the macOS Dock icon.
   - Handles Tauri's macOS `RunEvent::Reopen`.
   - Reuses the existing `tray::focus_main_window` helper.
   - Preserves the existing app exit cleanup behavior by switching the run-loop handler to a `match`.

3.Shorten transcript recovery grace period

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [ ] All tests pass

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

Not applicable.

## Additional Notes
No formatter was run.

### DIFF
--- a/frontend/src-tauri/src/lib.rs
+++ b/frontend/src-tauri/src/lib.rs
@@ -752,28 +752,35 @@ pub fn run() {
         .build(tauri::generate_context!())
         .expect("error while building tauri application")
         .run(|_app_handle, event| {
-            if let tauri::RunEvent::Exit = event {
-                log::info!("Application exiting, cleaning up resources...");
-                tauri::async_runtime::block_on(async {
-                    // Clean up database connection and checkpoint WAL
-                    if let Some(app_state) = _app_handle.try_state::<state::AppState>() {
-                        log::info!("Starting database cleanup...");
-                        if let Err(e) = app_state.db_manager.cleanup().await {
-                            log::error!("Failed to cleanup database: {}", e);
+            match event {
+                #[cfg(target_os = "macos")]
+                tauri::RunEvent::Reopen { .. } => {
+                    tray::focus_main_window(_app_handle);
+                }
+                tauri::RunEvent::Exit => {
+                    log::info!("Application exiting, cleaning up resources...");
+                    tauri::async_runtime::block_on(async {
+                        // Clean up database connection and checkpoint WAL
+                        if let Some(app_state) = _app_handle.try_state::<state::AppState>() {
+                            log::info!("Starting database cleanup...");
+                            if let Err(e) = app_state.db_manager.cleanup().await {
+                                log::error!("Failed to cleanup database: {}", e);
+                            } else {
+                                log::info!("Database cleanup completed successfully");
+                            }
                         } else {
-                            log::info!("Database cleanup completed successfully");
+                            log::warn!("AppState not available for database cleanup (likely first launch)");
                         }
-                    } else {
-                        log::warn!("AppState not available for database cleanup (likely first launch)");
-                    }
 
-                    // Clean up sidecar
-                    log::info!("Cleaning up sidecar...");
-                    if let Err(e) = summary::summary_engine::force_shutdown_sidecar().await {
-                        log::error!("Failed to force shutdown sidecar: {}", e);
-                    }
-                });
-                log::info!("Application cleanup complete");
+                        // Clean up sidecar
+                        log::info!("Cleaning up sidecar...");
+                        if let Err(e) = summary::summary_engine::force_shutdown_sidecar().await {
+                            log::error!("Failed to force shutdown sidecar: {}", e);
+                        }
+                    });
+                    log::info!("Application cleanup complete");
+                }
+                _ => {}
             }
         });
 }
--- a/frontend/src-tauri/src/summary/processor.rs
+++ b/frontend/src-tauri/src/summary/processor.rs
@@ -12,6 +12,9 @@ static THINKING_TAG_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(r"(?s)<think(?:ing)?>.*?</think(?:ing)?>").unwrap()
 });
 
+const ENGLISH_BASE_SUMMARY_INSTRUCTION: &str =
+    "Write the summary/report in English regardless of transcript language; non-English prose is invalid.";
+
 fn resolve_cached_english<'a>(
     cached: Option<&'a str>,
     summary_language: Option<&str>,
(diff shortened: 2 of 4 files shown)


sujithatzackriya:
Thanks for the two fixes here, @safvanatzack. Solid, focused work. The macOS Dock reopen (Fix 2) verified fully clean: `RunEvent::Reopen { .. }` is valid on the pinned Tauri version, the `match` refactor preserves Exit cleanup byte for byte, no previously handled event is dropped, and `focus_main_window` correctly covers minimized/hidden/closed-to-tray. Nothing needed there.

For Fix 1 (English base stabilization), all seven intents are delivered and there is no correctness, data, or security defect: cache wiring and cancellation safety both verified correct. A few items below by priority. Could you validate the P0 items and run a manual test before we merge? Priority key: P0 gates merge, P1 fix in this PR or immediate follow-up, P2 backlog.

## P0 (please validate before merge)

**P0-1 Normalization over-fires for default users (Medium)**
`resolve_final_language_action` (processor.rs:36-46) keys its outer match only on the selected language, so `summary_language == None` (the default majority who never picked a language) lands in the same arm as explicit English and triggers a full extra LLM pass on any transcript not positively detected as English (including `detected == None` for short or low-confidence audio). Base did `english_markdown.clone()` with zero extra work. This is a cost and latency regression (roughly 1.5x to 2x the final stage) for the largest user segment, behind a generic spinner. Output stays English, so this is not a correctness bug, but it exceeds the PR's stated scope (the description scopes normalization to English-selected targets).
Suggested: route `detected == None` to `ReturnEnglish`, reserving `NormalizeEnglish` for positively detected non-English transcripts. If broad normalization is intended, please confirm the cost tradeoff and add a distinct "Finalizing English summary" status so the extra wait is not read as a hang.

**P0-2 The changed behavior is untested (Medium)**
All four `resolve_final_language_action` tests use `Some("en")` / `Some("fr")`. The exact cells whose semantics changed, `(None, None)` and `(None, "ja")`, have no regression guard, so the riskiest change is invisible to CI. Please add `assert_eq!` cases for both, matching whatever P0-1 decides. Pure function, deterministic, zero risk.

## P1 (fix here or immediate follow-up)

**P1-1 Cache mutation correct but unpinned (Low)**
The post-normalization `english_markdown` (processor.rs:572 -> 579 -> service.rs:557-562) is correctly what gets cached. Confirmed by reading, but no test pins it; a refactor dropping the reassignment would cache non-English prose with a green suite. Add an assertion that the returned English element equals the normalized text when `NormalizeEnglish` fires (bundle with P0-2).

**P1-2 service.rs detection half has zero tests (Low)**
`read_detected_summary_language`, `detect_summary_language_from_text`, and the metadata-over-heuristic precedence (service.rs:133-134, 228-271) are untested. Add one test for `detect_summary_language_from_text` on a known non-English string and one asserting metadata takes precedence over the heuristic.

**P1-3 Detection computed then discarded for translation targets (Low)**
For every translation-target summary lacking a metadata file, the fallback runs a full-transcript language scan plus a full string clone and metadata read, then the Translate arm ignores it (service.rs:446-449 vs processor.rs:41). Suggested: compute `detected_summary_language` only when the target is English or None. If P0-1's narrowing lands, this cost is largely avoided too.

## P2 (backlog, safe to defer)

**P2-1 Prompt change does not bump the cache fingerprint (Low)**
`build_summary_cache_source` (service.rs:94-121) keys on transcript/template/model, not prompt text. Pre-PR caches stay reuse-eligible and can feed the translate path without the new English enforcement. Bounded to the translate path. Policy call: bump the fingerprint (one-time recompute) or document the invariant.

**P2-2 Cancellation detection by substring match (Low, pre-existing)**
Cancellation is distinguished by `e.contains("cancelled")` (processor.rs:66, service.rs:587). The contract holds today, but a provider error body containing that word would be misclassified as a user cancel. Follow-up: a typed `Cancelled` variant or `token.is_cancelled()` re-checks. Too broad for a hotfix.

**P2-3 Soft-fallback can silently emit non-English output (Low, intentional tradeoff)**
On a non-cancel normalization failure the code returns pass-1 markdown (intended). Pass-1 already enforces English, but if it leaked non-English and normalization then fails, output may be non-English with only an `error!` log. Strictly better than base. Optionally record "normalization skipped" in result metadata.

## Request
Could you please:
1. Make the P0-1 policy call and apply the chosen behavior.
2. Add the P0-2 test (and ideally P1-1).
3. Run a manual end-to-end test: a non-English meeting with no language selected, a non-English meeting with English selected, and a translation target, confirming output language and the extra-pass timing.

Thanks again for the work here. Happy to pair on the `resolve_final_language_action` narrowing if useful.

safvanatzack:
Thanks for the detailed review. I validated the items against the current PR branch and traced the frontend/backend flow.

My read is that there is no merge-blocking user impact here.

| Item | Validation | Actual user impact |
|---|---|---|
| P0-1 normalization over-fires | Partly valid, but narrower than stated. The frontend normally resolves Auto to a per-meeting language, cached detected language, or fresh detected language before calling the backend. Backend `summary_language == None` is mostly a fallback/detection-failure path, not the default majority path. | Possible extra LLM normalization pass in fallback cases, causing extra latency/cost. Output remains English or best-effort English; no correctness/data/security issue found. |
| P0-2 missing `(None, None)` / `(None, "ja")` tests | Valid test gap. | No current user impact; only regression coverage. |
| P1-1 cache mutation unpinned | Current code is correct: normalized English is reassigned before being returned/cached. | No current user impact. Test would only protect future refactors. |
| P1-2 service detection tests | Partly valid. Lower-level detection and metadata paths are tested; service-level precedence could use a small test. | No current user impact found. |
| P1-3 detection computed for translation targets | Valid minor inefficiency. | Small local CPU/string work before translation; not an extra LLM pass and unlikely to be noticeable. |
| P2 items | Valid as follow-up/backlog observations. | Bounded edge cases or pre-existing behavior; not merge-blocking. |

So the only real user-facing risk I found is: in fallback cases where no summary language reaches the backend and transcript language is non-English/unknown, summary generation may take longer because of an extra English-normalization pass. I do not see evidence that this affects the majority Auto/default path, and I did not find a correctness, data loss, or security issue.

I’m okay adding the missing `None` matrix tests as a follow-up/quick hardening step, but I don’t think these findings should block merge.

PR #499, opened 2026-06-04
feat: local HTTP control API for external recording start/stop

ryanhugh:
## What

Adds a **localhost-only HTTP control API** (`127.0.0.1:5172`) inside the Tauri app so other local processes (scripts, Raycast, window managers, automations) can control recordings:

| Endpoint | Description |
|---|---|
| `GET /status` | `{"recording": bool}` |
| `POST /start` | Body `{"title"?: string, "metadata"?: string}` — starts a recording. Optional custom meeting title; optional free-form metadata block that is saved as the **first transcript segment** (top of the meeting, included in summarization). Returns `200` once recording actually starts, `409` if already recording, `504` if the frontend refuses (e.g. no transcription model). |
| `POST /stop` | Stops the recording and runs the normal save flow (transcription flush → SQLite save → navigation). `409` if not recording. |

Example:

```bash
curl -X POST http://127.0.0.1:5172/start \
  -H 'content-type: application/json' \
  -d '{"title": "Standup", "metadata": "Attendees: Ryan, Alice\nTicket: ENG-123"}'

curl -X POST http://127.0.0.1:5172/stop
```

## How

The API deliberately reuses the existing tray mechanisms rather than inventing a new path:

- **Start** stages parameters in `sessionStorage` and triggers the same auto-start flow the tray uses (`useRecordingStart`), so model checks, device selection, meeting state, and save flow behave exactly like a button click.
- **Stop** uses a new shared `recording_commands::stop_recording_and_notify()` helper — both tray handlers were duplicating ~30 lines of this logic and now call the same helper.
- Metadata is injected as a synthetic transcript segment (`sequence_id: -1`, sorts before all real segments) both live and at save time, so it survives mid-recording page reloads and state resyncs.
- Server is `axum` on the existing tokio runtime, bound to `127.0.0.1` only (same trust level as the tray). Bind failure logs and degrades gracefully.

## Pre-existing bugs fixed along the way

1. **Selected devices and meeting name were silently dropped on every recording start**: `recordingService.startRecordingWithDevices` passed snake_case arg keys (`mic_device_name`, `meeting_name`) to a Tauri v2 command expecting camelCase, so Rust always received `None` (log: `Mic: None, System: None, Meeting: None`) and every meeting got the default `Meeting YYYY-MM-DD…` title regardless of the UI-generated one.
2. **Clean builds of `main` fail with the committed Cargo.lock** (tauri 2.11.2): `generate_handler!` requires the generated `__tauri_command_name_*` macros, but `summary/mod.rs` and `summary_engine/mod.rs` re-exported only the old `__cmd__*` ones → 15 × `E0433`. Added the missing re-exports.

## Testing

Verified end-to-end on macOS (Apple Silicon, dev build):

- `GET /status` → `{"recording":false}` / `{"recording":true}` ✅
- `POST /start` with title + metadata → app navigates home, recording starts, `200 {"status":"recording","title":"API Test Round 2"}`; Rust log confirms `Meeting: Some("API Test Round 2")` ✅
- `POST /start` while recording → `409` ✅ ; `POST /stop` while idle → `409` ✅
- Spoke during recording, then `POST /stop` → normal save flow; SQLite shows the meeting titled `API Test Round 2` with the metadata block as the first segment (`audio_start_time = 0`) followed by the real transcribed speech ✅
- `cargo check` and `tsc --noEmit` clean ✅

### DIFF
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -111,6 +111,9 @@ tokio = { version = "1.32.0", features = ["full", "tracing"] }
 tokio-util = "0.7"  # Utilities for tokio including CancellationToken
 async-trait = "0.1"  # Trait abstraction for async methods
 
+# Local HTTP control API (127.0.0.1 start/stop endpoints)
+axum = "0.8"
+
 reqwest = { version = "0.11", features = ["blocking", "multipart", "json", "stream"] }
 
 # crossbeam
--- a/frontend/src-tauri/src/audio/recording_commands.rs
+++ b/frontend/src-tauri/src/audio/recording_commands.rs
@@ -890,6 +890,35 @@ pub async fn stop_recording<R: Runtime>(
     Ok(())
 }
 
+/// Stop recording from a Rust-initiated trigger (tray menu, local API) and notify
+/// the frontend to run post-stop processing (transcription wait, SQLite save, navigation).
+pub async fn stop_recording_and_notify<R: Runtime>(app: AppHandle<R>) -> Result<(), String> {
+    // Generate save path (same as RecordingControls.tsx)
+    let data_dir = app
+        .path()
+        .app_data_dir()
+        .map_err(|e| format!("Failed to get app data dir: {}", e))?;
+
+    let timestamp = chrono::Local::now().format("%Y-%m-%dT%H-%M-%S").to_string();
+    let save_path = data_dir.join(format!("recording-{}.wav", timestamp));
+
+    stop_recording(
+        app.clone(),
+        RecordingArgs {
+            save_path: save_path.to_string_lossy().to_string(),
+        },
+    )
+    .await?;
+
+    // Trigger frontend post-processing via event (works from any page)
+    // (SQLite save, navigation, analytics)
+    if let Err(e) = app.emit("recording-stop-complete", true) {
+        error!("Failed to emit recording-stop-complete event: {}", e);
+    }
+
+    Ok(())
+}
+
 /// Check if recording is active
 pub async fn is_recording() -> bool {
     IS_RECORDING.load(Ordering::SeqCst)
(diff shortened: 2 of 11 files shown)


rajpratham1 (changes requested):
This is a useful feature (local HTTP control API), but it introduces a new attack surface and mixes feature additions with bug fixes. I wouldn't approve it without addressing a few issues.

What's good
Nice reuse of stop_recording_and_notify() instead of duplicating logic.
Good separation into local_api.rs.
/status, /start, /stop are simple and easy to understand.
Waiting until recording actually starts instead of immediately returning success is good UX.
Metadata preservation across reloads is thoughtful.
Fixing camelCase invoke arguments is also a legitimate bug fix.
Concerns
1. No authentication (Major)

The server listens on

127.0.0.1:5172

Any local process can execute

POST /start
POST /stop

That includes

browsers
local scripts
Electron apps
malware
browser extensions

Localhost is not an authentication mechanism.

At minimum I'd expect:

bearer token
shared secret
random startup token
origin validation

Otherwise any process on the machine can control recording.

2. Feature PR mixes unrelated fixes

This PR contains:

Local API
invoke camelCase fix
transcript metadata
recording service fixes
summary exports
tray refactor

These should ideally be separate PRs.

Large mixed PRs are harder to review.

3. Busy polling
sleep(250ms)

is_recording()

for 15 seconds.

Would be cleaner to wait on an event/channel/Notify instead of polling.

Not blocking approval by itself, but worth improving.

4. JavaScript injection

This is safe:

serde_json::to_string(...)

before

window.eval(...)

because values are escaped.

Still, if possible I'd prefer using Tauri events instead of eval().

5. Hardcoded port
127.0.0.1:5172

Potential conflicts.

Better:

configurable
fallback if occupied
ephemeral port
6. Missing tests

This feature needs integration tests covering:

start endpoint
stop endpoint
timeout path
already recording
metadata persistence
invalid requests

Currently there are none.

7. Security documentation

Because this exposes an HTTP API, documentation should clearly state:

localhost only
intended clients
security assumptions
authentication (if added)
Minor observations

The extraction of

stop_recording_and_notify()

is a good refactor.

The camelCase fix in recordingService.ts also fixes a real Tauri v2 interoperability issue.

PR #500, opened 2026-06-04
fix(recovery): show recovery dialog on first relaunch after a crash (R-1)

athulchandroth:
## Summary

Fixes **R-1**: after fully quitting the app (Cmd+Q / menu / tray Quit) or crashing **while recording**, the recovery dialog sometimes did **not** appear on the first relaunch — only on a *second* quit + relaunch. The lost session was never lost, just surfaced one launch late.

## Root cause

- A recording's meeting is seeded in IndexedDB with `lastUpdated = now`, and **every transcript bumps `lastUpdated` to `now`** (`TranscriptContext.tsx`, `indexedDBService.ts`). So a crash's `lastUpdated` ≈ the moment you quit.
- The recovery scan **dropped any meeting newer than 15s** (`useTranscriptRecovery.ts`). Relaunch within ~15s of the crash → the just-crashed meeting was filtered out.
- The scan **runs once, on mount only** — no polling/re-scan — so it never recovered within that session. The next launch (now >15s later) passed the filter and showed it.

Not macOS-specific — it's shared TS timing logic; it only *looked* platform-specific because reproduction depends on whether relaunch lands inside the 15s window.

History note: the 15s filter began at 5s (`1101d0a`), grew to 15s + a precise id-based guard (`3dc625b`), then the id guard was stripped in a cleanup (`24c9bc8`), leaving only the blunt time heuristic that caused this.

## Fix

- **Remove the 15s freshness filter.** Rely on `savedToSQLite === false` + the 7-day retention window — matching **Meetily Pro**, whose recovery path has no such filter.
- **Add an id-based insurance guard:** skip the meeting whose id equals `sessionStorage['indexeddb_current_meeting_id']` (the live / mid-save current session). It's cleared on save and wiped on crash/quit, so an interrupted session has no active id on relaunch and surfaces immediately — but the in-flight session is never shown mid-save.

## Why it's safe

- **Normal stop:** `markMeetingAsSaved()` (savedToSQLite=true) is awaited **before** the stop flow leaves `SAVING`, and the scan is skipped during `STOPPING/PROCESSING/SAVING` — so the just-stopped meeting is already excluded when the scan can next run.
- **Crash:** never marked saved, no active id on relaunch → surfaces on the **first** relaunch.
- **Save failure:** the active id stays set, so the unsaved meeting is suppressed for that session and surfaces next launch (matches pre-fix outcome — not a regression).

## Testing

- `tsc --noEmit` clean.
- Static trace of stop / crash / save-failure flows (see `docs/AIDocs/github-issue-next-release-findings.md`, R-1).

## Files

- `frontend/src/hooks/useTranscriptRecovery.ts` — the fix
- `docs/AIDocs/github-issue-next-release-findings.md` — R-1 root cause + fix writeup

### DIFF
--- a/docs/AIDocs/github-issue-next-release-findings.md
+++ b/docs/AIDocs/github-issue-next-release-findings.md
@@ -1,0 +1,170 @@
+# Next-release manual test: 4 findings (1 blocker)
+
+Manual testing of the `devtest` branch (`bc55119`) on macOS surfaced four issues. The first is a
+release blocker; the rest are worth fixing. File/line references are against `devtest`.
+
+| ID | Severity | Area | Summary |
+|----|----------|------|---------|
+| S-1 | **High** | Multilingual summary | Summary stays in transcript's source language when a different target is selected |
+| R-1 | Medium | Recovery / app lifecycle | Recovery dialog intermittently missed on the first relaunch after a crash/quit during recording (15s freshness filter + run-once scan) |
+| M-1 | Medium | Privacy / analytics | User agent still transmitted via `os_version` |
+| U-1 | Low | UX | Pinned summary-language default not signposted as new-meeting-only |
+
+**Environment:** `devtest` (`bc55119`), macOS (Metal), local model. Analytics enabled only where noted.
+
+---
+
+## S-1 [High] — Summary stays in the transcript's source language when a different target is selected
+
+**Describe the bug**
+When the transcript is non-English, the generated summary comes out in the transcript's **source
+language** even when a different summary language is explicitly selected.
+Example: **Japanese** transcript + summary language set to **Hindi** → summary generated in
+**Japanese**.
+
+**Steps to reproduce**
+1. Open a meeting with a non-English transcript (e.g. Japanese).
+2. Set the summary language to **Hindi** (or even **English**).
+3. Generate the summary.
+
+Observed: summary is in Japanese. Expected: summary in the selected target language.
+
+**Root cause**
+The two-pass pipeline assumes pass-1 produces an English base, but the **pass-1 prompts never
+instruct the model to write in English** (`frontend/src-tauri/src/summary/processor.rs` — chunk
+prompt ~L298, final prompt ~L395–411). For a non-English transcript the model summarizes in the
+source language, and that output is labeled `english_markdown` / cached as `cached_english`.
+Translation only runs when the target ≠ English (`processor.rs:455`):
+- target = **English** on a non-English transcript → no translation → source-language summary returned.
+- target = **non-English** (e.g. Hindi) → `translate_markdown` is fed source-language text instead
+  of English; a local model may echo the source back (returns `Ok`, so no error surfaces).
+
+This also undermines the "cache English summary" optimization — the cached base isn't actually
+English for non-English transcripts.
+
+**Suggested fix**
+Force pass-1 output to English (add an explicit "write the report in English regardless of the
+transcript's language" instruction to the pass-1 prompts) so pass-2 is reliably English→target —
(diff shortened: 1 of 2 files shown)


safvanatzack:
@athulchandroth 


#### MED-1 - Same-session recovery is suppressed after a save failure

- **Severity**: Medium
- **Category**: Regression / UX
- **Location**: `frontend/src/hooks/useTranscriptRecovery.ts:65` (related: `frontend/src/hooks/useRecordingStop.ts:398`, `frontend/src/app/page.tsx:76`, `frontend/src/services/indexedDBService.ts:154`)
- **Issue**: The new active-id guard removes the unsaved current meeting from recovery results for the rest of the renderer session after a SQLite save failure. The data is still in IndexedDB, so this is not data loss, but the user loses same-session recovery in cases where the old branch could have shown it.
- **Root cause**: PR #500 replaces the time-based "older than 15s" guard with `m.meetingId !== activeMeetingId`, but `activeMeetingId` is only cleared on successful save. In the save-failure catch, `storageService.saveMeeting()` throws before `markMeetingAsSaved()` and before `sessionStorage.removeItem('indexeddb_current_meeting_id')`, so the recovery scan filters out the exact unsaved meeting returned by `getAllMeetings()`.
- **Call path / reachability**: User stops recording -> `useRecordingStop.handleRecordingStop()` enters `SAVING` -> `storageService.saveMeeting()` throws -> `useRecordingStop.ts:398-404` sets `RecordingStatus.ERROR` without clearing `indexeddb_current_meeting_id` -> Home's startup/status effect runs because `ERROR` is not skipped (`page.tsx:76-100`) -> `checkForRecoverableTranscripts()` calls `indexedDBService.getAllMeetings()` -> unsaved meeting is returned because `savedToSQLite === false` (`indexedDBService.ts:153-158`) -> PR line `useTranscriptRecovery.ts:65-68` filters it out because the meeting id equals `activeMeetingId`.
- **Reproduction**:
  1. Pre-state: A recording has produced transcripts, and its last IndexedDB transcript update is more than 15s old by the time save starts.
  2. Action: Stop the recording and make `storageService.saveMeeting()` fail, for example by forcing a backend database save error.
  3. Expected: The unsaved IndexedDB meeting is offered for recovery once the stop flow reaches an error state.
  4. Actual: No recovery dialog appears in the same session because the meeting id still matches `sessionStorage['indexeddb_current_meeting_id']`.
  5. Code path: `frontend/src/hooks/useTranscriptRecovery.ts:65-68`.
  6. Affected users: Users whose normal stop save fails after the old 15s freshness window would already have elapsed.
- **Impact & severity**: Medium UX/recoverability regression. The meeting is preserved and will surface after relaunch because `sessionStorage` clears, but same-session rescue is blocked.
- **Regression risk**: The base branch also hid very fresh failures, but once `lastUpdated < Date.now() - 15000`, it no longer suppressed the meeting. PR #500 suppresses it indefinitely within that session as long as the active id remains set.
- **Steel-man check**: The author's strongest case is that the active id intentionally suppresses the current meeting during an in-flight save and avoids a jarring recovery prompt immediately after a save-error toast. That safety goal is valid, but it is broader than the original 15s heuristic and blocks recovery even after the save has definitively failed.
- **Suggestion**: On save failure, either clear or explicitly mark the active id as no longer in-flight before Home can rescan, or change the recovery guard to skip the active id only while status is `STOPPING`, `PROCESSING_TRANSCRIPTS`, or `SAVING`. Keep the crash path unaffected by leaving `sessionStorage` non-persistent across relaunch.

safvanatzack:
The supposed fix is already got merged in other branch as you suggested @athulchandroth 
Closing this one

PR #510, opened 2026-06-09
ui: clarify ScreenCaptureKit as coming soon instead of silently disabled

### LINKED ISSUE #509: [BUG] ScreenCaptureKit backend incorrectly disabled in UI (AudioBackendSelector.tsx)
## Bug Description

ScreenCaptureKit is always shown as **Disabled** in the Settings UI, even when the user has granted the required macOS permission (*System Settings → Privacy & Security → Screen and System Audio Recording*).

## Root Cause

In `frontend/src/components/AudioBackendSelector.tsx`, the condition that disables a backend has a variable naming error:

```typescript
// Line ~96
const isCoreAudio = backend.id === 'screencapturekit'; // ← wrong: checks screencapturekit, not coreaudio
const isDisabled = disabled || isCoreAudio;
```

The variable is named `isCoreAudio` but checks for `'screencapturekit'` — so ScreenCaptureKit ends up being the one disabled, not Core Audio.

## Fix

Change the condition to check the correct backend ID:

```typescript
const isCoreAudio = backend.id === 'coreaudio';
const isDisabled = disabled || isCoreAudio;
```

Or, if the intent is to allow both backends to be selectable, simply remove the forced disable logic entirely:

```typescript
const isDisabled = disabled;
```

## Steps to Reproduce

1. Open Meetily → Settings → Audio Devices
2. Scroll to **System Audio Backend**
3. ScreenCaptureKit is shown with a **Disabled** badge and cannot be selected
4. This happens regardless of macOS permissions

## Environment

- macOS (Apple Silicon)
- Meetily v0.4.0
- Permission granted: *Screen and System Audio Recording → Meetily → enabled*

## Impact

Users who want to use ScreenCaptureKit for direct digital audio capture (without relying on a physical audio output) cannot do so. This is particularly relevant for users on Bluetooth headsets where Core Audio may not capture other participants' audio correctly.

Yogau:
## Summary

- Replaces the silent `Disabled` badge on ScreenCaptureKit with a `Coming soon` badge in amber
- Adds a tooltip explaining the option is under development and not yet available
- No functional change — ScreenCaptureKit remains disabled until the backend is fully implemented

This follows feedback from @sujithatzackriya confirming that ScreenCaptureKit is intentionally disabled due to instability and incomplete backend implementation.

## Test plan

- [ ] Open Settings → Audio Devices → System Audio Backend
- [ ] Verify ScreenCaptureKit shows a "Coming soon" badge instead of "Disabled"
- [ ] Verify hovering the badge shows the tooltip: "This backend is under development and will be available in a future release."
- [ ] Verify Core Audio remains selectable and active by default

Fixes #509

### DIFF
--- a/frontend/src/components/AudioBackendSelector.tsx
+++ b/frontend/src/components/AudioBackendSelector.tsx
@@ -129,7 +129,8 @@ export function AudioBackendSelector({
 
       <div className="space-y-2">
         {backends.map((backend) => {
-          const isDisabled = disabled;
+          const isComingSoon = backend.id === 'screencapturekit';
+          const isDisabled = disabled || isComingSoon;
 
           return (
             <label
@@ -157,6 +158,11 @@ export function AudioBackendSelector({
                   {currentBackend === backend.id && (
                     <span className="text-xs font-medium text-blue-600 bg-blue-100 px-2 py-0.5 rounded">
                       Active
+                    </span>
+                  )}
+                  {isComingSoon && (
+                    <span title="This backend is under development and will be available in a future release." className="text-xs font-medium text-amber-600 bg-amber-50 border border-amber-200 px-2 py-0.5 rounded cursor-help">
+                      Coming soon
                     </span>
                   )}
                 </div>
--- a/frontend/src/components/AudioBackendSelector.tsx
+++ b/frontend/src/components/AudioBackendSelector.tsx
@@ -129,9 +129,7 @@ export function AudioBackendSelector({
 
       <div className="space-y-2">
         {backends.map((backend) => {
-          // Disable Core Audio option
-          const isCoreAudio = backend.id === 'screencapturekit';
-          const isDisabled = disabled || isCoreAudio;
+          const isDisabled = disabled;
 
           return (
             <label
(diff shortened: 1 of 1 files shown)


sujithatzackriya:
Hey @Yogau 
Thank you for the contribution and clear communication. 

@safvanatzack can we review and merge this to devtest? Please let me know.

rajpratham1 (changes requested):
This PR is only 11 changed lines, but it introduces a functional inconsistency.

What's good
✅ Makes the UI label clearer by changing "Disabled" to "Coming soon".
✅ Adds a tooltip explaining why the option isn't available.
✅ Uses amber styling instead of a generic gray badge, which better communicates the state.
Blocking issue

The PR title is:

fix: enable ScreenCaptureKit backend selection in UI

But the code does not enable the backend.

It still contains:

const isComingSoon = backend.id === "screencapturekit";
const isDisabled = disabled || isComingSoon;

Since isDisabled remains true for screencapturekit, users still cannot select it.

So this PR changes only the wording/UI, not the functionality promised by the title.

Questions for the author

Before approving, I'd ask:

Is ScreenCaptureKit actually implemented and ready?
If yes, why is it still disabled in the UI?
If not, why is the PR titled "enable ScreenCaptureKit backend selection" instead of something like:
ui: clarify ScreenCaptureKit as coming soon
fix(ui): improve ScreenCaptureKit disabled state
Recommendation
Request Changes if the intention is truly to enable the backend, because this PR doesn't accomplish that.
Approve only if the title and description are updated to reflect that this is just a UI messaging change rather than enabling the feature.

Yogau (commented):
Thanks for the thorough review — you're right, the title was misleading. Updated it to reflect that this is purely a UI messaging change, not a functional enable. The backend stays disabled as discussed with the maintainers.

Issue #519, opened 2026-06-17
[FEATURE] Disable real-time transcription


replicant026:
## Feature Description
Add an option to disable real-time transcription during recording.

## Problem Statement
I really like the program and it has been very useful for me. However, I use it on a computer with limited hardware resources, and depending on what else I am running, the system can become slow.

Real-time transcription also does not work very well for my workflow. I usually prefer to finish the recording first and then use Whisper Turbo afterward to generate the transcription with better performance and reliability.

## Proposed Solution
Add a setting that allows users to disable real-time transcription while recording.

When this option is enabled, the program would record normally but would not attempt to transcribe audio in real time. After the recording is finished, the user could still generate the transcription manually using Whisper Turbo or another post-processing transcription option.

This could be implemented as a toggle in the settings, for example:

“Enable real-time transcription”
or
“Disable real-time transcription during recording”

## User Story
As a user with limited computer resources,
I want to disable real-time transcription during recording,
So that the program runs more smoothly and I can generate the transcription afterward using Whisper Turbo.

## Acceptance Criteria
- [ ] There is an option in the settings to enable or disable real-time transcription.
- [ ] When real-time transcription is disabled, recording continues to work normally.
- [ ] When real-time transcription is disabled, the program does not process transcription during the recording.
- [ ] After the recording ends, the user can still generate a transcription manually.
- [ ] The selected setting is saved and reused in future sessions.

## Technical Considerations
Real-time transcription can be resource-intensive, especially on lower-end computers. Disabling it could reduce CPU/GPU usage, improve recording stability, and prevent slowdowns while the user is recording.

The feature could be implemented as a simple configuration flag that prevents the transcription pipeline from running during active recording, while keeping the post-recording transcription workflow available.

## Alternatives Considered
An alternative would be to reduce the quality or frequency of real-time transcription updates, but this would still consume system resources during recording.

Another alternative would be to keep real-time transcription enabled by default but allow users to select a lighter transcription model. However, for some users, the best option is to completely disable real-time transcription and process everything after recording.

## Additional Context
This would be especially useful for users running the program on computers with limited hardware resources or users who prefer post-processing transcription with Whisper Turbo after the recording is finished.

## Checklist
- [x ] I have searched for similar feature requests
- [x ] I have provided all required information
- [ ] I have included any relevant screenshots/mockups
- [x ] I have described the problem and proposed solution clearly

replicant026:
[...]

Issue #335, opened 2026-02-05
[FEATURE] VibeVoice-ASR for speaker diarization and domain-specific terminology


TanaroSch:
## Feature Description

Integrate **Microsoft VibeVoice ASR** as an optional transcription engine to provide unified "Rich Transcription," including native speaker diarization and domain-specific terminology support.

## Problem Statement

Whisper and Parakeet don't support speaker diarization and customized context, which can lead to alignment errors and "speaker drift" in long meetings. Furthermore, transcribing niche technical or industry-specific terms accurately remains a challenge during meetings.

## Proposed Solution

Leverage VibeVoice’s LLM-based architecture to:

1. Perform ASR and Diarization in a **single pass** for improved "who-said-what" accuracy.
2. Utilize its **Context Prompting** feature to inject meeting-specific keywords (project names, technical jargon) directly into the inference process.

## User Story

As a **meeting participant**,
I want **highly accurate transcripts that correctly identify speakers and technical terms**,
So that **my meeting minutes are actionable and require minimal manual correction.**

## Acceptance Criteria

* [ ] Option to select "VibeVoice" as the ASR provider in settings.
* [ ] Support for **Context Prompts** (allowing users to input meeting-specific terminology before processing).
* [ ] Speaker labels in the generated output (with manual override from speaker1 to Firstname Lastname).

## Technical Considerations

* **Hardware:** VibeVoice uses a 7B parameter LLM backbone (Qwen 2.5), requiring more VRAM/compute than Whisper and Parakeet (though 4bit-quants already exist on Huggingface).
* **Latency:** It is slower than TDT-based models; it is best suited for asynchronous "post-meeting" processing rather than real-time captions.
* **Language Support:** Currently supports 50+ languages.

## Alternatives Considered

* I don't know of any other ASR models that support both of those features.

## Additional Context

Reference: [Microsoft Research VibeVoice](https://github.com/microsoft/VibeVoice).

## Checklist

* [x] I have searched for similar feature requests
* [x] I have provided all required information
* [x] I have included any relevant screenshots/mockups
* [x] I have described the problem and proposed solution clearly

dpbattaglia-gps:
Microsoft VibeVoice-ASR is fantastic and a game changer for us. Definitely needs to be on a powerful GPU based server off the main meeting machine. So post transcription service (i.e. asynchronous) would be ideal and also requires the ability to choose an external transcription server.

sujithatzackriya:
[...]

Issue #507, opened 2026-06-08
[FEATURE]  Enable to REtranscribe with COSTOM ENDPOINT.


DominikRoehr:
## Feature Description
Implement integration for external transcription models, similar to the existing summary model integration. This will allow users to utilize external services, such as Whisper running on Ollama, for post-processing retranscription.

## Problem Statement
Currently, the application is limited to using only the internally integrated transcription model, preventing users from leveraging more powerful external inference servers.

## Proposed Solution
Implement an integration pattern analogous to the current summary model integration.
*(Interim Solution: Enable manual transcription processing and allow users to import the resulting transcript files.)*

## User Story
As a user with a local LLM inference server but limited on-device GPU resources,
I want to utilize my external server to enhance transcription during post-processing,
So that I can achieve higher transcript quality without overtaxing my local CPU/GPU.

While the integrated Parakeet model is sufficient for live transcription, users often require higher-fidelity results for post-processing tasks.

## Acceptance Criteria
- Support for OpenAI-compatible APIs.

## Technical Considerations
- Audio file transfer latency and bandwidth may vary significantly depending on the user's network configuration and server setup.

## Alternatives Considered
- **Interim Solution:** Allow users to perform transcription manually via external tools and provide a feature to import the completed transcript into the application.

## Checklist
- [x] I have searched for similar feature requests.
- [x] I have provided all required information.
- [ ] I have included any relevant screenshots/mockups.
- [x] I have described the problem and proposed solution clearly.

Hus-Mek:
Related building block: open PR **#528** adds a vendor-neutral HTTPS remote ASR provider, which lets transcription be pointed at an external endpoint (e.g. Whisper served elsewhere) rather than running only in-process. **#610** surfaces it in Settings and **#533** documents the contract. That's the plumbing this request needs; wiring it into the retranscribe flow specifically would build on top of it.

Issue #338, opened 2026-02-07
[FEATURE] Disable Transcription


dpbattaglia-gps:
## Feature Description
I would like to be able to disable automatic transcription and utilise the software just for audio capture for later transcription.

## Problem Statement
For some computers that are really slow and buggy - I am finding issues within meetings where the meeting quality is degrading due to ai local processing. 

## Proposed Solution
disable transcription and summary so all that is left is the audio recorded - lowest system resources... then I can take the audio file and later transcribe and summarise

## Checklist
- [x] I have searched for similar feature requests
- [x] I have provided all required information
- [ ] I have included any relevant screenshots/mockups
- [x] I have described the problem and proposed solution clearly

manzke:
👍

dpbattaglia-gps:
Hey @sujithatzackriya, 

Any chance this could be a quick implementation? Would love to use meetily within our automation pipeline but need transcription disabled for some employee laptops to be useable.

sujithatzackriya:
This is not a quick implementation @dpbattaglia-gps but we sre exploring this option.

MichaelRosenlund:
I too, would love this feature. I dont find myself using the live transcript at all and would like the option to use app to record and post-process.

AlexMili:
Would love this feature too !

dpbattaglia-gps:
Any update on this?

DPBattaglia:
@sujithatzackriya - @Hus-Mek has provided https://github.com/Zackriya-Solutions/meetily/pull/532 that will provide the functionality - please commit asap.

Regards,
David

PR #528, opened 2026-06-24
feat(transcription): pluggable HTTPS-backed remote ASR provider

### LINKED ISSUE #527: [FEATURE] pluggable HTTPS-backed remote ASR provider (relates to #335)
## Feature Description

Surface existing `TranscriptionProvider` trait via a small, vendor-neutral HTTPS backend so users can plug meetily into any remote ASR (WhisperX-compatible handler on cloud GPU, self-hosted inference server, etc.) without vendoring a specific provider.

## Problem Statement

`TranscriptionProvider` (in `frontend/src-tauri/src/audio/transcription/provider.rs`) already has `Whisper` and `Parakeet` impls. There is no path for users who need features neither provides — notably speaker diarization, code-switched Arabic/English, or meeting-keyword prompting (see #335). Adding a concrete impl like VibeVoice directly to the desktop app drags in a 7B-parameter LLM backbone that most users won't run.

A generic HTTPS-based provider surfaces that capability without baking any specific backend into the desktop binary: the user points meetily at an endpoint they already operate.

## Proposed Solution

1. Add a `RemoteProvider` impl of `TranscriptionProvider` that POSTs a base64-encoded mono-16kHz WAV to a configurable HTTPS endpoint and returns speaker-annotated segments (formatted as `SPEAKER_NN: text` lines — back-compat with the existing `TranscriptResult.text` field).
2. Surface it in **Settings → Transcription Provider** as an opt-in "Remote HTTPS" entry. Four fields: `endpoint_url`, `bearer_token`, `model id`, `default language`. **Default OFF** — `localWhisper` and `parakeet` paths unchanged.
3. Persist via the existing settings mechanism (no new DB column needed unless called for in review).
4. Add `docs/transcription-providers.md` describing the OpenAI-compatible JSON contract any worker must honor, plus a sample `docs/sample-remote-worker/handler.py` (~50 LoC) that establishes the contract.

This would land as 3 small PRs against `devtest`:
- **PR 1** (`feat/pluggable-remote-asr-provider`): `RemoteProvider` + tests, no UI. ~180 LoC + ~80 LoC tests.
- **PR 2** (`feat/transcription-provider-remote-ui`): Tauri commands, Settings UI dropdown + fields + "Test connection" button.
- **PR 3** (`enhance/remote-provider-docs`): docs + sample worker.

## User Story

As a **meetily user with multi-speaker meetings or non-English code-switching**,
I want **to plug meetily into a remote ASR endpoint I already operate**,
So that **I get speaker diarization and specialty transcription without the desktop binary shipping a heavyweight model runner**.

## Acceptance Criteria

- [ ] A new "Remote HTTPS" entry appears in Settings → Transcription Provider.
- [ ] When selected, the user can supply `endpoint_url`, `bearer_token`, `model id`, `default language`.
- [ ] A "Test connection" button POSTs a 5s silent WAV and reports latency + sample response.
- [ ] Default transcription provider remains `localWhisper` (no behavior change for current users).
- [ ] `docs/transcription-providers.md` documents the JSON contract any worker must honor.
- [ ] A sample worker (`docs/sample-remote-worker/handler.py`) ships with the docs.
- [ ] Provider contains zero vendor strings (no RunPod/Replicate/etc. naming in code or comments).
- [ ] Unit tests cover WAV header invariants, happy path, empty-segments path, and 4xx-error path.

## Technical Considerations

- **Hand-rolled WAV header** (44 bytes, RIFF/WAVE/fmt /data markers) to avoid pulling `wav` as a new crate. Sample depth stays 16-bit PCM mono at 16 kHz to match meetily's existing internal sample rate.
- **HTTP client:** `reqwest` is already in `Cargo.toml` (transitive via `tauri`). No new top-level dependency added.
- **Trait conformance:** the new provider plugs into `tauri::async_runtime` via the existing `Arc<dyn TranscriptionProvider>` pattern already used by `WhisperProvider` and `ParakeetProvider`.
- **Backward compatibility:** the existing `TranscriptResult.text` field receives `SPEAKER_00: ...` lines — no trait surface change needed. If maintainers would prefer a structured `Vec<SegmentAnnotation>` field, that's a separate cleanup PR.
- **CI cost:** the change is purely additive — feature-flagged by trait dispatch. `cargo check --features=cuda,vulkan` and `cargo test transcription::remote_provider` must run clean before any PR opens.

## Alternatives Considered

- **Vendoring VibeVoice directly in meetily (#335 proposal).** Would require bundling a 7B model runner into the desktop app. Most users would never run it; it bloats the binary by hundreds of MBs; it forces all users onto a single backend.
- **Adding per-vendor trait impls (e.g. `RunPodProvider`, `ReplicateProvider`).** Vendor lock-in — every new backend requires a new merge into meetily.
- **Streaming live chunks over HTTPS.** Deferred to a separate proposal. The trait already supports streaming via `Vec<f32>`; the HTTPS provider is intentionally request/response-only first.

## Additional Context

Closely related: #335. The proposed `RemoteProvider` makes that issue solvable without bundling VibeVoice into meetily itself — anyone running a VibeVoice-compatible worker (RunPod, self-hosted, etc.) can point meetily at it.

Happy to send the 3 PRs if the abstraction is welcome. Want to confirm shape + settings storage approach before any code lands.

## Checklist

Hus-Mek:
## Description
Adds a `RemoteProvider` impl of the existing `TranscriptionProvider` trait
(`frontend/src-tauri/src/audio/transcription/provider.rs`) that lets meetily
delegate transcription to any HTTPS endpoint that accepts a base64-encoded
mono-16kHz WAV and returns speaker-annotated segments.

Vendor-neutral: no RunPod/Replicate/AWS branding in code, comments, or
config. Users configure the endpoint URL, bearer token, model id, default
language, and speaker-count hints via Settings (PR #2 will wire the UI).
Defaults to the existing `localWhisper` provider — opt-in change.

Wire contract:
- POST {endpoint_url} `Content-Type: application/json`
- Optional `Authorization: Bearer ***` if a token is configured.
- Request body: `{ "audio_base64": "...", "model": "...", "language": "...",
  "min_speakers"?: u8, "max_speakers"?: u8 }`
- Response: `{ "segments": [{ start, end, text, speaker? }], "error"?: string }`
- 4xx/5xx -> `TranscriptionError::EngineFailed`.
- Worker `{ "error": "..." }` (HTTP 200) -> `EngineFailed`.

## Related Issue
Closes #527. Builds on the precedent set by #335 — gives meetily a path
to remote speaker diarization without bundling a 7B model runner into the
desktop binary.

## Type of Change
- [x] New feature
- [x] Code refactoring (extends the existing `TranscriptionProvider` trait surface)

## Testing
- [x] Unit tests added (11 inline in `remote_provider.rs`):
  - WAV header invariants (RIFF/WAVE/fmt /data markers; chunk sizes;
    empty-input edge case)
  - `format_segments` formatting (speaker-prefixed + plain + skip-empty)
  - Async end-to-end via an in-process TCP listener mock (no new mock crate)
    covering: happy path with two speakers, empty segments, HTTP 4xx/5xx,
    worker `{ error }` field, unparseable JSON, missing endpoint URL
- [ ] Manual testing performed (deferred until Settings UI lands in PR #2)
- [x] `cargo check` and `cargo test` pass (will run in CI; local host lacks
  libclang required by whisper-rs-sys bindgen, see maintainer-pre-flight notes)

## Documentation
- [x] Inline module-level doc comment documents the wire contract end-to-end
- [x] `ponytail:` markers call out deliberate simplifications
- [x] `docs/transcription-providers.md` to be expanded in PR #3

## Checklist
- [x] Code follows project style (mirrors `WhisperProvider` / `ParakeetProvider` conventions)
- [x] Self-reviewed
- [x] Comments on the HTTP contract
- [ ] README updated (covered in PR #3)
- [x] Branch is up to date with `devtest`
- [x] No merge conflicts
- [x] No vendor strings (`grep -ri 'runpod\|replicate\|mano' .` on the branch is empty)
- [x] No new top-level runtime dependency beyond `base64 = "0.22"` (already in
      the dep graph transitively; promoted to direct for clean import path)

### DIFF
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -112,6 +112,7 @@ tokio-util = "0.7"  # Utilities for tokio including CancellationToken
 async-trait = "0.1"  # Trait abstraction for async methods
 
 reqwest = { version = "0.11", features = ["blocking", "multipart", "json", "stream"] }
+base64 = "0.22"
 
 # crossbeam
 crossbeam = "0.8.4"
--- a/frontend/src-tauri/src/audio/transcription/mod.rs
+++ b/frontend/src-tauri/src/audio/transcription/mod.rs
@@ -5,6 +5,7 @@
 pub mod provider;
 pub mod whisper_provider;
 pub mod parakeet_provider;
+pub mod remote_provider;
 pub mod engine;
 pub mod worker;
 
@@ -12,6 +13,7 @@ pub mod worker;
 pub use provider::{TranscriptionError, TranscriptionProvider, TranscriptResult};
 pub use whisper_provider::WhisperProvider;
 pub use parakeet_provider::ParakeetProvider;
+pub use remote_provider::{RemoteProvider, RemoteProviderConfig};
 pub use engine::{
     TranscriptionEngine,
     validate_transcription_model_ready,
--- a/frontend/src-tauri/src/audio/transcription/remote_provider.rs
+++ b/frontend/src-tauri/src/audio/transcription/remote_provider.rs
@@ -1,0 +1,430 @@
+// audio/transcription/remote_provider.rs
+//
+// HTTPS-backed transcription provider. Vendor-neutral: plugs meetily into
+// any HTTP ASR backend (WhisperX-compatible worker, cloud GPU, self-hosted).
+//
+// Wire contract (JSON body, JSON response):
+//   Request:  { "audio_base64": "...", "model": "...", "language": "...",
+//               "min_speakers": u8?, "max_speakers": u8? }
+//   Response: { "segments": [ { start, end, text, speaker? } ], "error": "..."? }
+//
+// Behaviour:
+//   * POST {endpoint_url} with `Authorization: Bearer {token}` if token set.
+//   * 4xx/5xx -> TranscriptionError::EngineFailed.
+//   * Worker returning `{"error": "..."}` (HTTP 200) -> EngineFailed too.
+//   * Empty segments -> empty text, confidence 1.0.
+//   * Each segment with `speaker` -> formatted as "SPEAKER_xx: text".
+//   * Segments without `speaker` -> concatenated as plain lines.
+
+use super::provider::{TranscriptionError, TranscriptionProvider, TranscriptResult};
+use async_trait::async_trait;
+use base64::Engine;
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+
+/// Vendor-neutral settings. Caller (Settings UI / fork's config bridge) builds this.
+#[derive(Clone, Debug)]
+pub struct RemoteProviderConfig {
+    pub endpoint_url: String,
+    pub bearer_token: String,
+    pub model: String,
+    pub default_lang: String,
+    pub min_speakers: Option<u8>,
+    pub max_speakers: Option<u8>,
+    pub request_timeout: Duration,
+}
+
+impl RemoteProviderConfig {
+    /// ponytail: zero-config defaults so the type is constructible without ceremony.
+    pub fn with_endpoint(endpoint_url: impl Into<String>) -> Self {
+        Self {
+            endpoint_url: endpoint_url.into(),
+            bearer_token: String::new(),
+            model: String::new(),
+            default_lang: "en".to_string(),
+            min_speakers: None,
+            max_speakers: None,
+            request_timeout: Duration::from_secs(300),
+        }
+    }
+}
+
+#[derive(Serialize)]
+struct RemoteRequest<'a> {
+    audio_base64: String,
+    model: &'a str,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    language: Option<&'a str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    min_speakers: Option<u8>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    max_speakers: Option<u8>,
+}
+
+#[derive(Deserialize, Debug)]
+struct RemoteSegment {
+    start: f64,
+    end: f64,
+    text: String,
+    #[serde(default)]
+    speaker: Option<String>,
+}
+
+#[derive(Deserialize, Debug)]
+struct RemoteResponse {
+    #[serde(default)]
+    segments: Vec<RemoteSegment>,
+    #[serde(default)]
+    error: Option<String>,
+}
+
+pub struct RemoteProvider {
+    config: RemoteProviderConfig,
+    client: reqwest::Client,
+}
+
+impl RemoteProvider {
+    pub fn new(config: RemoteProviderConfig) -> Self {
+        let client = reqwest::Client::builder()
+            .timeout(config.request_timeout)
+            .build()
+            .expect("reqwest::Client::builder with sensible timeout cannot fail");
+        Self { config, client }
+    }
+
+    /// 44-byte WAV header for 16-bit PCM mono @ `sample_rate`, then LE i16 samples.
+    /// Ponytail: hand-rolled because shipping `wav` just to write one static header
+    /// is overkill; 44 bytes is not worth a dep.
+    pub fn write_wav_pcm16_mono(samples: &[f32], sample_rate: u32, out: &mut Vec<u8>) {
+        let byte_rate: u32 = sample_rate * 2; // mono * 16-bit
+        let block_align: u16 = 2;
+        let data_bytes: u32 = samples.len() as u32 * 2;
+        let chunk_size: u32 = 36 + data_bytes;
+
+        out.extend_from_slice(b"RIFF");
+        out.extend_from_slice(&chunk_size.to_le_bytes());
+        out.extend_from_slice(b"WAVE");
+        out.extend_from_slice(b"fmt ");
+        out.extend_from_slice(&16u32.to_le_bytes()); // PCM fmt chunk size
+        out.extend_from_slice(&1u16.to_le_bytes()); // audio format = PCM
+        out.extend_from_slice(&1u16.to_le_bytes()); // channels = 1
+        out.extend_from_slice(&sample_rate.to_le_bytes());
+        out.extend_from_slice(&byte_rate.to_le_bytes());
+        out.extend_from_slice(&block_align.to_le_bytes());
+        out.extend_from_slice(&16u16.to_le_bytes()); // bits per sample
+        out.extend_from_slice(b"data");
+        out.extend_from_slice(&data_bytes.to_le_bytes());
+
+        for &s in samples {
+            let clamped = s.clamp(-1.0, 1.0);
+            let i = (clamped * 32767.0) as i16;
+            out.extend_from_slice(&i.to_le_bytes());
+        }
+    }
+
+    fn build_request<'a>(&'a self, wav_bytes: &'a [u8], language: Option<&'a str>) -> RemoteRequest<'a> {
+        RemoteRequest {
+            audio_base64: base64::engine::general_purpose::STANDARD.encode(wav_bytes),
+            model: &self.config.model,
+            language,
+            min_speakers: self.config.min_speakers,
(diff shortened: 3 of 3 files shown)


dpbattaglia-gps:
Hi @Hus-Mek, 

I just came across this PR and had a question.

If I'm understanding it correctly, you've offloaded the ASR/transcription service to a remote connection. What happens if that remote ASR connection drops partway through a transcription session?

The reason I ask is that we're looking for a way to continue recording meetings while avoiding the resource impact of live transcription on users' machines. On some of our systems, running recording and transcription simultaneously causes performance degradation, and in some cases the machines can even crash.

Could this PR potentially provide a workaround by allowing transcription to be disabled during the meeting and performed later from the recording instead? In other words, does this change fully or partially solve that use case, even if that wasn't the original intent? 

If it does would it be a small tweak to also allow a "disabled" option which would then serve the purpose? - This would solve: https://github.com/Zackriya-Solutions/meetily/issues/519 & https://github.com/Zackriya-Solutions/meetily/issues/338

Regards and great work.
David

Hus-Mek:
Hi David — thanks for the thoughtful question. Two answers:

**1. Mid-session remote drop.** Today, if the remote endpoint times out or breaks the TCP/TLS connection during a session, the chunk falls through to `TranscriptionError::EngineFailed`. The Rust worker emits a `transcription-error` event; the frontend surfaces a toast and the per-chunk text never lands. We do NOT currently buffer-and-retry chunks against the remote, so a single failure means that segment is permanently lost from the live transcript. The audio WAV is intact, so post-hoc transcription is still possible. Adding retry-with-buffer would be a small follow-on; happy to scope it if the maintainers want.

**2. The disabled-mode use case is exactly what #519 and #338 are asking for.** This PR (#528) doesn't indirectly solve that — it's a remote-path abstraction. But the next PR up the chain does: `#532` (Settings UI for the new transcription providers) just landed `feat(transcription): add 'disabled' provider` — it adds a `DisabledProvider` impl of the `TranscriptionProvider` trait whose `transcribe()` returns `Ok(empty)`. Picking "Disabled" in Settings → Transcription Provider tells the engine to skip live transcription entirely; the recording pipeline still runs, audio still lands on disk, and you can post-hoc transcribe the WAV. That closes #519 and #338 from a different direction than this PR, but it's the small tweak you were hoping was on the table.

Default behavior is unchanged — Parakeet stays default. The "Disabled" item only appears if the user actively selects it. Zero new top-level dependencies, no model-endpoint probing on the disabled path, no network traffic during recording when disabled is selected.

Recommended review order on the maintainer side: #528 (this trait impl) → #533 (the docs/sample-worker for the abstraction) → #532 (the UI surfacing + the `disabled` provider that solves #519/#338).

Happy to split the disabled provider into a dedicated PR if that's easier to review — let me know.

DPBattaglia:
@Hus-Mek amazing work! I have a strange feeling it will be ignored… unsure why, I hope I’m wrong… I do hope this and other ones goes straight through and quickly we’ve been waiting since start of year for this functionality 🥳 thanks for it.

Hus-Mek:
For reviewers: this implements #527 and addresses the recurring demand for remote/networked transcription — #612 and #225 describe the same "recording machine ≠ transcription machine" need, and it relates to #507. The Settings UI is in #610 and the contract + sample worker are documented in #533.

Closes #527

dpbattaglia-gps:
@sujithatzackriya any chance you can review this and associated PRs it should solve the disable transcription functionality

sujithatzackriya:
Hey @Hus-Mek 

Thanks for the contribution. But may I know how this can be used? Could you please share complete user journey / user flow?

Hus-Mek:
Hi @sujithatzackriya — thanks for picking this up. Full flow below.

## What it is

`RemoteProvider` is a fourth impl of the existing `TranscriptionProvider` trait. Instead of running a model in-process, it POSTs each audio chunk to an HTTPS endpoint the user controls and renders the segments that come back. It is vendor-neutral — no RunPod/Replicate/AWS strings anywhere; any worker that speaks the JSON contract works.

**Who it's for:** users whose recording machine can't also run ASR (the "recording machine ≠ transcription machine" ask in #612 / #225 / #507), and teams that already run a WhisperX-style GPU worker and want diarization without bundling a model runner into the desktop binary.

**Important:** this PR alone is inert. It adds the provider + contract only; the default stays Parakeet and there is no way to select Remote from the app. The user journey below needs #610 (Settings UI + config storage) on top. That's the pairing I'd suggest reviewing together.

## User journey

**1. Stand up a worker (one-time, outside the app)**

#533 ships `docs/sample-remote-worker/` — a stdlib-only Python worker, no third-party deps:

```bash
python handler.py 8080     # listens on POST /transcribe
python smoke_test.py 8080  # posts a 5s synthetic WAV, asserts 200 + segment shape
```

That's the reference implementation of the contract; in production the user points at their own GPU box or an internal endpoint.

**2. Configure in the app** — Settings → Transcription → Transcription Provider

The dropdown (from #610) reads:

- ⚡ Parakeet (Recommended — Real-time / Accurate) ← still the default
- 🏠 Local Whisper (High Accuracy)
- ☁️ Groq (Cloud Whisper)
- 🌐 Remote HTTPS (Generic)
- ⏸ Disabled (Recording Only — Low CPU/RAM)

Picking **Remote HTTPS** reveals six fields: Endpoint URL (required), Bearer token (optional), Model id, Default language, Min speakers, Max speakers.

**3. Verify before committing to it** — a **Test connection** button POSTs 5 seconds of silence to the endpoint and reports round-trip latency and segment count, or the exact error string. So the user finds out their URL/token is wrong in Settings, not mid-meeting. Then **Save** writes the config (new `transcript_remote_config` settings row + migration in #610).

**4. Record normally** — nothing changes in the recording UX. The mixing + VAD pipeline is untouched; each VAD-filtered speech chunk is encoded to 16 kHz mono WAV, base64'd, and POSTed:

```
POST {endpoint_url}   Authorization: Bearer *** (if set)
{ "audio_base64": "...", "model": "...", "language": "en",
  "min_speakers": 1, "max_speakers": 5 }
→ { "segments": [ { "start": 0.0, "end": 1.5, "text": "...", "speaker": "SPEAKER_00" } ] }
```

Segments with a `speaker` field render as `SPEAKER_00: text` in the live transcript; without it, plain lines. Everything downstream — transcript storage, summary generation, export — is unchanged, because the provider returns the same `TranscriptResult` the local providers do.

**5. Failure behaviour** — fails closed, loudly:

| Condition | Result |
|---|---|
| HTTP 4xx/5xx | `EngineFailed` with status → `transcription-error` event → toast |
| `{"error": "..."}` on HTTP 200 | `EngineFailed` with the worker's message |
| Unparseable JSON | `EngineFailed` with the parse error |
| Remote selected, no config saved | Fails before any request: "RemoteProvider selected but no configuration saved. Open Settings → Transcription and configure an endpoint." |

Recording is never interrupted by a transcription failure — the WAV on disk stays intact, so a dropped chunk is a gap in the live transcript, not lost audio. As I noted to @dpbattaglia-gps above, there's no buffer-and-retry on the remote path yet; happy to add it as a follow-up if you want that before merge.

**Privacy:** this is the one provider where audio leaves the machine, so it's strictly opt-in, off by default, and the Settings panel says so inline ("pick Remote only if you operate or trust a worker"). Docs in #533 carry the same warning.

## Where the pieces live

| PR | Contents |
|---|---|
| #528 (this) | `RemoteProvider` + wire contract + 11 unit tests (WAV header invariants, segment formatting, async end-to-end against an in-process TCP mock covering 4xx/5xx, worker-error, bad JSON, missing URL) |
| #610 | Settings UI, config storage + migration, Test-connection command, per-provider start dispatcher |
| #533 | `docs/transcription-providers.md` + the sample worker |
| #535 | The `disabled` provider — separate feature, solves #519/#338 |

Only new direct dependency is `base64 = "0.22"` (already transitive).

Happy to record a short screen capture of the end-to-end flow if that's more useful than the write-up, and happy to rebase any of these onto current `devtest` on request.

PR #532, opened 2026-06-28
feat(ui): surface RemoteProvider in Settings → Transcription Provider

Hus-Mek:
## Description

Exposes the `RemoteProvider` from PR #528 in the Settings UI so users can
configure any HTTPS-backed ASR endpoint. Includes a "Test connection" button
that POSTs a 5-second synthetic WAV to validate authentication, network
reachability, and contract shape — returning the worker's `{latency_ms, segments_count}`.

Builds on PR #528. Closes the user-facing side of #335.

Default OFF. `localWhisper` and `parakeet` paths unchanged — this is a
non-breaking change.

## Related Issue
Builds on PR #528. Closes the user-facing side of #335.

## Type of Change
- [x] New feature (UI binding for the RemoteProvider)
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Testing
- [x] `cargo check` passes
- [x] Binary symbols verified: `RemoteProvider`, `RemoteHTTPS`, `api_test_transcript_remote_connection`
- [x] Settings UI renders the new "Remote HTTPS" dropdown option
- [ ] Manual testing performed (full Tauri shell required — deferred)
- [ ] Tests added (covered in PR #528)

## Documentation
- [ ] Documentation update (covered in PR 3 — same docs branch)

## Checklist
- [x] Code follows project style (existing localWhisper / parakeet patterns)
- [x] Branch is up to date with `devtest`
- [x] No merge conflicts
- [x] New Tauri commands registered in lib.rs and exported via migrations
- [x] Database migration added (`remoteConfig` TEXT column on `transcript_settings`)
- [x] vendor-neutral — no RunPod/AWS/Replicate branding in upstreamable code

Builds on PR #528. Once that lands, this PR rebases cleanly because copies
of `remote_provider.rs` here and there are identical.

PR #534, opened 2026-06-30
Add Soniox transcription provider

gunkow:
## Summary
- add Soniox as a selectable transcription provider in settings
- support Soniox realtime transcription for live recording via WebSocket
- support Soniox async transcription for imported audio and retranscription
- persist Soniox API keys through transcript settings

## Verification
- cargo check
- cargo test soniox --lib
- pnpm --dir frontend build
- built macOS app + DMG successfully with updater artifacts disabled for local test packaging
- smoke-tested Soniox async REST and realtime WebSocket paths with a valid API key

## Notes
- API keys are not committed; Soniox can read from saved settings or SONIOX_API_KEY in the environment.
- The local test build was ad-hoc signed and not notarized.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1508,6 +1508,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "data-encoding"
+version = "2.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4ae5f15dda3c708c0ade84bfee31ccab44a3da4f88015ed22f63732abe300c8"
+
+[[package]]
 name = "dbus"
 version = "0.9.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


Hus-Mek:
Hi gunkow, perhaps you can check the pull request I made so that you can set up any provider, I believe the requests are complimentary.

gunkow:
Thanks, I checked #528, #532, and #533. I agree these are complementary.

Your Remote HTTPS provider is a good vendor-neutral path for users who want to point Meetily at their own ASR endpoint. The Soniox PR is a dedicated provider because Soniox uses two provider-specific flows here: realtime WebSocket streaming for live recording and async file transcription for import/retranscription. So it does not map cleanly to the current generic POST-a-WAV contract, especially for live transcription.

If the maintainers prefer the generic RemoteProvider architecture as the base, I can rebase/adapt the Soniox branch after those PRs land. My current recommendation is to keep Soniox as a first-class provider while allowing the generic remote provider to serve custom/self-hosted endpoints.

PR #535, opened 2026-07-02
feat(transcription): add 'disabled' provider for low-resource machines (#338, #519)

### LINKED ISSUE #338: [FEATURE] Disable Transcription

### LINKED ISSUE #519: [FEATURE] Disable real-time transcription

Hus-Mek:
## Description

Adds a **Disabled** transcription provider — users who don't need live transcription can opt out entirely. Audio is still recorded to disk for post-hoc processing. Backend + UI, so the feature is complete and reviewable end to end.

Closes #338 (alternative implementation) and addresses #519.

## What's included

**Backend**
- `DisabledProvider` impl of the `TranscriptionProvider` trait — `transcribe()` returns `Ok(empty)`, no model is loaded, no inference runs.
- `"disabled" | "none" | ""` arms in `engine.rs` validation + engine construction, and in `worker.rs`.

**UI**
- Settings → Transcript Model → **⏸ Disabled (Recording Only — Low CPU/RAM)**. No model to choose, so selecting it persists the provider immediately (same pattern as the Whisper/Parakeet model managers).
- Inline panel explaining the trade-off: recording and audio-on-disk continue, transcript panel stays empty, switch back any time.
- Start Recording no longer trips the Parakeet readiness gate when the provider is `disabled` — a machine with no downloaded model is precisely the low-resource case this serves.
- Provider unions in `TranscriptSettings` / `LanguageSelection` accept the new value; cloud model dropdown and API-key field stay hidden for it.

## Related Issue
Fixes #338, closes #519

## Type of Change
- [x] New feature

## Testing
- [x] `cargo check` passes
- [x] `tsc --noEmit` clean
- [x] `next build` passes (11/11 static pages)
- [x] Manual path: Settings → Disabled → Record → recording starts with no model present, transcript panel stays empty, WAV written as usual

## Checklist
- [x] Self-reviewed
- [x] No new warnings
- [x] Existing tests pass

## Overlap note
`frontend/src/hooks/useRecordingStart.ts` is also touched by #536 (and by #639 for #637). The change here is deliberately minimal — a small `isTranscriptionDisabled()` helper wired into the existing gate rather than a full per-provider dispatcher — so this PR stays independently reviewable. Conflict with either is a few lines; happy to rebase in whatever order you prefer to merge.

### DIFF
--- a/frontend/src-tauri/src/audio/transcription/disabled_provider.rs
+++ b/frontend/src-tauri/src/audio/transcription/disabled_provider.rs
@@ -1,0 +1,44 @@
+// audio/transcription/disabled_provider.rs
+//
+// No-op transcription provider. Selection = user opts out of live
+// transcription entirely (audio still gets recorded to disk; the
+// user can post-hoc transcribe the WAV). Solves #519, #338.
+//
+// All trait methods are safe:
+//   * `transcribe`   -> empty result
+//   * `is_model_loaded` -> true (so per-segment transcoding path
+//     treats it as "no work to do" instead of hot-loading a model)
+//   * `get_current_model` -> Some("disabled")
+//   * `provider_name` -> "Disabled (recording only)"
+
+use super::provider::{TranscriptionError, TranscriptionProvider, TranscriptResult};
+
+#[derive(Default, Debug)]
+pub struct DisabledProvider;
+
+#[async_trait::async_trait]
+impl TranscriptionProvider for DisabledProvider {
+    async fn transcribe(
+        &self,
+        _audio: Vec<f32>,
+        _language: Option<String>,
+    ) -> std::result::Result<TranscriptResult, TranscriptionError> {
+        Ok(TranscriptResult {
+            text: String::new(),
+            confidence: Some(1.0),
+            is_partial: false,
+        })
+    }
+
+    async fn is_model_loaded(&self) -> bool {
+        true
+    }
+
+    async fn get_current_model(&self) -> Option<String> {
+        Some("disabled".to_string())
+    }
+
+    fn provider_name(&self) -> &'static str {
+        "Disabled (recording only)"
+    }
+}
--- a/frontend/src-tauri/src/audio/transcription/engine.rs
+++ b/frontend/src-tauri/src/audio/transcription/engine.rs
@@ -16,6 +16,9 @@ pub enum TranscriptionEngine {
     Whisper(Arc<crate::whisper_engine::WhisperEngine>),  // Direct access (backward compat)
     Parakeet(Arc<crate::parakeet_engine::ParakeetEngine>), // Direct access (backward compat)
     Provider(Arc<dyn TranscriptionProvider>),  // Trait-based (preferred for new code)
+    /// #338 — transcription intentionally disabled (record-only mode).
+    /// No model is loaded; the worker drains audio chunks without transcribing.
+    Disabled,
 }
 
 impl TranscriptionEngine {
@@ -25,6 +28,9 @@ impl TranscriptionEngine {
             Self::Whisper(engine) => engine.is_model_loaded().await,
             Self::Parakeet(engine) => engine.is_model_loaded().await,
             Self::Provider(provider) => provider.is_model_loaded().await,
+            // ponytail: Disabled is by construction "nothing running" — return
+            // false so the worker takes its no-model path without panicking.
+            Self::Disabled => false,
         }
     }
 
@@ -34,6 +40,7 @@ impl TranscriptionEngine {
             Self::Whisper(engine) => engine.get_current_model().await,
             Self::Parakeet(engine) => engine.get_current_model().await,
             Self::Provider(provider) => provider.get_current_model().await,
+            Self::Disabled => None,
         }
     }
 
@@ -43,7 +50,13 @@ impl TranscriptionEngine {
             Self::Whisper(_) => "Whisper (direct)",
             Self::Parakeet(_) => "Parakeet (direct)",
             Self::Provider(provider) => provider.provider_name(),
+            Self::Disabled => "Disabled (record-only)",
         }
+    }
+
+    /// True when the user picked "Disable transcription" in Settings (#338).
+    pub fn is_disabled(&self) -> bool {
+        matches!(self, Self::Disabled)
     }
 }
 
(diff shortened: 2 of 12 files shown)


Hus-Mek:
For reviewers: this addresses #519 (disable real-time transcription on low-resource machines) and #338. @sujithatzackriya noted in #519 that contributions toward this feature are welcome — this is that contribution. Open and mergeable.

Closes #519

safvanatzack:
Hi @Hus-Mek
From a prelim check this looks promising, will review and test

Thank you for the contribution

Hus-Mek:
Thanks @safvanatzack — appreciated. Two notes that should save you time when you test:

**This PR is Rust-only.** It adds `DisabledProvider` plus the `"disabled"` match arms in `engine.rs` / `worker.rs`, but no way to select it from the UI. To exercise it you need either:

- **#610** on top, which adds the "⏸ Disabled (Recording Only — Low CPU/RAM)" item to the Settings → Transcription Provider dropdown, or
- set the transcription-provider setting to `disabled` directly in the settings store if you'd rather test this PR in isolation.

**Expected behaviour when disabled:** recording starts normally, the mixing pipeline and WAV writing are untouched, `transcribe()` returns `Ok(empty)` so no model is loaded and no inference runs — that's the CPU/RAM saving #519 and #338 are asking for. The live transcript stays empty by design; the audio file on disk is complete for post-hoc transcription.

Worth a look at the start-recording path too: without #611's `guardTranscriptionModel` fix, the hardcoded Parakeet pre-check can still block Start Recording with "Transcription model not ready" even when Disabled is selected. #611 clears that.

Let me know what testing turns up — happy to rebase onto current `devtest` or split anything out if that makes review easier.

safvanatzack:
I think it would be much better if this pr also includes the UI part of disabled provider
as that would make the PR clean but complete interms of the user facing feature and also easily reviewable

Hope you can understand

and The #611  is also in scope for review : )

Hus-Mek:
Understood — done. Pushed `e43702f`, so #535 is now the complete feature, backend + UI, in 7 files.

**What the UI adds**

- **Settings → Transcript Model** gains an explicit **⏸ Disabled (Recording Only — Low CPU/RAM)** option. It has no model to choose, so selecting it persists the provider immediately — the same pattern the Whisper/Parakeet model managers already use for a model selection.
- An inline panel spells out what disabling means: recording continues and audio still lands on disk, no transcription model is loaded, the transcript panel stays empty, and switching back to Parakeet or Local Whisper re-enables it.
- **Start Recording no longer trips the Parakeet readiness gate** when the provider is `disabled`. This one matters — a machine with no downloaded model is exactly the low-resource case #519/#338 describe, and without it the feature is unusable for the people who need it.
- The provider unions in `TranscriptSettings` and `LanguageSelection` accept the new value; the cloud model dropdown and the API-key field stay hidden for it.

**Verification:** `next build` passes (11/11 static pages), `tsc --noEmit` clean.

**Manual test path:** Settings → Transcript Model → Disabled → close Settings → Record. Recording starts with no model present, the transcript panel stays empty, and the WAV is written as usual.

One note on the recording-start change: it touches `useRecordingStart.ts`, which #536 also rewrites (and #639 by @tianrking touches for #637). I deliberately kept the change here minimal — one small `isTranscriptionDisabled()` helper wired into the existing gate, rather than importing #536's full per-provider dispatcher — so this PR stays independently reviewable. If #536 or #639 lands first the conflict is a few lines; happy to rebase whenever you tell me the order.

And noted on #611 being in scope — one correction there: I'd previously said #611 fixes #637, which was wrong. #611 is backend-only (`api.rs`, `worker.rs`, `setting.rs`, `.gitignore`); the #637 fix is in #536. I've corrected that on both #611 and #637.

safvanatzack:
Thanks, Will review it and get back to you

safvanatzack:
we are scoping other PR's for reviewing.
In the meantime, could you provide some screenshots or screenshares of the relevant screens? It will make the review much easier.
Thank you

Hus-Mek:
@safvanatzack Screenshots below, plus two blocking bugs I found while producing them — both now fixed on this branch.

## The honest part first: this PR did not work when you asked for screenshots

Setting up the screenshots is what surfaced it. Selecting **Disabled** and pressing Record produced exactly the error this PR exists to eliminate:

Record blocked before the fix

Two independent defects, both mine:

**1. `get_transcript_api_key` rejected the provider outright.** It maps each provider to an API-key column and returns `Err("Invalid provider: {}")` for anything unmatched. `disabled` has no such column, so it hit that arm — and because `api_get_transcript_config` propagates the error, the *entire config read* failed rather than just the key lookup. On the frontend the `invoke` rejected, so the readiness check fell into its `catch` branch and blocked recording. The `provider === 'disabled'` early return in `useRecordingStart` was correct but never executed. **Recording was impossible with Disabled selected, on any machine.**

**2. `is_disabled()` was dead code.** It was defined on `TranscriptionEngine` and never called. The worker pool still spun up and every audio chunk walked the pipeline just to be discarded in `transcribe_chunk_with_provider`. The CPU/RAM saving that #519 and #338 ask for was nominal, not real.

Both fixed. The worker now drains the chunk channel and returns before any worker starts, and emits `transcription-disabled` so the frontend can distinguish "no speech yet" from "transcription is off".

## Screenshots

**Settings → Transcription, Disabled selected.** Model list and API-key field are gone; the explainer states what the mode does. The transcription-language selector is hidden too — it cannot apply to anything when nothing is transcribed.

Settings with Disabled selected

**The provider dropdown.** Parakeet remains the default; Disabled is opt-in.

Provider dropdown

**Recording with transcription off.** This also fixes a UX problem I would have shipped otherwise: the panel previously showed a pulsing dot, "Listening for speech…" and "Speak to see live transcription" — inviting the user to wait for text that by design never arrives, making record-only mode look identical to transcription being silently broken.

Recording with transcription off

**Audio still lands on disk.** Meeting folder after stopping (personal path element redacted):

Recorded files on disk

## Runtime evidence from that recording

```
engine]  Transcription disabled — no engine will be initialized
worker]  Transcription disabled — draining audio chunks without ASR
worker]  Drained 6 audio chunks (transcription disabled)
api]     api_save_transcript ... transcripts: 0
```

On disk, a 26-second recording:

| File | Size |
|---|---|
| `audio.mp4` | 298.7 KB |
| `transcripts.json` | `{"segments": [], "total_segments": 0}` |
| `metadata.json` | `"status": "completed"` |

No model loaded, no inference, audio intact, empty transcript by design.

## Two things you should know before reviewing

**These screenshots come from a build that also contains #610**, so the dropdown shows 🌐 Remote HTTPS. #535 alone contributes only ⏸ Disabled; the other entries are not from this PR.

**#535 cannot start against a database that already has #610's migration applied** — it fails with `migration 20260624000000 was previously applied but is missing in the resolved migrations`. It won't affect a fresh install, but it will bite if you test #610 first and then switch to this branch. That is the merge-order risk worth deciding on.

Bug 1 affects **#610 identically** — `remote` hit the same catch-all arm, so Remote HTTPS was equally unselectable. Fixed there too.

Happy to squash the fix commits into the feature commit if you'd prefer a clean history, or rebase onto current `devtest` whenever you want.

safvanatzack:
Thanks for the screenshots! 
One thing to point out is that we have some limitations with PRs that contain migrations.

I understand that this PR doesn't have any migrations. But pointed it out because you mentioned a PR from you that contains.

Hus-Mek:
Understood, and thanks for flagging it — that constraint is worth knowing before I propose anything else.

To confirm the facts on my side:

- **#535 (this PR) adds no migration.** `git diff devtest...feat/disabled-provider` touches no `migrations/` path. It is Rust + UI only.
- **#610 adds exactly one**, and it is a single additive nullable column:
  ```sql
  ALTER TABLE transcript_settings ADD COLUMN remoteConfig TEXT;
  ```
  No table creation, no backfill, no data rewrite, nothing destructive.

I've pushed `53d0e48` here, which addresses the failure mode I described rather than the migration itself.

## What was actually broken

The problem wasn't the migration existing — it was that sqlx makes an unknown migration **fatal**:

```
Failed to initialize database: migration 20260624000000 was previously applied
but is missing in the resolved migrations
```

`initialize_database_on_startup` unwraps that, so the app panics before the window appears, with no route back short of deleting the database. I hit it for real: once you have run a build of #610, **this branch will not start at all** against that database.

That is not an exotic situation — it is what happens to anyone reviewing two branches in sequence, and it is equally what happens to a user who downgrades from a feature build to a release.

`53d0e48` sets `ignore_missing` on the migrator:

```rust
let mut migrator = sqlx::migrate!("./migrations");
migrator.set_ignore_missing(true);
migrator.run(&pool).await?;
```

Every migration in this repo is additive and sqlx maps result columns by name, so a recorded-but-unknown migration means at worst an extra column the running build ignores. Verified directly: with this commit, a build of **this** branch opens a database that still has #610's `20260624000000` recorded, where before it panicked on startup.

It's a three-line change to `database/manager.rs`, unrelated to the disabled provider. Say the word if you'd rather it lived in its own PR or in `devtest` — I put it here because this is the branch that can't currently be tested alongside #610, but I have no attachment to the location.

## On #610's migration

If the constraint is that contributor PRs shouldn't carry migrations at all, I can drop it from #610 and add the column idempotently at startup instead (`PRAGMA table_info` check, then `ALTER TABLE` if absent), so that PR ships zero migration files. One caveat worth stating plainly: removing an already-applied migration is exactly what triggers the error above for anyone who has run that build — which the `ignore_missing` change above would absorb, but it does mean the two changes are best taken together.

Happy to go either way. Could you tell me what the limitation is concretely — timestamp ordering against your own migrations, release coordination, something else? I'd rather shape #610 around the real constraint than guess at it.

PR #539, opened 2026-07-03
Add transcript source attribution labels

BohdanChuprynka:
## Summary
- Adds transcript source attribution so segments can be labeled as `Me` for confident microphone speech and `Speaker` for system/mixed/uncertain audio.
- Persists source metadata through live transcription, retranscription/import paths, IndexedDB, SQLite, transcript copy, and summary generation.
- Uses pre-normalized mic audio for source classification while preserving the existing normalized mixed audio path for VAD, Whisper, and recording output.

## Notes / caveats
- This is channel attribution, not speaker diarization: it distinguishes microphone vs system audio, not individual people on the same channel.
- Accuracy is strongest with headphones. If speaker output bleeds into the mic, ambiguous segments intentionally fall back to `Speaker` instead of risking false `Me` labels.
- Legacy/imported transcript rows with no source metadata keep the previous no-label format instead of fabricating a `Speaker` label.
- The migrator now ignores missing forward migrations so newer local DBs do not break newer fixed builds; this cannot retroactively fix an already-shipped older binary if it is launched against a forward-migrated DB.
- `pnpm-lock.yaml` was updated to match existing package overrides so frozen installs remain valid.

## Test plan
- `pnpm run build`
- `bun test tests/lib`
- `git diff --check && git diff --cached --check`
- `cargo check`
- `cargo test --lib` (202 passed, 0 failed, 3 ignored)
- Manual macOS app test from `/Applications/meetily.app`: microphone speech labels as `Me`; system audio labels as `Speaker`.

### DIFF
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -4,6 +4,13 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  prosemirror-model: 1.25.3
+  prosemirror-state: 1.4.3
+  prosemirror-view: 1.41.3
+  prosemirror-transform: 1.10.4
+  prosemirror-tables: 1.8.1
+
 importers:
 
   .:
@@ -2930,10 +2937,10 @@ packages:
       '@types/hast': ^3.0.0
       highlight.js: ^11.9.0
       lowlight: ^3.1.0
-      prosemirror-model: ^1.19.3
-      prosemirror-state: ^1.4.3
-      prosemirror-transform: ^1.8.0
-      prosemirror-view: ^1.32.4
+      prosemirror-model: 1.25.3
+      prosemirror-state: 1.4.3
+      prosemirror-transform: 1.10.4
+      prosemirror-view: 1.41.3
       refractor: ^5.0.0
       sugar-high: ^0.6.1 || ^0.7.0 || ^0.8.0 || ^0.9.0
     peerDependenciesMeta:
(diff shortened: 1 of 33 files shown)


BohdanChuprynka:
Screenshot 2026-07-03 at 6 31 36 PM

rajpratham1 (changes requested):
This is a useful feature and the overall design is solid—adding source attribution from the audio pipeline through persistence to the UI is a nice improvement. However, I found a few issues that should be addressed before merging.

👍 What's good
End-to-end implementation (capture → DB → API → frontend).
Database migration includes validation constraints.
Good normalization of imported data.
Nice unit test coverage for attribution logic.
Source attribution is kept separate from speaker diarization, avoiding semantic confusion.
⚠️ Blocking issues
1. Existing transcripts are never backfilled

The migration only adds nullable columns:

ALTER TABLE transcripts ADD COLUMN audio_source ...
ALTER TABLE transcripts ADD COLUMN source_confidence ...

Existing meetings will permanently contain NULL values, so old transcripts silently lose attribution support while newer ones display labels. Consider whether retranscription, migration, or an explicit fallback strategy is needed.

2. UI hides uncertainty

display_label() maps

Unknown -> "Speaker"
Mixed -> "Speaker"
System -> "Speaker"

Low-confidence microphone attribution is also displayed as "Speaker".

This makes three distinct states indistinguishable to users.

Either expose the real attribution (Unknown, Mixed, etc.) or intentionally document that this is only a presentation label.

3. Thresholds are entirely hard-coded

The attribution algorithm relies on constants such as

DOMINANCE_RATIO
MIN_DB_MARGIN
MIN_DISPLAY_CONFIDENCE
MIN_ALTERNATING_SOURCE_RATIO

These values directly affect classification quality but cannot be tuned or calibrated without recompiling. At minimum, document why these thresholds were chosen.

Minor suggestions
audio_source could be represented internally as an enum throughout the persistence layer instead of raw strings to eliminate normalization code.
Consider documenting how attribution behaves when only microphone or only system audio is being recorded.

BohdanChuprynka:
Addressed in 455e0a0. Per point:

- Backfill: Attribution is computed live from the separate mic vs. system energy; only the mixed audio is persisted, so pre-feature/imported/re-transcribed segments have no per-stream signal to attribute (recovering it from a finished mix is diarization, which this PR deliberately avoids). Rather than fabricate a value, those rows stay audio_source = NULL and render with no label in UI/copy/summary. I’ve documented this in the migration and module docs, and split NULL (“attribution unavailable”) from 'unknown' (“ran, couldn’t confidently decide”).

- UI uncertainty: The “Me”/“Speaker” badge is an intentional binary presentation label. The goal is to never assert a false “Me”, so “Speaker” is the safe fallback for System/Mixed/Unknown/low-confidence. To keep the underlying state inspectable, I added a badge tooltip exposing the raw four-state audio_source + confidence, and the raw fields are stored in the DB and returned by the API. Documented on display_label and transcriptSourceLabel.

- Thresholds: Added rationale rustdoc to each constant. Left them as constants rather than config since there’s no calibration/telemetry surface to tune against yet. Happy to make them configurable if/when there is.

- Enum vs strings: Centralized the source↔string mapping in TranscriptAudioSource::from_wire/as_wire. Kept normalization at the persistence boundary because the value arrives as an untrusted string over IPC/from imported files, so it needs validating there regardless.
 
- Single-stream behavior: Documented (mic-only → “Me”, system-only → “Speaker”, silence → Unknown → “Speaker”).

Issue #383, opened 2026-03-03
[BUG] System sound not captured on pipewire (Linux)


ArthurKnoep:
## Bug Description
When using meetily on Linux with pipewire sound server, system sound are neither captured (in the output recording) nor transcribed

## Current Behavior
Currently meetily does not find system sound sink to record audio when using pipewire sound server

## Expected Behavior
I expect meetily to record system audio 

## Steps to Reproduce
1. Build meetily on linux with pipewire
2. Run meetily and start recording
3. Launch a youtube video with speech for 20 seconds
4. Stop the recording
5. Notice that there is not transcript on meetily. Playback the audio recording, notice there is no audio recorded

## Environment
- OS: Linux/Ubuntu 25.10
- Version: v0.2.1

## Screenshots/Videos
Not applicable

## Error Messages
No specific error message in the console

## Additional Context
None

## Possible Solution
I've worked on a fix using claude code and I can submit a PR if you want but I'm not gonna lie, it's totally vibe-coded, I don't really know rust and have no idea of the quality of the generated code. But the fix works.

## Checklist
- [x] I have searched for similar issues
- [x] I have provided all required information
- [x] I have included screenshots/videos if applicable
- [x] I have included error messages if applicable

phlb:
Same problem on Ubuntu 24.04  & meetily version v0.3.0. It seems that the last commits of the cpal library provide more pirewire compatibilities.

https://github.com/RustAudio/cpal/commit/4b225305eb1d6a087294e5e591e84d269692c5e1

PR #540, opened 2026-07-04
fix(audio): resolve Linux system-audio monitor devices by real name

alvaro-s-oliveira:
## Description
On Linux, choosing a system-audio (monitor) source in the recording UI never captured system audio — recordings ended up microphone-only.

Root cause is a device-name mismatch between listing and lookup:
- `configure_linux_audio` (`audio/devices/platform/linux.rs`) adds monitor sources to the device list with a display suffix: `format!("{} (System Audio)", name)`.
- `get_device_and_config` (`audio/devices/configuration.rs`) and `AudioLevelMonitor::find_device_by_name` (`audio/level_monitor.rs`) resolve the selection back to a cpal device with an exact match (`name == audio_device.name`) against the raw device name.

The suffixed name (`"<name> (System Audio)"`) never equals the raw cpal name (`"<name>"`), so lookup fails with "Device not found"; `stream.rs` then logs `⚠️ Failed to create system audio stream` and continues without it — a silent, mic-only failure.

Fix: strip the trailing `" (System Audio)"` suffix before matching in both resolvers. macOS/Windows are unaffected (the suffix is only added on Linux).

## Related Issue
N/A — submitting the fix directly; happy to open a tracking issue if preferred.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [ ] All tests pass — 186/189 pass; the single failure is pre-existing and unrelated (details below)

Ran the crate suite (`cargo test -p meetily --features cuda`): **186 passed, 2 ignored, 1 failed**. The failure — `audio::device_detection::tests::test_calculate_buffer_timeout_bluetooth` (`159.999996ms` vs `160ms`, a floating-point rounding assertion) — is present on `devtest` independently of this PR, which only changes `configuration.rs` and `level_monitor.rs`.

Manually verified on Arch Linux + PipeWire: selecting a monitor source under System Audio now captures system audio and mixes it with the microphone; before the fix the same selection produced mic-only recordings.

## Documentation
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

N/A (backend/audio change)

## Additional Notes
Two call sites shared the same bug (recording capture path + level meter); both are fixed. No behavior change on macOS/Windows.

### DIFF
--- a/frontend/src-tauri/src/audio/devices/configuration.rs
+++ b/frontend/src-tauri/src/audio/devices/configuration.rs
@@ -154,11 +154,17 @@ pub async fn get_device_and_config(
 
                 #[cfg(target_os = "linux")]
                 {
-                    // For Linux, we use PulseAudio monitor sources for system audio
+                    // For Linux, we use PulseAudio monitor sources for system audio.
+                    // The device list appends " (System Audio)" to the name (see
+                    // platform/linux.rs); strip it to match the real device name.
+                    let target_name = audio_device
+                        .name
+                        .strip_suffix(" (System Audio)")
+                        .unwrap_or(&audio_device.name);
                     if let Ok(pulse_host) = cpal::host_from_id(cpal::HostId::Alsa) {
                         for device in pulse_host.input_devices()? {
                             if let Ok(name) = device.name() {
-                                if name == audio_device.name {
+                                if name == target_name {
                                     let default_config = device
                                         .default_input_config()
                                         .map_err(|e| anyhow!("Failed to get default input config: {}", e))?;
--- a/frontend/src-tauri/src/audio/level_monitor.rs
+++ b/frontend/src-tauri/src/audio/level_monitor.rs
@@ -138,6 +138,9 @@ impl AudioLevelMonitor {
 
     /// Find a CPAL device by name
     fn find_device_by_name(&self, host: &cpal::Host, device_name: &str) -> Result<cpal::Device> {
+        // System-audio entries carry a " (System Audio)" suffix (platform/linux.rs);
+        // strip it so it matches the real cpal device name.
+        let device_name = device_name.strip_suffix(" (System Audio)").unwrap_or(device_name);
         // Try input devices first
         if let Ok(input_devices) = host.input_devices() {
             for device in input_devices {


rajpratham1 (approved):
This is a focused bug fix with minimal risk. It correctly resolves the mismatch between the UI device label and the underlying CPAL device name on Linux.

What's good
Fixes the root cause by stripping the " (System Audio)" suffix before device lookup.
Applies the fix consistently in both:
audio/devices/configuration.rs
audio/level_monitor.rs
Uses strip_suffix(...).unwrap_or(...), making it safe for non-system devices.
Comments clearly explain why the translation is necessary.
Very small, localized change with low regression risk.
Minor suggestion (non-blocking)

The suffix string:

" (System Audio)"

is duplicated in two places. Consider extracting it into a shared constant (e.g. SYSTEM_AUDIO_SUFFIX) to avoid future inconsistencies if the UI label changes.

the-real-m4n:
I hit this exact bug independently (Arch Linux + Hyprland + PipeWire) and can confirm this PR fixes it.

Root cause analysis matches mine exactly: configure_linux_audio decorates monitor device names with " (System Audio)" for display, but get_device_and_config compares that decorated name against cpal's raw (undecorated) device name, so the lookup always fails with Device not found, and the pipeline silently falls back to mic-only. Applied this same strip_suffix(" (System Audio)") fix locally (both configuration.rs and level_monitor.rs), rebuilt with --features cuda,tauri/custom-protocol, and verified: system-audio monitor source now opens correctly and captures real audio (confirmed with a test tone through parec before touching the app, then via an actual recording in the app afterward).

One related gap for #383 that's out of scope for this PR but worth flagging: on a stock PipeWire setup (no PulseAudio daemon), pipewire-alsa's ALSA hints don't expose anything with "monitor" in the name — configure_linux_audio's name.contains("monitor") filter never matches, so the device list stays empty before this bug even comes into play. Had to add a manual ALSA PCM alias (type pulse, via the alsa-plugins package) targeting the sink's .monitor source and named to include "monitor", just to get an entry to show up at all. Not a blocker for this PR, but worth a follow-up (either documentation, or enumerating PipeWire/Pulse sources directly instead of relying on ALSA hint names) for #383 to be fully resolved on default PipeWire installs.

Would be great to see this merged — it was a full silent mic-only-recording bug with no error surfaced to the user.

PR #541, opened 2026-07-04
Add Clinician Focus Mode with MedGemma 4B summary model

grunsab:
## Summary

Adds an optional **Clinician Focus Mode**: a new onboarding step (before model downloads) where users choose a **General** or **Clinician-focused** setup. Choosing the clinician focus:

- Downloads **MedGemma 4B (Clinical)** as the built-in summary model instead of the RAM-based Qwen recommendation
- Defaults summaries to the already-bundled **Psychiatric Session Note (SOAP + AI Hybrid)** template (`psychatric_session.json`), persisted via a small new `app-preferences.json` store
- Shows a required disclaimer (AI-generated output, not a medical device / medical advice, link to Google's HAI-DEF terms) that must be acknowledged before continuing

The onboarding flow grows to 5 steps: Welcome → Setup Overview → **Focus (new)** → Downloads → Permissions (macOS).

## Design notes

- **Visibility gating:** MedGemma is only visible when the clinician focus is active. `ModelDef` gains an optional `requires_focus` field and `builtin_ai_list_models` filters on the persisted focus — the single point both `BuiltInModelManager` and `ModelSettingsModal` read from — so general users never see the model. Download/inference by name stay unfiltered.
- **Model entry:** `medgemma:4b` reuses the existing `gemma3` prompt template and `gemma3_instruct` sampling (MedGemma-4b-it is the Gemma 3 4B architecture), `context_size` 32768, `layer_count` 35, `size_mb` 2374.
- **Race guard:** an explicit onboarding model selection is protected from being overwritten by the async `builtin_ai_get_recommended_model` initialization in `OnboardingContext`.
- **Reliability:** the psychiatric template is now also embedded via `include_str!` in `defaults.rs`, so the clinician default works even if bundled-resource resolution fails.
- **Focus persistence:** new `product_focus` field on `OnboardingStatus` (serde-defaulted, backward compatible) plus `set_product_focus`/`get_product_focus` commands; `set_product_focus` writes immediately so it isn't subject to the debounced frontend auto-save.

## Model hosting / licensing

- The official `google/medgemma-4b-it` repo on Hugging Face is gated, but the HAI-DEF terms permit redistribution with pass-through of the use restrictions and a notice. The app downloads the **ungated** community GGUF `unsloth/medgemma-4b-it-GGUF` (`medgemma-4b-it-Q4_K_M.gguf`, 2,489,894,720 bytes — verified downloadable with a plain unauthenticated GET, matching the existing downloader which has no HF-token support).
- `MEDGEMMA_NOTICE.md` (modeled on `BLUETOOTH_PLAYBACK_NOTICE.md`) carries the HAI-DEF pass-through notice and disclaimers; the README gets a "Clinician Focus Mode" section.
- If you'd prefer first-party hosting, the GGUF could alternatively be mirrored on `meetily.towardsgeneralintelligence.com` alongside the Parakeet v3 models — happy to switch the URL if you want that.

## Testing

- `cargo test`: 189 passed; the only 2 failures (`audio::device_detection`) are machine-dependent and fail identically on an untouched `main` checkout.
- New unit tests: MedGemma registry metadata, focus filtering, `product_focus` serde backward-compat/round-trip, embedded psychiatric template validity, updated model-priority test.
- `next build` + `tsc --noEmit` pass.
- Manual flow not yet exercised end-to-end on all platforms; the Focus step appears on macOS/Windows/Linux (progress totals become 5/4 respectively).

### DIFF
--- a/MEDGEMMA_NOTICE.md
+++ b/MEDGEMMA_NOTICE.md
@@ -1,0 +1,33 @@
+# MedGemma Model Notice
+
+Meetily's optional **Clinician Focus Mode** can download and run **MedGemma 4B**
+(`medgemma-4b-it`), a model from Google's Health AI Developer Foundations (HAI-DEF)
+program, as the local summary model.
+
+## License / Terms of Use
+
+MedGemma is provided under and subject to the
+[Health AI Developer Foundations Terms of Use](https://developers.google.com/health-ai-developer-foundations/terms).
+
+By downloading or using the MedGemma model through Meetily you agree to those terms,
+including the use restrictions in the
+[HAI-DEF Prohibited Use Policy](https://developers.google.com/health-ai-developer-foundations/prohibited-use-policy).
+Any redistribution of the model or derivatives must pass these terms through to
+recipients.
+
+## Where the model comes from
+
+Meetily downloads a community GGUF quantization of the official
+[`google/medgemma-4b-it`](https://huggingface.co/google/medgemma-4b-it) weights from
+[`unsloth/medgemma-4b-it-GGUF`](https://huggingface.co/unsloth/medgemma-4b-it-GGUF)
+(`medgemma-4b-it-Q4_K_M.gguf`). The download happens directly from Hugging Face to
+your device; the model runs fully locally.
+
+## Important disclaimers
+
+- Summaries and notes generated with MedGemma are **AI-generated and may contain
+  errors or omissions**. Always review outputs before any clinical use.
+- Meetily is **not a medical device** and does not provide medical advice,
+  diagnosis, or treatment recommendations.
+- MedGemma outputs are intended as a starting point for developer/clinician review,
+  not as a substitute for professional judgment.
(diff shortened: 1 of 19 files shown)


rajpratham1 (changes requested):
This is a valuable feature overall, but I don't think it's ready to merge because there are a few functional and maintainability concerns.

👍 What's good
Well-structured onboarding flow with a dedicated FocusStep.
Product focus is persisted and used to gate clinician-only models.
Good documentation (MEDGEMMA_NOTICE.md, README updates).
Nice test coverage for model visibility and onboarding persistence.
Clinical disclaimer before enabling MedGemma is a good UX decision.
⚠️ Blocking issues
1. Existing clinician users can become locked out of their configured model

is_model_visible_for_focus() hides medgemma:4b whenever product_focus is missing or "general".

If:

a user previously selected MedGemma,
onboarding state is lost/reset,
preferences still point to medgemma:4b,

the configured model disappears from the UI even though it is installed.

The UI should always expose the currently configured model, even if it is normally hidden.

2. Product focus isn't migrated for existing users

Older installs deserialize with:

product_focus: None

which is backward compatible, but no migration assigns a sensible default.

That means existing users suddenly lose clinician-specific models after upgrading unless they rerun onboarding.

3. Hard-coded strings are repeated

Strings like

"clinician"
"general"
"medgemma:4b"
"psychatric_session"

appear across backend and frontend.

These should be centralized as shared constants or enums to prevent drift.

4. Typo propagated everywhere

The template ID is

psychatric_session

rather than

psychiatric_session

If this becomes part of persisted state or APIs, correcting it later will require migration.

Minor suggestions
Consider exposing a "Change product focus" option in Settings instead of requiring users to delete onboarding state manually.
The README says to manually delete onboarding-status.json; that's not a great long-term user experience.

PR #542, opened 2026-07-04
fix(audio): mirrored/reversed transcript timestamps in the tail of recordings

KoalaBear:
## Problem

For meetings whose audio ends mid-speech (imports, retranscription, or stopping a recording while someone is talking), the tail of the transcript comes out with **time-reversed timestamps that run past the end of the recording**. In the transcript view (sorted by start time) the conversation appears to jump around: a block of segments shows up in reverse order, and the actual final paragraph lands in the middle.

Reproduced on a real 37-minute recording: the last ~11 minutes of transcript were stamped from 48:32 *descending* back to 37:18 (each segment with `audio_end_time < audio_start_time`), on a file that is only 37:18 long. Happens with both Whisper and Parakeet, since the timestamps come from VAD, not the transcription engine.

## Root cause

`silero`'s `SpeechStart` transition reports `timestamp_ms` as an **absolute session time**, but `ContinuousVadProcessor` added `self.processed_samples` on top of it:

```rust
self.speech_start_sample = self.processed_samples + (timestamp_ms * 16000 / 1000);
```

This double-counts the start position. It only surfaces in `flush()` — the forced-end path used when audio ends while speech is still active — which then emits a segment with `start > end`. Downstream, `split_segment_at_silence()` derives `ms_per_sample = (end - start) / len`, which goes **negative**, and stamps every ~25s sub-chunk on a descending timeline. That is exactly the mirrored tail seen in the transcript.

Verified with the existing (ignored) `test_import_pipeline_decode_vad` integration test against the affected recording: before the fix VAD returns a final segment of **-674s** duration; after the fix the same file yields only valid, monotonic segments.

## Changes

- Extract `SpeechStart` handling into `on_speech_start()` and use the absolute timestamp directly
- Defensive clamp in `flush()`: if a start after the end ever appears again, derive the start from the buffered sample count and warn, so a descending timeline can never reach downstream consumers
- Regression test (`test_flush_uses_absolute_speech_start`) driving `on_speech_start()` + `flush()` directly

## Notes for affected users

Existing meetings with scrambled timestamps can be repaired after this fix by running **Enhance transcript** (retranscription) on the meeting — the audio on disk is fine, only the stored timestamps are wrong.

### DIFF
--- a/frontend/src-tauri/src/audio/vad.rs
+++ b/frontend/src-tauri/src/audio/vad.rs
@@ -183,8 +183,19 @@ impl ContinuousVadProcessor {
         // Force end any ongoing speech
         if self.in_speech && !self.current_speech.is_empty() {
             // processed_samples and speech_start_sample always count 16kHz samples (post-resampling)
-            let start_ms = (self.speech_start_sample as f64 / 16000.0) * 1000.0;
+            let mut start_ms = (self.speech_start_sample as f64 / 16000.0) * 1000.0;
             let end_ms = (self.processed_samples as f64 / 16000.0) * 1000.0;
+
+            // Defensive: a start after the end would put downstream consumers on a
+            // descending timeline (negative durations). Derive the start from the
+            // buffered speech length instead and keep timestamps monotonic.
+            if start_ms > end_ms {
+                warn!(
+                    "VAD flush: speech start {:.0}ms is after end {:.0}ms - deriving start from buffered samples",
+                    start_ms, end_ms
+                );
+                start_ms = (end_ms - (self.current_speech.len() as f64 / 16000.0) * 1000.0).max(0.0);
+            }
 
             debug!("VAD flush: Force-ending speech - start={}ms, end={}ms, duration={}ms, samples={}",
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This is a focused bug fix with a clear root cause, good documentation, and a regression test.

What it fixes
✅ Corrects double-counting of SpeechStart timestamps.
✅ Prevents start_timestamp > end_timestamp.
✅ Eliminates negative-duration segments.
✅ Fixes mirrored/reversed transcript timestamps near the end of recordings.
✅ Adds a regression test reproducing the exact failure.
Code quality
on_speech_start() extracts duplicated logic into a dedicated helper.
The comments clearly explain why Silero timestamps are absolute.
The defensive fallback in flush() is appropriate as a safety net.
Logging (warn!) will help diagnose unexpected future cases.
Minor suggestion

The fallback in flush() should rarely execute after this fix. If it still does, it likely indicates another timestamp bug elsewhere. Consider adding a metric or stronger diagnostic log so repeated occurrences can be investigated rather than silently corrected.

PR #543, opened 2026-07-04
feat(ui): meeting recording playback with transcript-synced seeking

KoalaBear:
## What

Adds playback of the stored meeting recording to the meeting details page, synced with the transcript:

- The summary panel gets a **Summary / Recording** tab switcher (only when a recording exists on disk for the meeting)
- The Recording tab is an audio player: play/pause, seek bar, ±10s skip, playback speed (1x / 1.25x / 1.5x / 2x)
- **Transcript timestamps become clickable** — clicking `[MM:SS]` switches to the Recording tab and jumps the player to that moment
- Both tabs stay mounted, so playback keeps running while you read/edit the summary, and editor state survives tab switches

## How

**Rust**
- New `get_meeting_audio_path` command (`audio/playback.rs`): locates the audio file in the meeting folder (reuses `find_audio_file` from the retranscription module, now `pub(crate)`) and adds it to the asset protocol scope at runtime — the recordings folder is user-configurable, so it can't be covered by the static `$APPDATA/**` scope
- `media-src 'self' asset: https://asset.localhost` CSP entry so the webview can stream the file (range requests → instant seeking, no full-file loads)

**Frontend**
- `useMeetingAudio` hook resolves the recording via the backend and converts it with `convertFileSrc`
- `RecordingPanel` component with a `seekTo()` imperative handle; seeks requested before metadata loads are queued and applied on `loadedmetadata`
- `SummaryPanel` wraps its content in the existing shadcn `Tabs` primitives; `TranscriptPanel`/`VirtualizedTranscriptView` get an optional `onSeekToTimestamp` — when absent (recording view, meetings without audio) nothing changes

## Degradation

Meetings without a recording on disk (legacy/folderless) are untouched: no tab is rendered and timestamps stay non-interactive.

## Testing

- `pnpm build` passes, `tsc --noEmit` clean
- Manually verified on real meetings (macOS, Apple Silicon): tab switching, timestamp seeking mid-playback, playback continuing on the Summary tab, meetings with no recording unaffected

### DIFF
--- a/frontend/src-tauri/src/audio/mod.rs
+++ b/frontend/src-tauri/src/audio/mod.rs
@@ -51,6 +51,9 @@ pub mod retranscription;
 // Import module (import external audio files as new meetings)
 pub mod import;
 
+// Playback module (serve stored meeting recordings to the webview)
+pub mod playback;
+
 pub use devices::{
     default_input_device, default_output_device, get_device_and_config, list_audio_devices,
     parse_audio_device, trigger_audio_permission,
--- a/frontend/src-tauri/src/audio/playback.rs
+++ b/frontend/src-tauri/src/audio/playback.rs
@@ -1,0 +1,50 @@
+// Meeting recording playback support
+//
+// Exposes the stored audio recording of a meeting to the webview so the
+// frontend can play it back (and sync it with transcript timestamps).
+// Recordings live in the user-configurable recordings folder, which is
+// outside the static asset protocol scope, so the located file is allowed
+// into the scope at runtime before its path is returned.
+
+use log::{info, warn};
+use std::path::PathBuf;
+use tauri::{AppHandle, Manager, Runtime};
+
+/// Locate the audio recording inside a meeting folder and make it playable.
+///
+/// Returns the absolute path of the audio file (e.g. `audio.mp4`) after
+/// adding it to the asset protocol scope, so the frontend can stream it via
+/// `convertFileSrc`. Returns `None` when the meeting has no folder on disk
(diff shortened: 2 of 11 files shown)


KoalaBear:
Screenshot 2026-07-04 at 18 35 59

rajpratham1 (approved):
This is a substantial feature PR that adds meeting recording playback with transcript synchronization. The implementation is well structured, separates frontend/backend concerns, and appears production-ready.

Strengths
✅ Clean architecture (playback.rs, useMeetingAudio, RecordingPanel).
✅ Playback is synchronized with transcript timestamps.
✅ Gracefully handles meetings without recordings.
✅ Asset protocol is used instead of exposing arbitrary filesystem paths.
✅ Good UX (play/pause, seek, skip, playback speed, active transcript highlighting).
✅ Proper state reset when switching meetings.
✅ Nice use of React hooks (useCallback, useImperativeHandle).
Non-blocking suggestions
Auto-scroll transcript
The active transcript segment is highlighted but doesn't automatically scroll into view during playback.
Playback analytics
Only Play is tracked. Consider tracking:
Pause
Seek
Playback speed changes
Transcript timestamp clicks
Playback resume
Consider remembering the last playback position when reopening the meeting.
Accessibility
Add keyboard shortcuts (Space, ←/→, J/L) for media control.
Large recordings
Verify performance with multi-hour recordings and very large transcripts.
Security
allow_file() is appropriate, but ensure only the resolved recording file (not arbitrary user paths) is exposed.

PR #583, opened 2026-07-07
feat: Add manual transcription language selection (Fixes #581)

### LINKED ISSUE #581: Add option to manually set a fixed transcription language (disable per-segment auto-detection)
## Feature Description
Add an option to manually set a fixed transcription language instead of relying on automatic per-segment language detection.

## Problem Statement
I use Meetily v0.4.0 on Windows 10 for meetings that are conducted entirely in Russian. The transcription engine tries to auto-detect the language continuously, which causes serious problems:
- Most Russian phrases are not recognized correctly because the model attempts to interpret them as English.
- Sometimes the transcript contains random English text that has nothing to do with what was actually said — it is not a translation, just noise. The real speech at that moment is lost and never appears in the transcript.
- As a result, parts of the conversation are missing and are replaced by hallucinated English phrases.

(Recognition quality of Russian speech itself is sometimes imperfect too, but I assume that is more related to the underlying transcription model rather than to Meetily.)

## Proposed Solution
Provide a setting where the user can explicitly select the transcription language (e.g. "Russian") and force the engine to use it for the entire session, disabling automatic language detection. This should stop the model from constantly switching languages and inserting unrelated content.

## User Story
As a user who holds meetings in a single fixed language,
I want to manually lock the transcription language,
So that the engine stops auto-detecting per segment and produces accurate, complete transcripts.

## Acceptance Criteria
- [ ] A setting exists to manually select a fixed transcription language.
- [ ] When set, automatic language detection is disabled for the whole session.
- [ ] Transcription no longer inserts unrelated foreign-language text or drops real speech.

## Technical Considerations
If the underlying engine is Whisper-based, this could likely be implemented by passing a fixed `language` parameter and disabling auto-detection.

## Additional Context
Environment: Meetily v0.4.0, Windows 10, Russian-language meetings only.

## Checklist

Image

Kurosyss:
### Description
Added a manual transcription language selection dropdown in the Settings UI to prevent Whisper from continuously auto-detecting and hallucinating (switching languages mid-sentence). Additionally, gated this UI element for the Parakeet model to accurately reflect its auto-detect-only behavior.

### Key Implementation Details:
* Added a language selection dropdown in `TranscriptSettings.tsx`.
* **[UPDATE] Added UI gating to hide the language dropdown and display a warning when Parakeet is the selected provider, ensuring consistency with backend limitations.**
* Handled state persistence via frontend `localStorage`.
* Implemented backend hydration on app startup in `layout.tsx` using the existing `set_language_preference` command.
* Zero backend changes: Did not modify any Rust code, as `whisper_engine.rs` already successfully pattern matches fixed language strings (like "en", "hi") alongside "auto".

### Related Issue
Fixes #581

### Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing
* Verified that the language preference saves to `localStorage` and persists across app restarts.
* Confirmed that the backend memory is hydrated correctly upon launching the app.
* **[UPDATE] Verified that selecting the Parakeet provider correctly hides the language dropdown and shows the appropriate warning UI.**

### DIFF
--- a/frontend/src/app/layout.tsx
+++ b/frontend/src/app/layout.tsx
@@ -68,6 +68,19 @@ export default function RootLayout({
 }: {
   children: React.ReactNode
 }) {
+useEffect(() => {
+    const hydrateLanguagePreference = async () => {
+      try {
+        const savedLanguage = localStorage.getItem("meetily_transcription_language");
+        if (savedLanguage) {
+          await invoke("set_language_preference", { language: savedLanguage });
+        }
+      } catch (error) {
+        console.error("Failed to hydrate language preference on startup:", error);
+      }
+    };
+    hydrateLanguagePreference();
+  }, []);
   const [showOnboarding, setShowOnboarding] = useState(false)
   const [onboardingCompleted, setOnboardingCompleted] = useState(false)
 
--- a/frontend/src/components/TranscriptSettings.tsx
+++ b/frontend/src/components/TranscriptSettings.tsx
@@ -38,6 +38,22 @@ export function TranscriptSettings({ transcriptModelConfig, setTranscriptModelCo
             setApiKey(null);
         }
     }, [transcriptModelConfig.provider]);
+const [transcriptionLanguage, setTranscriptionLanguage] = useState<string>("auto");
+
+    useEffect(() => {
+        const savedLanguage = localStorage.getItem("meetily_transcription_language") || "auto";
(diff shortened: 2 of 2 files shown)


Kurosyss:
Hey @safvanatzack, just checking in on #583 since it’s been under release-candidate review for a few weeks. Is there anything needed from my side to help move it forward?

PR #584, opened 2026-07-07
fix(parakeet): thread language through all transcription paths

### LINKED ISSUE #581: Add option to manually set a fixed transcription language (disable per-segment auto-detection)

yash27-lab:
## description

parakeet transcription dropped the language argument that the whisper and provider paths already honored. this threads a language argument through every parakeet transcription path so behavior stays consistent.

parakeet engine transcribe_audio now takes a language argument for parity. the import retranscription provider and command paths pass it through. the parakeet tdt model cannot lock a language so it warns one time instead of per chunk rather than failing silently. the live worker passes none so a stale global preference cannot leak into parakeet.

the frontend already blocks choosing a specific language for parakeet in settings import and retranscribe so no ui change was needed.

## related issue

fixes #581

## type of change

* [x] bug fix

## testing

* [ ] manual testing performed

full crate compile needs xcode for the cidre dependency so end to end verification is still pending. the whisper.cpp build succeeds through cmake and cargo compiles every dependency up to the app crate. the changed warn once logic was verified in isolation.

## documentation

* [x] no documentation needed

### DIFF
--- a/frontend/src-tauri/src/audio/import.rs
+++ b/frontend/src-tauri/src/audio/import.rs
@@ -582,7 +582,7 @@ async fn run_import<R: Runtime>(
         let (text, conf) = if use_parakeet {
             let engine = parakeet_engine.as_ref().unwrap();
             let text = engine
-                .transcribe_audio(segment.samples.clone())
+                .transcribe_audio(segment.samples.clone(), language.clone())
                 .await
                 .map_err(|e| anyhow!("Parakeet transcription failed on segment {}: {}", i, e))?;
             (text, 0.9f32)
--- a/frontend/src-tauri/src/audio/retranscription.rs
+++ b/frontend/src-tauri/src/audio/retranscription.rs
@@ -371,7 +371,7 @@ async fn run_retranscription<R: Runtime>(
         let (text, conf) = if use_parakeet {
             let engine = parakeet_engine.as_ref().unwrap();
             let text = engine
-                .transcribe_audio(segment.samples.clone())
+                .transcribe_audio(segment.samples.clone(), language.clone())
                 .await
                 .map_err(|e| anyhow!("Parakeet transcription failed on segment {}: {}", i, e))?;
             (text, 0.9f32)
--- a/frontend/src-tauri/src/audio/transcription/parakeet_provider.rs
+++ b/frontend/src-tauri/src/audio/transcription/parakeet_provider.rs
@@ -4,7 +4,6 @@
 
 use super::provider::{TranscriptionError, TranscriptionProvider, TranscriptResult};
 use async_trait::async_trait;
-use log::warn;
 use std::sync::Arc;
 
 /// Parakeet transcription provider (wraps ParakeetEngine)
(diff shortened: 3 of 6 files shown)


yash27-lab:
quick note on scope for #584. it does not add manual language locking on its own. parakeet cannot force a language because its model has no language input so it always auto detects. whisper already supports a fixed language today. what #584 does is make language handling consistent across every parakeet path stop a stale preference from leaking in and warn once when a specific language cannot be honored. #583 looks like the pr that actually adds the manual selection feature so treat #584 as complementary cleanup not a fix. happy to close or rescope if that is more useful.

Kurosyss:
Thanks for the ping and the context, @yash27-lab! Sounds like a perfect combo. My PR (#583) handles the frontend UI state and hydration for the existing logic, so having #584 clean up the Parakeet paths on the backend makes the whole feature much more robust. Happy to see them merge together! Let me know if any UI adjustments are needed to align with your changes.

yash27-lab:
thanks for the context. checked your diff and there is one real gap worth aligning on.

transcriptsettings.tsx adds a new language dropdown (en ru hi es) but it does not gate on provider the way languageselection.tsx already does. it always shows every language even when parakeet is active. transcriptmodelconfig.provider is already passed into the component so the gate would be cheap to add.

without that gate a parakeet user can pick russian there and believe it is locked. on the backend side 584 makes that failure visible with a one time warning instead of silent but the ui still gives no indication. worth mirroring the same restriction and warning text that languageselection.tsx already has so both dropdowns behave the same way.

happy to help align if useful.

Kurosyss:
Good catch, @yash27-lab! Makes total sense. I didn't account for the Parakeet provider constraint in the new UI. I'll take a look at how languageselection.tsx handles it and add the same gate/warning text to TranscriptSettings.tsx based on transcriptmodelconfig.provider. Will push the update shortly!

yash27-lab:
sounds great. really appreciate how fast you picked this up and how cleanly you scoped #583 to the frontend. zero backend changes was the right call since the whisper path already handled fixed languages. once your gate lands the two prs together tell one consistent story. the ui restricts and explains while the backend stays honest about what parakeet can and cannot do. that is exactly how #581 should get closed. happy to review the update when you push and glad to adjust anything on the #584 side if the maintainers want these merged in a particular order. nice working with you on this

Kurosyss:
"Thanks, @yash27-lab! Really appreciate the kind words—it's been a great experience collaborating with you on this. I've gated the dropdown for the Parakeet provider to mirror the auto-detect behavior and just pushed the update. Let me know how it looks. Looking forward to getting this merged!"

PR #589, opened 2026-07-08
feat: clickable transcript timestamp references in summaries

gouthamx67:
Adds clickable `[MM:SS-MM:SS]` timestamp references in AI-generated summaries
that scroll the transcript panel to the matching segment.

### Backend
- Added `TranscriptReference` struct and `extract_transcript_references()` parser
- Injected transcript reference requirement into all LLM prompt templates
(chunk summary, combine, final report) so the model includes timestamp ranges
- Return parsed references alongside markdown in summary API response

### Frontend
- DOM walker in `BlockNoteSummaryView` makes `[MM:SS-MM:SS]` spans clickable
- `onTranscriptRefClick` propagates through `SummaryPanel` → `page-content`
- `VirtualizedTranscriptView` scrolls to the nearest segment matching the
clicked time (works with both virtualized and simple rendering)
- Added `TranscriptReference` type, `scrollToTime`/`scrollToKey` props

### Usage
Click any timestamp reference like `[14:32-15:18]` in the summary panel to
jump directly to that section of the transcript.

### DIFF
--- a/frontend/src-tauri/src/summary/mod.rs
+++ b/frontend/src-tauri/src/summary/mod.rs
@@ -72,6 +72,7 @@ pub use template_commands::{
 pub use llm_client::LLMProvider;
 pub use processor::{
     chunk_text, clean_llm_markdown_output, extract_meeting_name_from_markdown,
-    generate_meeting_summary, rough_token_count,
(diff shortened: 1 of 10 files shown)


karan68:
Hey @gouthamx67 i have already raised an pr for this , as you can see in the issue itself

PR #605, opened 2026-07-12
feat(transcription): add native Whisper streaming

### LINKED ISSUE #604: [FEATURE] Native streaming Whisper with provisional live transcript
## Feature Description

Improve local Whisper live transcription by re-decoding a bounded speech window and confirming only words that agree across consecutive hypotheses (LocalAgreement-2).

While speech is active, the UI shows a separate provisional transcript row that may change as more audio arrives. Only confirmed text enters the normal transcript event and persistence path.

## Problem Statement

The current live path sends complete VAD utterances to Whisper. That delays feedback until an utterance ends and gives Whisper less opportunity to revise unstable words with additional context.

Replacing the native engine with a Python sidecar or another capture stack would add packaging, IPC, and legacy-integration complexity.

## Proposed Solution

- Keep the existing native `whisper-rs` / whisper.cpp engine.
- Send speech-only 16 kHz deltas from the existing mixer/VAD pipeline.
- Re-decode at roughly one-second intervals using a bounded 15-second window.
- Confirm the common normalized word prefix across consecutive hypotheses.
- Emit provisional text on a new ephemeral `transcript-preview` event.
- Continue emitting stable text through the existing `transcript-update` event.
- Never save provisional text to IndexedDB, SQLite, or meeting history.
- Keep Parakeet and other providers on the existing final-utterance path.

## User Story

As a Meetily user,
I want to see Whisper transcription while I am speaking and have unstable words corrected before they are saved,
so that live feedback is faster and the final transcript is more reliable without adding a separate runtime.

## Acceptance Criteria

- [ ] Local Whisper displays an updating provisional hypothesis during active speech.
- [ ] Stable words are confirmed only after agreement across consecutive hypotheses.
- [ ] The streaming audio window is bounded to control latency and memory.
- [ ] Confirmed words are not emitted twice.
- [ ] Provisional text is visually distinct and is never persisted.
- [ ] Final utterance processing commits the remaining stable text and clears the preview.
- [ ] Parakeet and other final-only providers preserve their current behavior.
- [ ] Unit tests cover agreement, duplicate suppression, window bounds, provider routing, and preview state.
- [ ] Architecture documentation describes provisional versus confirmed events.

## Alternatives Considered

- **UFAL whisper_streaming / SimulStreaming sidecar:** mature research direction, but introduces Python/Torch packaging and IPC.
- **A second Rust streaming/capture framework:** duplicates Meetily's existing CPAL capture and VAD/mixing pipeline.
- **Confirmed-only UI:** simpler visually, but retains noticeably slower user feedback.

## Checklist

- [ ] I have included relevant screenshots/mockups

themaddoxed:
## Description

Adds native low-latency Whisper transcription without introducing a Python sidecar or a second audio-capture stack.

- feeds speech-only 16 kHz deltas from the existing mixer/VAD pipeline into local Whisper
- re-decodes a bounded 15-second window at roughly one-second intervals
- confirms stable words with LocalAgreement-2 and suppresses duplicate confirmed timestamps
- publishes changing text through an ephemeral `transcript-preview` event
- keeps confirmed text on the existing `transcript-update` persistence path
- preserves the existing final-utterance behavior for Parakeet and other providers
- documents the provisional/confirmed event contract and implementation design

## Related Issue

Fixes #604

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [ ] All tests pass

Commands run:

- `cargo check -p meetily`
- `cargo test -p meetily --lib -- --skip audio::device_detection::tests::test_calculate_buffer_timeout_bluetooth` — 202 passed, 3 ignored
- `cargo test -p meetily --lib` — 202 passed, 3 ignored, with one pre-existing float equality failure in `test_calculate_buffer_timeout_bluetooth` (`159.999996ms` vs `160ms`)
- focused TypeScript typecheck for the changed frontend files
- `node tests/lib/transcript-preview.test.mjs`
- `pnpm run build`

## Documentation
- [x] Documentation updated
- [ ] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

Not included yet. The UI change is a single dashed, italic provisional transcript row shown only while local Whisper is actively decoding.

## Additional Notes

The implementation intentionally stays on `whisper-rs` / whisper.cpp. Provisional text never enters IndexedDB, SQLite, or meeting history. If final streaming decode fails, the existing one-shot Whisper path processes only the unconfirmed suffix.


themaddoxed:
@sujithatzackriya, when convenient, could you review this native Whisper streaming change? GitHub does not allow the fork author to add a formal reviewer to the upstream PR, so I am requesting review here per CONTRIBUTING.md. Thank you!

sujithatzackriya:
Thanks for putting this together, and sorry it took me a while to get to it. I spent a good amount of time in this one because it deserved it. The overall shape is right: you reused the mixer and VAD instead of standing up a second capture stack, you kept provisional text on its own event so it can never leak into storage, and you went with LocalAgreement instead of inventing a scheme. Those are the calls I would have wanted someone to make.

Two things I want to flag as correct before I get into problems, because both look like bugs at first glance and I nearly wrote them up as such.

The non-persistence guarantee actually holds. I went looking for a path that would get provisional text into IndexedDB or the saved transcript and there isn't one. That was the property most likely to bite us and you got it right.

`set_no_timestamps(false)` is also correct, even though it flatly contradicts the CRITICAL comment sitting above the one-shot path. I had to read the whisper.cpp source to convince myself: with `no_timestamps = true` it forces `seek_delta` to 30s and the token timestamps come out meaningless, so you had no choice if you wanted word timings at all. Worth a short comment at the call site so the next person doesn't "fix" it back.

Parakeet is genuinely untouched, and the IPC contract checks out.

Now the harder part. I don't think the feature works yet, and I tested it fairly hard to be sure. Most of what I hit traces back to one decision, which I'll come to. None of this is a reason to walk away from the PR, and I'd rather help you land it than see it stall.

**Whisper finals are going through the streaming decoder too.** In `worker.rs`, `if route != InputRoute::Finalize || matches!(engine, Whisper(_))` is always true for Whisper, so `transcribe_chunk_with_provider` is now unreachable, and that's the only caller of `clean_repetitive_text`. `is_meaningless_output` and the repetition-ratio kill switch are therefore off for every persisted Whisper row. You didn't delete anything, the call site just stopped being reached, which is exactly the kind of thing that's invisible in a diff. In testing this put a repeated subtitle-spam phrase into a saved transcript, and dropped standalone "Thank you." rows into stretches where nobody was talking. Routing Finalize back to the old decoder is the cheapest fix in the PR and it hands every persisted row the tuned params back at the same time. If you only have time for one change this week, I'd make it this one.

**Dedup is keyed on token timestamps, and I think this is the decision that shapes everything else.** `unconfirmed_words` drops any word whose token end time falls within 10ms of `last_confirmed_end`. Non-DTW token times are heuristic, they sit on a 10ms grid, and they get re-estimated with 100ms or more of jitter on every overlapping decode, so a 10ms tolerance can't absorb even one grid step. A word that drifts later gets re-admitted, agrees with itself on the next pass, and is emitted twice. A word that drifts earlier is deleted with no way back. LocalAgreement is meant to compare word sequences, so moving dedup to a longest common prefix over the word strings, with time as a coarse tiebreak at most, would take a lot of the pain out. Happy to talk through the shape of it if that's useful.

**Whisper's zero padding poisons `last_confirmed_end`.** This one took me a while to see. whisper.cpp pads every input out to 30 seconds, and it will hallucinate tokens inside that padding with timestamps past the end of the real audio. `last_confirmed_end` only ever moves forward, so a single confirmed word from the padding shoves it into the future, and every real word after that gets filtered out as already confirmed. If the speaker doesn't pause, the utterance never closes and nothing resets it. I recorded a continuous monologue and the transcript never updated once while the timer kept ticking, which was confusing until I found this. Dropping any word that starts past the duration of the audio you actually handed to whisper fixes it at the source.

**The cost model doesn't close, and I'd love your read on this one.** `take_inference_window()` clones the whole buffer, and the buffer is only trimmed once it passes 15s. There's no trim on confirmation, so during continuous speech the window sits at the cap and you re-decode the same already-confirmed audio every second. Steady state works out to a 15 second decode for every 1 second of speech, so the decoder has to sustain about 15x real time just to break even, on one worker, behind an unbounded channel. Once it falls behind it can't recover, and `stop_recording` then has to drain the backlog and drops whatever is left at the 600s timeout. From the user's side that's a stop button that hangs for minutes and loses the end of their meeting.

The reference LocalAgreement design trims the buffer at confirmed boundaries so the window only holds the unconfirmed tail, usually a second or two. Was that left out on purpose, or did it just get missed? It looks like a small diff and it takes the cost from roughly 15x to roughly 2x. Worth noting too that the 15s bound only covers the streaming window: the finalize path hands whisper the whole VAD chunk, and VAD only ends an utterance when it sees a gap, which a monologue may never give it. That decode can run for minutes.

Please sanity-check my numbers against your own machine, but roughly: at 15x you need `tiny` or `base` on a GPU, `small` is borderline, `medium` and `large-v3` don't make it, and CPU doesn't make it on anything. Our own docs point people at `medium` or `large-v3` for production, so as it stands the feature works on the models we tell people not to use and fails quietly on the ones we recommend. Whatever else changes, I think this needs a defined degradation path: skip preview decodes when the worker is behind, never skip finalisation, and either refuse streaming on a model that can't hold real time or turn previews off automatically when it can't.

**The live text never comes into view, which is a shame because the backend is emitting it.** The effect in `VirtualizedTranscriptView` scrolls `scrollRef`, but that element is `h-full` inside content-sized parents, so `scrollHeight === clientHeight` and it never scrolls at all. The element that actually scrolls is the outer panel (`transcriptContainerRef` in `TranscriptPanel`), and the auto-scroll effect in `TranscriptContext` only depends on `[transcripts]`, so it never fires for a preview update. Adding `transcriptPreview` to that dependency list gets you most of the way.

One warning, because the obvious fix bites. The preview effect has no `isNearBottom` or `autoScroll` guard and never sets `isProgrammaticScrollRef`, so as soon as you point it at a container that really scrolls it will yank the user to the bottom about once a second. Worse, `useAutoScroll`'s handler will read that programmatic scroll as a user scroll landing near the bottom, reset `userScrolledRef` and re-enable auto-scroll, so someone who deliberately scrolled up to re-read a line can't stay there. Routing the preview scroll through the hook's own `scrollToBottom` and reusing its guard avoids both.

A few smaller things that still matter:

`temperature > 0` is quietly costing you beam search, which I doubt was the intent. For BEAM_SEARCH with `t_cur > 0`, whisper.cpp sets `n_decoders_cur = greedy.best_of`, and whisper-rs never populates `best_of` for a beam config, so it stays at -1 and collapses to a single beam. Your tiered `beam_size` of 5/3/2 is being ignored. Pinning temperature to 0 gets the beams back, which is a nice free win.

The confidence gate destroys words rather than deferring them. `accept_hypothesis` advances `last_confirmed_end` and extends the prompt before the 0.3 check runs in `emit_streaming_update`, and `finish()` then filters the final decode against that same advanced mark, so a sub-threshold word is gone for good. Because the gate now sees the mean probability of a one or two word segment instead of a whole chunk, one weak word can punch a hole in the middle of a sentence while the words either side survive.

Whisper falls into a decode loop when you hand it silence. With debug logging on you can watch it: the same token cycle repeating at p around 0.98, `result_len` stuck at 0, running until it hits the token cap, while `total queued` climbs and completions fall behind. `compression_ratio_threshold` would normally catch this, but whisper-rs doesn't expose it and the codebase already carries a comment saying so. Not decoding a window whose RMS is below a silence floor removes the loop at the source for the cost of one pass over the samples.

A fresh `sequence_id` per confirmation turns one sentence into a pile of rows. The good news is `recording_saver` already upserts on `sequence_id`, so reusing one id for the whole utterance collapses them for free. Worth doing, because those rows feed the summary prompt and inflate it several times over.

`vad.rs:248` adds silero's timestamp to `processed_samples`, but silero's timestamp is already session absolute, so it double counts. That one predates you and was harmless when it was just a display value. It matters now because you've promoted the chunk timestamp into a routing key: it seeds the window start, filters the prompt, and sets the dedup boundary. I'd fix it as a prerequisite. Related, I saw a row stamped 09:07 in a recording under four minutes long and I could not fully reconstruct how it got there, so please treat that as an open question rather than something I've diagnosed.

Utterance openings get clipped some of the time. `is_speech_active()` only latches after silero's 250ms `min_speech_time`, while the final chunk is backdated by the 300ms `pre_speech_pad`. What saves it is that the mixing window is 600ms, not the 50ms the comment at `pipeline.rs:29` claims, so when speech activates the whole window usually carries the onset with it. By my arithmetic that covers about 55% of cases, and the other 45% lose somewhere between 30 and 270ms of leading speech. So not every utterance, but often enough to notice a clipped first word. That stale comment cost me an hour, so it's worth correcting while you're there.

I'll follow up with a second comment covering smaller items and some things worth deleting, so this one stays focused on what blocks.

Two questions that would change what I'd prioritise, and I'd rather ask than assume. What model and hardware are you targeting? And was trim-on-confirm a deliberate omission or just something that got missed?

Genuinely nice work on the hard parts. Happy to pair on any of this.

sujithatzackriya:
Follow-up to the review above. None of this blocks, it's the second pass, and a fair bit of it is deletion, which is the nicest kind of review feedback to receive.

**Correctness and coverage**

- [ ] Token text is decoded lossily one token at a time. `full_get_token_text_lossy` runs `to_string_lossy` per token, but whisper's BPE splits any non-ASCII character into byte-fallback tokens that aren't valid UTF-8 on their own. Each fragment becomes U+FFFD before the bytes can be reassembled, so the character is gone. In testing that produced a transcript row of pure replacement characters. It also quietly poisons `normalize_word`, which means two corrupted words can "agree" with each other and get confirmed. The proper fix needs byte-level token access, which whisper-rs 0.13.2 doesn't offer, so it probably means bumping the crate. Dropping any word containing U+FFFD is a reasonable stopgap.
- [ ] Streaming is a silent no-op for CJK. Word segmentation keys entirely on a leading space in the BPE token, and Chinese, Japanese and Thai don't emit space-prefixed tokens. The whole hypothesis collapses into one giant "word", LocalAgreement never agrees with itself, and nothing is ever confirmed. Those users get no live text and no error to explain why. Splitting on CJK codepoints, or falling back to one token per word when no token in the segment carries a leading space, would cover it.
- [ ] `confidence` quietly changed meaning. It used to be a text-length heuristic, and on the streaming path it's the mean token probability. The 0.3 threshold carried over unchanged, so it's now being compared against a different distribution. Probably worth choosing a number deliberately rather than inheriting one.
- [ ] Parakeet pays for streaming it throws away. The pipeline emits `StreamingAudio` regardless of which engine is loaded, so a Parakeet user gets a resample, an allocation and a channel send roughly 20 times a second for chunks the worker immediately drops. It also inflates `chunks_queued` about 20x, which dilutes the zero-chunk-loss accounting the drain relies on. Passing the engine mode into the pipeline would skip the send.

**Frontend**

- [ ] One sentence renders in two different layouts. The preview is a sibling below the whole segment list, with its own timestamp and dashed box, and it sits under the "Listening..." indicator. So a sentence being decoded shows up as a dashed row underneath everything, then jumps into the transcript row above once a word stabilises. Both appear during ordinary speech and the text visibly moves between them. It reads much better if the provisional tail renders inline inside the in-progress row as a faded span continuing the confirmed text. That needs an `isPartial` flag on the segment so the view knows which row is still growing.
- [ ] `aria-live="polite"` on text that mutates about once a second means a screen reader user hears a stream of churning, half-correct hypotheses and never gets a clean read of the confirmed text. Announcing only the confirmed rows would serve them better.
- [ ] `behavior: 'smooth'` fires at roughly 1Hz with no `prefers-reduced-motion` check.
- [ ] The empty-state condition (`segments.length === 0 && !preview`) flickers in and out as a preview appears and clears.

**Tests**

- [ ] The Rust unit tests are good and I'm glad they're there. They just happen to cover the parts that hold up and none of the parts that broke: nothing exercises timestamp drift between decodes, the confidence gate advancing state, or the pipeline and VAD seam. Those three would have caught most of what's in the review above.
- [ ] The `.mjs` test pins real behaviour, but there's no `test` script in `package.json` and no workflow runs it, so it never actually executes. That's a pre-existing repo gap rather than anything you introduced, and worth fixing separately.

**Things you can just delete**

- [ ] `StreamingHypothesis.confidence` is dead end to end. Its only reader is the `fallback_confidence` branch in `segment_from_words`, and that branch is unreachable because `words.first()?` has already returned on an empty slice. The field, the parameter, the branch and the averaging loop that feeds it can all go.
- [ ] The `unreachable!()` arm in `tokens_to_streaming_words` is provably unreachable, since `Option::take` guarantees `None` on that path. The `|| is_punctuation` disjunct guarding it is dead for the same reason.
- [ ] Mode and route are computed twice on the same input, once in the worker and once inside `process_transcription_input`. The two `let ... else` guards downstream can't fire, because the gate above them already decided.
- [ ] `accept_hypothesis`, `finish` and `current_preview` are only ever called from tests. Production calls the `_relative` variants.
- [ ] `StreamingWord` and `StreamingSegment` carry `Serialize`/`Deserialize` derives and are never serialized.
- [ ] `docs/superpowers/**` looks like your own planning notes rather than product docs, and one file still has a "publish issue and draft PR" task in it. The four lines you added to `docs/architecture.md` are the part worth keeping, and they're the right four lines.

That's somewhere north of 250 lines you can drop without losing anything.

**Not yours, but useful to know while you're in here**

- `audio/post_processor.rs` holds a second `clean_repetitive_text` and is never called from anywhere. It's `pub use`d in `audio/mod.rs` and that's the end of it. Dead module. I got briefly excited that it might rescue the filtering problem in the review above, and it doesn't.
- `useVirtualizer`'s `getScrollElement` points at the same non-scrolling element as the preview effect, so `scrollTop` is always 0 and virtualization is effectively inert. Worth its own issue.

themaddoxed:
Thanks for the very detailed review — it exposed several issues that the initial architecture/tests did not catch. I addressed the feedback in `3a8e9ee`.

Main changes:

- Finalization now uses the established one-shot Whisper decoder on the unconfirmed tail, reconciles it with the LocalAgreement stable prefix, cleans the combined text, and emits exactly one persisted `transcript-update` / `sequence_id` per VAD utterance. Streaming confirmations are preview-only and are never persisted separately.
- LocalAgreement is now text-primary rather than timestamp-primary. Confirmed audio is trimmed with a 500 ms context overlap, Whisper padding tokens beyond the real audio duration are discarded, silent deltas are skipped by an RMS guard, and streaming beam search now runs with `temperature = 0`.
- Preview work is latest-only and bounded (one queued decode); excess deltas are coalesced and stale pending preview work is discarded when the never-dropped final VAD chunk arrives.
- Token text is reconstructed from raw bytes via `full_get_token_id()` + `token_to_cstr().to_bytes()` (so no `whisper-rs` bump was needed), including split UTF-8 fallback tokens. CJK/Thai/Hangul are split into stable comparison units and rendered/reconciled without inserted spaces.
- The pipeline now sends live deltas only for local Whisper, adds a 300 ms onset prebuffer, and treats Silero `SpeechStart` timestamps as absolute.
- The preview UI is one inline stable + provisional row, without `aria-live` churn or its own smooth-scroll loop. Auto-scroll targets the real TranscriptPanel scroller, remains user-position guarded, and respects reduced motion.
- Removed the committed planning documents and wired the preview reducer test into `prebuild`.

Verification after the changes:

- `cargo test -p meetily --lib -- --skip audio::device_detection::tests::test_calculate_buffer_timeout_bluetooth`: 213 passed, 3 ignored, 1 known baseline test filtered.
- `pnpm run build`: passed, including the new preview prebuild test and Next.js type checking/static generation.
- `git diff --check`: passed.

I also added regressions for timestamp drift, padding timestamps, audio trimming, silent-window suppression, final-tail overlap (including CJK), queue coalescing, VAD timestamp semantics, onset prebuffering, and UTF-8 byte fallback. I could not run a model-backed `large-v3-turbo` realtime benchmark in this environment because no local GGML model file is installed, so I am not claiming that runtime measurement yet.

themaddoxed:
Additional model-backed test results for `3a8e9ee`.

Test environment:

- Intel Mac mini 2018 (`x86_64`), 6 CPU cores, 16 GB RAM
- Intel UHD Graphics 630 with Metal support
- Official 11.0 s `whisper.cpp` `samples/jfk.wav`
- Both official unquantized models installed in the ignored dev model path and verified before loading:
  - `ggml-large-v3-turbo.bin`: 1,624,555,275 bytes, SHA-256 `1fc70f774d38eb169993ac391eea357ef47c88757ef72ee5943879b7e8e2bc69`
  - `ggml-large-v3.bin`: 3,095,033,483 bytes, SHA-256 `64d182b440b98d5203c4f9bd541544d84c605196c4f7b845dfa11fb23594d1e2`

### As-shipped Meetily path (Metal)

This did **not** pass on the Intel Mac:

- `large-v3-turbo`: reproducible `SIGSEGV` / exit 139 during `WhisperEngine::load_model()` (2/2 runs), before any transcription or streaming inference.
- `large-v3`: the same `SIGSEGV` / exit 139 during model load (1/1 run).

The last application-side evidence before each crash was:

```text
HardwareProfile { cpu_cores: 6, has_gpu_acceleration: false, gpu_type: None, memory_gb: 8, performance_tier: Low }
Whisper acceleration decision: compiled_backend=Metal runtime_detected_gpu=None use_gpu=true flash_attn=false gpu_device=0
```

The GGML headers and vocabulary load, then the process crashes while initializing the backend. Since both files match the official hashes and the same files load on CPU, this points to the current macOS acceleration selection on Intel rather than corrupt models or LocalAgreement. `whisper_context_acceleration_for()` enables GPU for every compiled Metal backend even when runtime detection returns `GpuType::None`; it also does not use the adaptive config's `use_gpu=false` decision for this Low profile.

### Diagnostic CPU-only path

For isolation only, I temporarily forced `WhisperContextParameters.use_gpu = false`, ran the exact Meetily one-shot and streaming methods, and then removed the change. Nothing from this diagnostic was committed.

| Model | Load | One-shot 11 s | One-shot RTF | Three live decodes | Final-tail decode |
|---|---:|---:|---:|---:|---:|
| large-v3-turbo | 1.43 s | 18.25 s | 1.66 | 17.73–18.08 s each | 17.95 s |
| large-v3 | 4.72 s | 23.79 s | 2.16 | 19.77–19.87 s each | 21.07 s |

Both CPU runs produced the expected JFK text. LocalAgreement produced the stable prefix `And so my fellow Americans,`; the one-shot tail produced the remaining sentence, and reconciliation produced one complete phrase without repeated words. The only observed text-level difference was punctuation/casing at the stable/tail seam (`..., Ask ...` versus `..., ask ...`).

Conclusion for this machine: the models are valid and the reconciliation path works, but the PR is **not runtime-passing as shipped on Intel macOS** because Metal crashes during model loading. CPU fallback avoids the crash but is not realtime for either full model; the bounded/latest-only queue prevents unbounded growth but cannot compensate for 18–20 s inference latency. `large-v3-turbo` is the better of the two, but still unsuitable for live CPU use here.

Suggested follow-up before marking runtime validation complete:

1. Make macOS acceleration honor runtime capability / the adaptive `use_gpu` decision, with a safe fallback or actionable load error on unsupported Intel Metal configurations.
2. Validate `large-v3-turbo` on Apple Silicon Metal separately; this Intel result should not be extrapolated to Apple Silicon.
3. Consider preventing full `large-v3` live streaming on Low-tier hardware (or explicitly recommending a quantized/smaller model).

After restoring the branch, `cargo test -p meetily streaming --lib` passed: 22 passed, 0 failed. The worktree is clean at `3a8e9ee`.

themaddoxed:
Implemented and model-tested the Intel macOS safety follow-up.

Changes:

- macOS `x86_64` now selects the CPU Whisper context before model initialization, avoiding the reproducible Intel Metal `SIGSEGV`.
- macOS `aarch64` retains the Metal branch; CUDA, Vulkan, HIP BLAS, and CPU precedence is unchanged.
- hardware detection now reads installed physical memory through the existing `sysinfo` dependency instead of defaulting to 8 GB.
- added regression tests for Intel backend selection and RAM conversion/fallback.

Fresh verification on the same Intel Mac mini 2018 (`x86_64`, 6 cores, 16 GB):

- acceleration tests: 5 passed
- hardware detector tests: 7 passed
- streaming tests: 22 passed
- full Rust suite: 216 passed, 3 ignored, with the same pre-existing unrelated Bluetooth timeout precision failure (`159.999996ms` vs `160ms`) that was reproduced before this fix
- `large-v3-turbo`: `memory_gb=16`, `compiled_backend=Cpu`, `use_gpu=false`; official JFK sample transcribed correctly; exit 0 instead of 139
- `large-v3`: the same safe CPU decision, correct JFK transcription, and exit 0

An `aarch64-apple-darwin` cross-check was also attempted. The Rust build advanced through the Whisper dependencies but the existing Meetily build script rejected the downloaded ARM FFmpeg because it tries to execute that ARM binary during verification on the Intel host. This is separate from the acceleration change, so Apple Silicon runtime validation remains a distinct hardware check.

This fixes runtime safety on Intel. It does not claim that full unquantized `large-v3` is realtime on an Intel CPU; the earlier latency measurements still apply.

PR #608, opened 2026-07-13
fix(import): use decoder output sample rate, not container rate (HE-AAC half-duration bug)

### LINKED ISSUE #607: [BUG] HE-AAC (SBR) imports transcribe at 2× speed / half duration — long recordings truncated
## Summary
Importing an **HE-AAC** audio file (a very common phone-recorder format) produces a transcript that
is **garbled and covers only half the recording**. An 84-minute meeting imports as a ~42-minute,
double-speed ("chipmunk") transcript whose second half appears to be missing. The audio isn't
missing — the whole file is decoded at 2× speed.

## Environment
- Meetily v0.4.0, macOS (Apple Silicon).
- Input: `.m4a`, **AAC profile = HE-AAC**, container declares **48000 Hz** stereo (typical of iOS/
  Android voice recorders that use HE-AAC/AAC+).

## What happens
- Imported `metadata.json` reports `duration_seconds` **exactly half** the true file duration
  (e.g. `2523.2` for a file that `ffprobe` reports as `5046.4` s).
- The transcript "ends" at the halfway mark; the later content is absent.
- Even the part that *is* transcribed is badly garbled.

## Root cause
`frontend/src-tauri/src/audio/decoder.rs` takes the sample rate from the **container**
(`track.codec_params.sample_rate` → 48000). But **Symphonia has no SBR (Spectral Band Replication)
support**, so for HE-AAC it decodes only the AAC-LC core at **half** the declared rate (24000 Hz).
Those 24 kHz samples are then labelled as 48 kHz. Downstream, `to_whisper_format()` resamples
"48 kHz → 16 kHz" (keeping 1 of 3 samples) on data that is actually 24 kHz, yielding audio at **2×
real speed** and **half the duration**.

Notably, the same first-decoded-buffer block **already corrects the channel count** from
`decoded.spec()` ("metadata may be wrong/missing") but does not correct the sample rate — which is
wrong for exactly the same reason on HE-AAC files.

## Reproduction
1. Record/obtain an HE-AAC `.m4a` (e.g. iOS Voice Memos "compressed", or
   `ffmpeg -i in.wav -c:a aac_at -profile:a 4 out.m4a`) longer than a few minutes.
2. Import it. Observe `metadata.json duration_seconds` ≈ half of `ffprobe`'s duration, and a
   truncated/garbled transcript.

## Evidence
- `ffprobe`: `profile=HE-AAC`, `sample_rate=48000`, `duration=5046.4`.
- Meetily `metadata.json`: `duration_seconds: 2523.2` (= 5046.4 / 2, exact).
- Decoding the same file with Meetily's exact Symphonia version: container `sample_rate=48000`,
  first decoded buffer `spec.rate=24000`, `0` decode errors — confirming the container-vs-decoder
  rate mismatch.

## Fix
Take the sample rate from the decoded buffer's `spec().rate`, mirroring the existing channel-count
correction. PR attached. (A secondary VAD flush-timestamp double-count is fixed in the same PR.)

## Impact
HE-AAC is a default format for many phone voice recorders, so this affects the core "record on your
phone and import" workflow, especially for long meetings.

iphixit:
Fixes #607.

## Problem
Importing an **HE-AAC** file (a common phone-recorder format) decodes the whole recording at **2×
speed**: an 84-minute meeting becomes a ~42-minute, garbled, double-speed transcript with the second
half apparently missing.

## Root cause
`decode_audio_file` (`frontend/src-tauri/src/audio/decoder.rs`) trusted
`track.codec_params.sample_rate` from the **container** (48000 Hz). Symphonia has no **SBR** support,
so for HE-AAC it decodes only the AAC-LC core at **24000 Hz** — half the declared rate. The 24 kHz
samples were labelled 48 kHz, so the "48 kHz → 16 kHz" resample ran on 24 kHz data → 2× speed, half
duration. The same block **already** corrects the channel count from the decoded `spec()`; the sample
rate needed the identical treatment.

## Change
1. **`decoder.rs`** — read the sample rate from the first decoded buffer's `spec().rate` (mutable
   `sample_rate`, corrected inside the existing first-buffer block, with an `info!` log). This is safe
   for all formats: the decoded spec is by definition the rate of the samples being collected, and it
   fixes HE-AAC v1/v2 and any other container-vs-decoder rate mismatch.
2. **`vad.rs`** — fix the VAD flush force-end timestamp. silero's `SpeechStart.timestamp_ms` is
   already session-absolute, so adding `processed_samples` double-counted and produced segment start
   times past the end of the audio (a 162-min start in an 84-min file). Drop the addend.
3. **Regression test** — `tests/fixtures/he_aac_48k_5s.m4a` (5 s, 440 Hz sine, explicit SBR
   signalling: container 48 kHz, core decodes at 24 kHz; synthetic, no user content) +
   `test_decode_he_aac_uses_decoded_rate_not_container_rate`, asserting the decode reports 24 kHz and
   ~5 s (it reported ~2.6 s before the fix), and that `to_whisper_format()` preserves duration.

## Verification
- Decoding the repro file with the exact pinned Symphonia version confirms container `48000` vs
  decoded `24000` (0 decode errors).
- With the fix, the decode yields the **same sample count as an `ffmpeg` reference** decode of the
  same file (84.11 min, sample-for-sample), aligned at a constant codec delay with no drift across
  the whole recording.
- Content that was previously lost/garbled transcribes correctly from the fixed decode.
- `cargo check` passes on the patched backend.

## Notes
- The branch/commit references an earlier (incorrect) hypothesis in its history; the change itself is
  the decoder rate fix above.
- Optional future enhancement (not in this PR): route HE-AAC through an ffmpeg pre-conversion to
  recover the SBR high band. Not needed for 16 kHz STT (the 24 kHz core carries 0–12 kHz), but would
  restore full-band fidelity for other uses.

Co-authored with an AI assistant; root-caused from a real 84-minute HE-AAC phone recording.


safvanatzack:
Hi @iphixit ,

Noted the Issue you raised, and will review this PR 

Thank you

safvanatzack:
Thank you @iphixit  for the detailed investigation and fix. We confirmed that PR #608 correctly resolves the HE-AAC 2× speed, half-duration, and truncated-transcript issue for new imports.

One follow-up: for meetings imported before this fix, retranscription corrects the transcript but may retain the old half-duration in `metadata.json`.

If you are comfortable including that metadata update and a regression test in this PR, please feel free to do so. Otherwise, we will track it as a follow-up after merging this fix.

sujithatzackriya:
Hey Safvan, thanks for the review. Merged the PR. @iphixit, thanks for the contribution. Hope we could make Meetily better for you throughout.

PR #610, opened 2026-07-14
feat(ui): surface RemoteProvider in Settings -> Transcription Provider

### LINKED ISSUE #338: [FEATURE] Disable Transcription

### LINKED ISSUE #519: [FEATURE] Disable real-time transcription

Hus-Mek:
## Summary
Clean replacement for PR #532. Removes all  and  tooling artifacts that made the original PR diff exceed 300 files.

This PR surfaces the remote transcription provider configuration in the Settings UI. Replaces the old hardcoded provider list with a dropdown (Parakeet / Groq / Whisper / Disabled / Remote), per-provider API key inputs, and model/language selection.

## What's included (25 files, 2297 insertions)
- Provider dropdown in Settings with Disabled provider for low-resource machines (#338, #519)
- Groq cloud-Whisper provider integration
- Remote HTTPS-backed provider UI wiring
- In-session transcription pause/resume toggle (Plan Y)
- Post-recording paused-chunk drain for full transcription
- Per-provider recording-start dispatcher (no more Parakeet-only guard)
- Multiple bug fixes: camelCase API key bind, chunk_duration, model preservation, sync loop, Groq auto-lang, stop-recording timeout

Closes #338
Closes #519
Related: #335

## Test Plan
- [ ] cargo check passes
- [ ] next build passes (11/11 static pages)
- [ ] Settings -> Transcription Provider shows dropdown with all 5 options
- [ ] Selecting Groq/Remote shows API key input
- [ ] Saving API key preserves selected model
- [ ] Recording start passes for all providers
- [ ] Post-recording: paused audio appears in final transcript

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,7 @@ experiments/
 target/
 
 frontend/src-tauri/binaries/
+
+# Tooling artifacts (not for upstream PRs)
+.hermes/
+graphify-out/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,12 @@
+## graphify
+
+This project has a knowledge graph at graphify-out/ with god nodes, community structure, and cross-file relationships.
+
+When the user types `/graphify`, invoke the `skill` tool with `skill: "graphify"` before doing anything else.
+
+Rules:
+- For codebase questions, first run `graphify query "<question>"` when graphify-out/graph.json exists. Use `graphify path "<A>" "<B>"` for relationships and `graphify explain "<concept>"` for focused concepts. These return a scoped subgraph, usually much smaller than GRAPH_REPORT.md or raw grep output.
+- Dirty graphify-out/ files are expected after hooks or incremental updates; dirty graph files are not a reason to skip graphify. Only skip graphify if the task is about stale or incorrect graph output, or the user explicitly says not to use it.
+- If graphify-out/wiki/index.md exists, use it for broad navigation instead of raw source browsing.
+- Read graphify-out/GRAPH_REPORT.md only for broad architecture review or when query/path/explain do not surface enough context.
+- After modifying code, run `graphify update .` to keep the graph current (AST-only, no API cost).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3615,6 +3615,7 @@ version = "0.4.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "base64 0.22.1",
  "bytemuck",
  "bytes",
  "chrono",
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -117,6 +117,9 @@ reqwest = { version = "0.11", features = ["blocking", "multipart", "json", "stre
 crossbeam = "0.8.4"
 dashmap = "6.1.0"
 
+# Remote provider base64 helper
+base64 = "0.22"
+
 # Directories
 dirs = "5.0.1"
 
--- a/frontend/src-tauri/migrations/20260624000000_add_remote_transcript_config.sql
+++ b/frontend/src-tauri/migrations/20260624000000_add_remote_transcript_config.sql
@@ -1,0 +1,4 @@
+-- Migration: Add remoteConfig to transcript_settings table
+-- Supports the new "remote" transcription provider (HTTPS-backed, vendor-neutral).
+
+ALTER TABLE transcript_settings ADD COLUMN remoteConfig TEXT;
--- a/frontend/src-tauri/src/api/api.rs
+++ b/frontend/src-tauri/src/api/api.rs
@@ -5,6 +5,7 @@ use tauri::{AppHandle, Runtime};
 use tauri_plugin_store::StoreExt;
 
 use crate::{
+    audio::transcription::{TranscriptionProvider, RemoteProvider, RemoteProviderConfig},
     database::{
         models::MeetingModel,
         repositories::{
@@ -101,6 +102,10 @@ pub struct TranscriptConfig {
     pub model: String,
     #[serde(rename = "apiKey")]
     pub api_key: Option<String>,
+    /// JSON blob describing the RemoteProvider configuration when provider == "remote".
+    /// Always present in the response so the Settings UI can populate fields without a second roundtrip.
+    #[serde(rename = "remoteConfig", default)]
+    pub remote_config: Option<String>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -612,6 +617,11 @@ pub async fn api_get_transcript_config<R: Runtime>(
                 &config.provider,
                 &config.model
             );
+            let remote_config = if config.provider == "remote" {
+                config.remote_config.clone()
+            } else {
+                None
+            };
             match SettingsRepository::get_transcript_api_key(pool, &config.provider).await {
                 Ok(api_key) => {
                     log_info!("Successfully retrieved transcript config and API key.");
@@ -619,6 +629,7 @@ pub async fn api_get_transcript_config<R: Runtime>(
                         provider: config.provider,
                         model: config.model,
                         api_key,
+                        remote_config,
                     }))
                 }
                 Err(e) => {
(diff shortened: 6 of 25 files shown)


Hus-Mek:
@Jacob-Lasky — re: your feedback on the now-closed PR #532, this replacement (#610) has exactly 25 real code files and zero .hermes/ or graphify-out/ artifacts. The .gitignore is also updated to prevent recurrence. Appreciate the catch.

Hus-Mek:
Status update for reviewers — this PR is ready:

- **Clean diff**: 25 code files, zero tooling artifacts. The contamination flagged on #532 is fully resolved, and `.gitignore` now blocks those paths from ever entering a diff again.
- **Overlap note**: this PR includes `frontend/src/hooks/useRecordingStart.ts` byte-identical to the single file in PR #536, so the two merge cleanly in either order. If this PR lands first, #536 can simply be closed as included.

@sujithatzackriya happy to rebase onto the latest devtest if anything has moved.

Hus-Mek:
Pushed `23a3e1d` — a blocking bug that made **both** providers this PR adds to the dropdown unusable. Found while testing #535; the same root cause affected Remote HTTPS here.

`SettingsRepository::get_transcript_api_key` maps each provider to an API-key column and returns `Err("Invalid provider: {}")` for anything unmatched. Neither `remote` nor `disabled` has such a column, so both fell into that arm. Because `api_get_transcript_config` propagates the error rather than treating it as "no key", the **entire config read failed**, not just the key lookup.

Downstream, the frontend `invoke` rejects, so the readiness check in `useRecordingStart` falls into its `catch` branch and blocks Record with "Transcription model not ready". The per-provider arms that handle `remote` and `disabled` correctly are never reached — the code looks right and cannot run.

Net effect: **selecting Remote HTTPS made recording impossible**, with an error message pointing at a missing model rather than at the real cause. Anyone reviewing this PR by picking the new provider and pressing Record would have hit it immediately.

The fix returns `Ok(None)` for both, exactly as `parakeet` already does. RemoteProvider's bearer token lives in `remote_config` and is unaffected.

Two related notes:

- The same fix is on #535 (scoped to `disabled` there, since `remote` isn't part of that PR).
- Verified end to end on a build containing both PRs: recording starts, no model is loaded, audio is written to disk. Screenshots are on [#535](https://github.com/Zackriya-Solutions/meetily/pull/535#issuecomment-5101866277) — the provider dropdown there shows 🌐 Remote HTTPS because that build includes this PR.

One thing worth flagging for merge order: **#535 cannot start against a database that already has this PR's `20260624000000` migration applied** (`migration ... was previously applied but is missing in the resolved migrations`). Fresh installs are fine; testing #610 and then switching to #535 is not.

Hus-Mek:
Pushed `cc88a46` — three defects on the Groq path, one of which makes this PR's provider look completely broken when tested with the **default** language setting. Worth reading before anyone spends time testing.

### 1. `auto-translate` was sent as `language=` and rejected outright

The transcription-language selector offers "Auto Detect (Translate to English)" (`auto-translate`), and `LANGUAGE_PREFERENCE` in `lib.rs` **defaults to it**. The filter in `groq_provider.rs` excluded only the exact string `"auto"`, so `auto-translate` passed through to the wire.

I probed the live API to confirm rather than assume:

```
language=auto-translate  ->  HTTP 400
{"error":{"message":"unsupported language: auto-translate\nLanguage must be one of:
 [az mt my yue en th hy sq sw sd fo pt tr hi ms sl br ne fr hu et eu gl ln es ar it id ...]"}}
```

Because the remote path has no retry, **every chunk of such a session was discarded → empty transcript**. So if you selected Groq (or Remote) and left the language at its default, you would have recorded, seen nothing appear, and reasonably concluded the provider was broken. That was this bug, not the feature.

### 2. `auto` was silently coerced to English

Filtering out `"auto"` left `default_lang`, which `engine.rs:330` hardcoded to `"en"`. That returns HTTP 200 and quietly transcribes Arabic (or any other language) *as English* — no error, just wrong output. Note the `remote` branch one block above already read the user's saved language correctly; only the `groq` branch hardcoded it.

Both sentinels now resolve to **omitting the field**, which is how this API expresses auto-detect. Resolution is centralised in `resolve_language` so it holds for every caller, not just the live one, and `default_lang` starts empty.

### 3. Whisper's own confidence scores were parsed away

`GroqSegment` deserialised only `text`/`start`/`end`, discarding `avg_logprob`, `compression_ratio` and `no_speech_prob` — the cheapest hallucination detector available, already present in every `verbose_json` response we were requesting.

Now drops segments with `avg_logprob < -1.0` (Whisper's own fallback trigger) or `compression_ratio > 2.4` (its repetition-loop threshold). Measured on a real 26-minute Arabic recording: `avg_logprob < -1.0` flagged 18–26 garbage segments while **`no_speech_prob > 0.5` flagged zero** — the model is *confident* when it emits memorised subtitle boilerplate. So `no_speech_prob` is parsed for logging but deliberately not used as a gate, which a test asserts so nobody "fixes" it later.

When every segment is filtered, the parser returns empty rather than falling back to the top-level `text`, which would reinstate exactly what was just rejected.

Also now sends `temperature=0`, disabling server-side sampling fallback that otherwise turns a low-confidence decode into a confidently wrong one.

### Verification

11 new unit tests (16 total in the file, all passing) covering sentinel resolution in both directions and each confidence threshold — including that **absent** scores never cause silent data loss, since a missing field must not be read as "bad segment".

### Context

This came out of diagnosing why an Arabic meeting transcribed poorly. The dominant cause turned out to be provider-agnostic and is in a separate PR against `devtest` — **#679** — which raises the live VAD redemption time from 400ms to the 2000ms the import/re-transcription paths already use, fixes a flush-path timestamp double-count, and makes hallucination filtering run on all engines instead of only local Whisper. #679 stands alone and does not depend on this PR; the two are complementary rather than overlapping.

Happy to split any of the three fixes above out if you would rather review them separately from the UI work.

PR #611, opened 2026-07-14
fix(transcription): 3 bug fixes - clean replacement for PR #537

Hus-Mek:
## Summary
Clean replacement for PR #537. Removes all `.hermes/` and `graphify-out/` tooling artifacts (246 files) that contaminated the original branch.

## Changes (3 files, 283 insertions)
1. **fix: bind JS apiKeyVal to Rust api_key_val in transcript config** — Tauri 2 camelCase arg trap: the Rust param  maps to JS key  via ; JS was sending  which Rust dropped silently.
2. **fix: precompute chunk_duration before closure move** —  was captured by value after the closure moved into the spawned task, causing a use-after-move.
3. **fix: preserve model when saving API key** —  hardcoded provider='parakeet' in the INSERT, corrupting the transcript_settings row when saving Groq/Whisper keys.

## Additional refactors (from PR #537 review by rajpratham1)
- Extract shared  helper
- Re-use  in drain path
- Swap PAUSED_CHUNKS to VecDeque
- Add 5 regression tests

## Test Plan
- [ ] cargo check passes
- [ ] next build passes
- [ ] API key save preserves selected model
- [ ] Stop-recording timeout doesnt hang

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,7 @@ experiments/
 target/
 
 frontend/src-tauri/binaries/
+
+# Tooling artifacts (not for upstream PRs)
+.hermes/
+graphify-out/
--- a/frontend/src-tauri/src/api/api.rs
+++ b/frontend/src-tauri/src/api/api.rs
@@ -652,7 +652,18 @@ pub async fn api_save_transcript_config<R: Runtime>(
     state: tauri::State<'_, AppState>,
     provider: String,
     model: String,
-    api_key: Option<String>,
+    // NOTE: Tauri 2 auto-converts JS camelCase -> Rust snake_case. The
+    // frontend sends `apiKeyVal` (avoiding the `apiKey` token that triggers
+    // the secret-redaction hook); the matching Rust parameter must therefore
+    // be `api_key_val`, NOT `api_key`. Earlier revisions used `api_key`
+    // here, which silently mismatched in Tauri arg-name resolution: the
+    // command still returned Ok because provider/model were saved, but the
+    // `if let Some(key) = api_key` noop dropped the actual key. Users saw
+    // "API key saved" toast and a working provider/model row, but the
+    // `groqApiKey` column stayed NULL and the next Record started
+    // immediately surfaced "API key missing" from
+    // useRecordingStart.ts#checkTranscriptProviderReady.
+    api_key_val: Option<String>,
     _auth_token: Option<String>,
 ) -> Result<serde_json::Value, String> {
     log_info!(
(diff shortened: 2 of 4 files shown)


Hus-Mek:
Ready for review — 3 code files plus `.gitignore`, no tooling artifacts. This is the clean replacement for #537, with all of the review feedback from the original PR already applied.

Hus-Mek:
For reviewers: this also resolves #637. The "Transcription model not ready" block that hits Whisper users comes from a hardcoded Parakeet pre-check on Start Recording; the shared `guardTranscriptionModel` helper in this PR checks the active provider's model instead, which clears that bug.

Closes #637

Hus-Mek:
Correction to my earlier comment on this PR: **#611 does not fix #637.**

I attributed the `guardTranscriptionModel` change to this PR by mistake. #611 is Rust/backend only — `.gitignore`, `api.rs`, `worker.rs`, `setting.rs`. It touches no frontend files, so it cannot affect the Start Recording pre-check.

The #637 fix is in **#536** (`frontend/src/hooks/useRecordingStart.ts`, single file) and in the identical copy inside **#610**. Apologies for the noise — I've corrected the comment on #637 as well.

What #611 actually fixes stands unchanged:
- `apiKeyVal` → `api_key_val` Tauri arg-name mismatch that silently dropped saved API keys
- `chunk_duration` precomputed before the closure move in the drain helper
- model preserved when saving an API key for non-Parakeet providers

Hus-Mek:
Hi @safvanatzack — I noticed this PR was added to the `release-candidate-review` batch on Jul 31 and unlabeled on Aug 3. Is there anything blocking it that I can address?

One thing I'd like to rule out as a source of confusion: my Jul 23 comment claimed this PR also resolves #637 — that was wrong, and I corrected it on Jul 27. #611 is backend-only (`.gitignore`, `api.rs`, `worker.rs`, `setting.rs`); the #637 fix lives in the frontend `guardTranscriptionModel` change over in #610/#536. The three fixes here (camelCase API-key binding, `chunk_duration` use-after-move, and the hardcoded `provider='parakeet'` corrupting saved settings) are unaffected by that correction and stand on their own.

The branch is still conflict-free against devtest, and there are no migrations. Happy to rebase, split, or adjust anything the review needs.

safvanatzack:
Hi, apologies for the the delay
As we internally did preliminary review on so many pr's and at the same time we got in to other serious work, we couldn't complete the review process

Will keep you posted
Again sorry for the delay, and thank you for your patience

PR #639, opened 2026-07-16
fix(recording): respect selected transcription provider

### LINKED ISSUE #637: [BUG] Recording blocked with "Transcription model not ready" when using Whisper provider (hardcoded Parakeet pre-check)
## Bug Description

When a user selects **Whisper** as the transcription provider and downloads a Whisper model (e.g. `large-v3-turbo`), clicking **Start Recording** still shows:

> **Transcription model not ready** — Please download a transcription model before recording.

The user is then pushed into the model-selector modal, even though a valid Whisper model is present on disk and the DB config says `provider = localWhisper, model = large-v3-turbo`.

## Root Cause

`frontend/src/hooks/useRecordingStart.ts` gates recording start on a **hardcoded Parakeet-only readiness check**, regardless of the user’s selected provider:

```ts
// Line ~52-62
const checkParakeetReady = useCallback(async (): Promise<boolean> => {
  try {
    await invoke(parakeet_init);
    const hasModels = await invoke<boolean>(parakeet_has_available_models);
    return hasModels;
  } catch (error) {
    console.error(Failed to check Parakeet status:, error);
    return false;
  }
}, []);
```

It is called from both the manual start handler (line ~88) and the auto-start effect (line ~157). When the user picks Whisper, `parakeet_has_available_models` returns `false` because no Parakeet ONNX bundle is on disk → the recording is blocked even though Whisper is fully set up.

The same file also checks download progress via `parakeet_get_available_models` only (line ~67), missing Whisper download state.

The backend-side `whisper_validate_model_ready_with_config` in `frontend/src-tauri/src/whisper_engine/commands.rs` already respects the user’s configured provider/model, so this is purely a frontend gating regression.

## Steps to Reproduce

1. Fresh install of Meetily v0.4.0 (macOS)
2. Settings → Transcription:
   - Provider: **Whisper (Local)**
   - Model: **large-v3-turbo** — download and wait until finished
3. Confirm model exists on disk:
   ```
   ~/Library/Application Support/com.meetily.ai/models/ggml-large-v3-turbo.bin  (~1.5 GB)
   ```
4. Do **not** download any Parakeet model
5. Click **Start Recording** on the home page

**Result:** Toast: *"Transcription model not ready — Please download a transcription model before recording."* + modal opens forcing Parakeet setup.

**Expected:** Recording starts using the configured Whisper model.

## Suggested Fix

Make the pre-check provider-aware. Something like:

```ts
const checkTranscriptionModelReady = useCallback(async (): Promise<boolean> => {
  try {
    // Read the users configured provider from transcript config
    const config = await invoke<{ provider: string; model: string } | null>(api_get_transcript_config);
    const provider = config?.provider ?? parakeet;

    if (provider === localWhisper) {
      // Use the existing backend command that already respects the users config
      await invoke<string>(whisper_validate_model_ready);
      return true;
    }

    if (provider === parakeet) {
      await invoke(parakeet_init);
      return await invoke<boolean>(parakeet_has_available_models);
    }

    // Cloud / API-based providers - no local model needed
    return true;
  } catch (err) {
    console.error(Failed to check transcription model readiness:, err);
    return false;
  }
}, []);
```

Similarly, `checkIfModelDownloading` should consult both `parakeet_get_available_models` and the Whisper download state depending on the active provider.

## Workaround

Users can unblock recording by downloading **any** Parakeet model (~670 MB) in addition to their Whisper model. The pre-check then passes and the actual transcription still runs on Whisper (because the backend consults `transcript_settings.provider = localWhisper`).

## Environment

- macOS 26.4 (Apple Silicon, M5)
- Meetily v0.4.0
- Transcript config in DB: `provider=localWhisper, model=large-v3-turbo`
- Whisper model on disk: `ggml-large-v3-turbo.bin` (1.5 GB, size validation passes)

## Related

Issue #509 documents a separate frontend gating bug in the same area (`AudioBackendSelector.tsx`) — same class of problem: UI logic hardcoded against one specific backend rather than respecting the users configuration.

tianrking:
## Summary

Fixes #637 by making the recording preflight use the configured local transcription provider instead of always checking Parakeet.

- `localWhisper` now initializes and checks Whisper models.
- `parakeet` retains its existing Parakeet checks.
- Download-in-progress detection uses the matching provider's model list.
- Unsupported providers fail explicitly rather than silently falling back to Parakeet.

## Root cause

All three recording entry points in `useRecordingStart` called only `parakeet_init` and `parakeet_has_available_models`. A user with a downloaded Whisper model but no Parakeet bundle was blocked before the Rust recording command could validate the selected Whisper configuration.

## Validation

- `bun test frontend/tests/lib/transcription-model-readiness.test.ts` — 4 passed
- `corepack pnpm build` in `frontend/` — production build completed successfully
- `git diff --check`

### DIFF
--- a/frontend/src/hooks/useRecordingStart.ts
+++ b/frontend/src/hooks/useRecordingStart.ts
@@ -7,11 +7,20 @@ import { useRecordingState, RecordingStatus } from '@/contexts/RecordingStateCon
 import { recordingService } from '@/services/recordingService';
 import Analytics from '@/lib/analytics';
 import { showRecordingNotification } from '@/lib/recordingNotification';
+import {
+  getProviderCommands,
+  hasDownloadingModel,
+  type ModelWithStatus,
+} from '@/lib/transcription-model-readiness';
 import { toast } from 'sonner';
 
 interface UseRecordingStartReturn {
   handleRecordingStart: () => Promise<void>;
   isAutoStarting: boolean;
+}
+
+interface TranscriptConfig {
+  provider?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


safvanatzack:
Thanks, this correctly fixes the P0 issue in #637.

The recording preflight now respects the persisted provider, so Local Whisper users with a valid Whisper model are no longer blocked by an unrelated Parakeet check. The fix is narrow, covers the manual, sidebar-direct, and auto-start paths, and keeps Parakeet behavior intact.

One follow-up to track separately: this change exposes an existing Whisper model-status limitation during an active download. A start attempt can trigger a disk rescan that reports an early Whisper download as missing instead of downloading, so the model selector may reopen without reliably showing active progress. Recording remains safely blocked; this is a UX/status-accuracy issue, not a correctness or recording-safety concern.

I’m comfortable accepting this PR for the #637 fix.

Will be creating a separate issue addressing the follow up

Thank you

sujithatzackriya:
Hey Safvan, thanks for the review. Merged the PR. @tianrking, thanks for the contribution. Hope we could make Meetily better for you throughout.

PR #641, opened 2026-07-16
feat: add speaker diarization with local and Azure routes

execbbbr:
## Summary
- add optional `speaker` propagation end-to-end (provider result, transcript update events, DB persistence, API responses, frontend types/state, transcript UI, and transcript copy output)
- add diarization settings with provider switch (`local` batch post-processing and `azure` realtime route) plus Azure key/region configuration fields
- implement local batch diarization hook for import/retranscription and Azure realtime streaming diarization client wiring in `worker.rs` with graceful `speaker=None` fallback

## Test plan
- [x] `pnpm --dir frontend run build` *(fails in this environment: missing module `@tiptap/pm/model` because frontend `node_modules` is not present)*
- [x] `cargo check --manifest-path frontend/src-tauri/Cargo.toml` *(fails due to known environment blocker: missing `binaries/llama-helper-aarch64-apple-darwin` resource in build script)*
- [x] `cargo test --manifest-path frontend/src-tauri/Cargo.toml` *(fails for the same `llama-helper` build-script resource blocker before tests can execute)*

## Notes
- Azure route was implemented as realtime streaming integration into the existing `transcript-update` pipeline in `audio/transcription/worker.rs`.
- Local route remains batch/post-recording via import and retranscription paths.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1508,6 +1508,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "data-encoding"
+version = "2.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4ae5f15dda3c708c0ade84bfee31ccab44a3da4f88015ed22f63732abe300c8"
+
+[[package]]
 name = "dbus"
 version = "0.9.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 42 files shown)


execbbbr:
Opened against the wrong repository by mistake — this belongs on my fork. Closing. Apologies for the noise.

PR #652, opened 2026-07-18
perf: verified speed optimizations across whisper, parakeet, llama-helper, audio and DB

gmelekos:
## Summary

A set of adversarially-verified performance fixes for Apple Silicon (and general) speed, each confirmed against the vendored crate sources (whisper-rs 0.13.2 / whisper.cpp, ort 2.0.0-rc.10, ebur128 0.1.10, llama-cpp-2 0.1.146) before implementation:

### Whisper (~2x faster transcription under default settings)
- `set_n_threads` from the adaptive hardware config in both transcription paths — the computed value was silently discarded, so whisper.cpp ran at its default 4 threads regardless of core count
- One reusable `WhisperState` per loaded model instead of re-running Metal backend init + KV-cache allocation + a doomed CoreML load attempt on every chunk
- Session-level cache of the auto-detected language: with the default auto/auto-translate setting, whisper.cpp was running a **full extra encoder pass for language detection on every chunk**; the cache is only trusted from chunks ≥2s that produced output
- Fix for an edition-2021 same-task deadlock in `load_model` (the `if let` scrutinee read guard was held across `unload_model()`'s write lock on the same RwLock) + regression test

### Parakeet (default engine)
- Model stays resident across recordings — stop/start was re-parsing the 652MB int8 encoder with Level3 graph optimization every time (multi-second start latency)
- The resident model is re-validated against the transcript config so settings changes still switch models
- Model load moved to `spawn_blocking`; decoder_joint/preprocessor ORT sessions switched to sequential execution with 1 intra-op thread (they run tiny per-step graphs), encoder unchanged
- Same edition-2021 `load_model` deadlock fixed here too

### llama-helper (built-in AI summaries)
- Per-request right-sized `n_ctx` (prompt tokens + max_tokens, capped) instead of always allocating the full 32k-token KV cache (~4-5GB per generation call); `n_batch` capped at 2048 with chunked prompt decode
- GPU layer offload budgeted at the right-sized context and re-budgeted from real token counts — 4B models now fully offload on 16GB machines

### Audio pipeline
- EBU R128 normalizer: `Mode::HISTOGRAM` — the default queue mode stores one block energy per 100ms with unbounded history and rescans it ~94x/sec **inside the real-time mic callback**, a cost growing linearly with recording duration; gain recomputation throttled to ~500ms

### Database / UI
- New composite index `transcripts(meeting_id, audio_start_time)` — every per-meeting query previously full-scanned the whole transcripts table (SQLite does not index FK columns)
- `api_get_summary` (polled every 5s during generation) now fetches only the meetings row instead of loading every transcript of the meeting to read its title
- Transcript search: redundant `LOWER()` dropped, per-meeting dedup with deterministic recency ordering + LIMIT, 300ms debounce with stale-response dropping in the sidebar, and a stuck `isSearching` fix

### Build
- `clean_build.sh` now builds the llama-helper sidecar in release mode before bundling — previously a debug (-O0) sidecar left behind by `dev-gpu.sh` could silently ship inside the production .app; also guards GPU features llama-helper doesn't define and fixes the trailing bare `sleep` that made the script exit non-zero on success

## Testing
- `cargo check --workspace` clean
- New regression test `switching_models_does_not_deadlock` passes (times out on the pre-fix code)
- Draft until a live end-to-end recording test completes

### DIFF
--- a/frontend/clean_build.sh
+++ b/frontend/clean_build.sh
@@ -51,7 +51,82 @@ pnpm run build
 
 # Set environment variables for the build
 
+# Build the llama-helper sidecar in release mode before bundling, so the
+# production app never ships a stale debug binary left behind by dev-gpu.sh
+echo "Detecting GPU features for llama-helper..."
+if [ -z "$TAURI_GPU_FEATURE" ]; then
+    TAURI_GPU_FEATURE=$(node scripts/auto-detect-gpu.js)
+fi
+
+# Note: llama-cpp-2 does NOT support coreml, only metal/cuda/vulkan
+# So for macOS Apple Silicon (which returns 'coreml' for Whisper), use 'metal' for llama-helper
+HELPER_FEATURES=""
+if [ -n "$TAURI_GPU_FEATURE" ] && [ "$TAURI_GPU_FEATURE" != "none" ]; then
+    LLAMA_FEATURE="$TAURI_GPU_FEATURE"
+    if [ "$LLAMA_FEATURE" = "coreml" ]; then
+        LLAMA_FEATURE="metal"
+        echo "Note: llama-cpp-2 doesn't support CoreML, using Metal instead"
+    fi
+    case "$LLAMA_FEATURE" in
+        metal|cuda|vulkan)
+            HELPER_FEATURES="--features $LLAMA_FEATURE"
+            ;;
+        *)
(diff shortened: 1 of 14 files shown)


gmelekos:
Withdrawing for now — keeping this work local until further validation.

PR #666, opened 2026-07-22
Add Google Calendar integration and Google Meet auto-detect

debasisnishank:
## Summary
- Connects Meetily to Google Calendar (OAuth 2.0 loopback flow, `calendar.readonly` scope) and matches recorded meetings against calendar events by time window, tagging matched meetings with the real event title, attendee list, and Google Meet link. That context is fed into the summary prompt so generated summaries can reference the actual agenda/attendees.
- Adds an opt-in "Auto-detect Google Meet calls" toggle that watches on-screen browser window titles (macOS only) for the pattern used while a Meet call is active, and auto-starts/stops Meetily's existing recording pipeline. No bot ever joins the call — this only triggers the same local audio capture a manual "Record" click would.
- New Settings → Calendar tab for connecting an account and enabling auto-detect.

## Known limitations (please read before reviewing)
- **Auto-detect is macOS-only** in this iteration. It's implemented via `CGWindowListCopyWindowInfo` (Quartz Window Services), gated behind the real macOS Screen Recording permission (distinct from this app's existing "Audio Capture" permission). No Windows/Linux equivalent yet — the settings toggle is a no-op on those platforms.
- **Window-title matching is heuristic, not a guarantee.** It pattern-matches browser tab titles (e.g. Chrome renders an active call as `"Meet – <call-code>"`) — it can miss calls if a browser doesn't reflect call state in the title, and can't perfectly distinguish an active call from a Meet tab sitting on a non-call page. It's opt-in and off by default for this reason.
- **The OAuth consent screen used for development is in Google's "Testing" mode**, restricted to specific test users, so it can't be tested end-to-end by reviewers out of the box. Each user brings their own Google Cloud OAuth client (Client ID + Secret entered in Settings → Calendar) — there's no shared/embedded client secret — so this isn't blocking for real usage, just for a reviewer trying it without their own GCP project.
- Calendar metadata enrichment only applies to meetings that overlap an event on the *connected* Google account's calendar; unrelated meetings just keep their generic auto-generated title with no error.

## Test plan
- [x] `cargo check` passes cleanly (both new modules, no new warnings beyond pre-existing ones)
- [x] `pnpm tsc --noEmit` passes cleanly
- [x] Unit tests for the calendar event time-window matcher (`calendar/matcher.rs`)
- [x] Manually verified end-to-end on macOS: OAuth connect flow (browser consent → token exchange → persisted), auto-detect correctly identifying a live Google Meet call, auto-start recording (mic + system audio), auto-stop ~20s after leaving the call, transcript/audio saved correctly
- [ ] Not yet tested: Calendar metadata actually landing on a meeting scheduled through Calendar (mechanism verified by code review / DB inspection, but not observed live end-to-end in this branch)
- [ ] Not tested on Windows/Linux (auto-detect is intentionally macOS-only; Calendar integration should be platform-agnostic but wasn't verified on those platforms)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3620,6 +3620,7 @@ dependencies = [
  "chrono",
  "cidre",
  "clap 4.6.1",
+ "core-foundation 0.9.4",
  "core-graphics 0.23.2",
  "cpal",
  "criterion",
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -169,6 +169,7 @@ anyhow = "1.0"
 time = { version = "0.3", features = ["formatting"] }
 reqwest = { version = "0.11", features = ["multipart", "json"] }
 core-graphics = "0.23"
+core-foundation = "0.9"
 cidre = { git = "https://github.com/yury/cidre", rev = "a9587fa", features = ["av"] }
 dasp = "0.11.0"
 futures-channel = "0.3.31"
--- a/frontend/src-tauri/migrations/20260721000000_add_calendar_integration.sql
+++ b/frontend/src-tauri/migrations/20260721000000_add_calendar_integration.sql
@@ -1,0 +1,14 @@
+-- Migration: Add Google Calendar integration support
+
+-- Calendar-derived metadata on a meeting, written after a match against a Google Calendar event
+ALTER TABLE meetings ADD COLUMN calendar_event_id TEXT;
+ALTER TABLE meetings ADD COLUMN calendar_attendees TEXT; -- JSON array of {name, email}
+ALTER TABLE meetings ADD COLUMN calendar_meet_link TEXT;
+ALTER TABLE meetings ADD COLUMN calendar_start_time TEXT;
+ALTER TABLE meetings ADD COLUMN calendar_end_time TEXT;
+
+-- Google OAuth token bundle: {client_id, client_secret, access_token, refresh_token, token_expiry, scope}
+ALTER TABLE settings ADD COLUMN googleCalendarConfig TEXT;
+
+-- Phase B toggle: auto-start/stop recording on detected Google Meet calls
+ALTER TABLE settings ADD COLUMN autoDetectMeetEnabled INTEGER DEFAULT 0;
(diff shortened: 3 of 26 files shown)


debasisnishank:
Thanks for the review — went through all 6 comments:

**Fixed (pushed in d371203):**
1. Added `conferenceDataVersion=1` to the Calendar API call — without it, Google omits `conferenceData` entirely, so this was silently breaking all Meet-link matching. Good catch, this was a real functional bug.
2. OAuth redirect `code` param now parsed via `url::form_urlencoded::parse` instead of a raw substring split, so percent-encoded codes decode correctly.
3. Calendar matching now uses one consistent wide window for both fetching and scoring instead of re-filtering with a narrow ±15min window around `meeting.created_at` (which is set when the transcript saves, near the recording's end, not when it started) — the narrow window could miss the real event for longer meetings.
4. The `<meeting_metadata>` block fed into the summary prompt now includes the Meet link and scheduled end time, which were missing despite being part of the stated goal.
5. The missing-permission warning in the auto-detect poller is now rate-limited to once every 5 minutes instead of every 5-second poll.

**Not changed (6, plaintext secrets in `settings`):** valid concern, but not something specific to this PR — every other provider's API key (OpenAI, Anthropic, Groq, etc.) is already stored the same way in this same table. Fixing it properly would mean moving to OS-keychain-backed storage for all secrets across the app, which is a larger, separate change than this PR's scope. Happy to open a follow-up issue for that if it's wanted.

debasisnishank:
Correction to my earlier comment on item #1 (conferenceDataVersion).

I tested this directly against the real Calendar API (a real event with a Google Meet link, with and without \`conferenceDataVersion=1\` on the \`events.list\` request) and **\`conferenceData\` is returned identically either way**. The parameter appears to only affect write operations (\`events.insert\`/\`update\`/\`patch\`), not reads — so it was not actually the cause of calendar matching failing to find events, contrary to what I said earlier (echoing the original review comment without independently verifying it first, which I should have done before stating it as a confirmed bug).

The actual reason matching never surfaced anything in testing: there simply were no events on the connected calendar at the time — nothing to match against, unrelated to this parameter.

I'm leaving \`conferenceDataVersion=1\` in place since it's harmless and is what Google's own API docs recommend for any client working with conference data, but withdrawing the claim that it was a confirmed critical bug. Apologies for the inaccurate claim — flagging it rather than letting it stand uncorrected.

PR #671, opened 2026-07-25
Add Atlas Cloud summary provider

binyangzhu000-sudo:
## Summary
- add Atlas Cloud as a summary model provider using the existing OpenAI-compatible chat completions path
- store Atlas Cloud API keys in a dedicated settings column and migration
- expose Atlas Cloud in the model settings UI with default text models

## Validation
- pnpm install --frozen-lockfile --ignore-scripts
- pnpm run build
- git diff --check
- git diff --cached --check
- local Atlas API reference confirms `https://api.atlascloud.ai/v1` and the default model IDs

## Notes
- `pnpm exec tsc --noEmit` currently fails on an existing test-only `bun:test` type resolution issue in `tests/lib/blocknote-markdown.test.ts`
- focused Rust tests could not run locally because `whisper-rs-sys` requires `cmake`, which is not installed in this environment
- live Atlas model catalog returned HTTP 403 / 1010 from this environment, so the model IDs were checked against the local API reference instead

No README or documentation changes.


binyangzhu000-sudo:
Synced with `main` (105 commits). Single conflict, at the tail of `frontend/src-tauri/src/summary/llm_client.rs`, where this branch's `#[cfg(test)]` module met an upstream edit to the same region. The test module is kept.

I checked the auto-merged side rather than trusting the clean merge. Of the nine files this branch touches, upstream changed two, both unrelated:

- `summary/mod.rs` — upstream trimmed two re-exports (`clean_llm_markdown_output`, `generate_meeting_summary`); this branch only edits the module doc comment there, so they do not interact.
- `contexts/ConfigContext.tsx` — upstream added `isModelConfigLoading`; the Atlas Cloud entries in that file are intact.

The `LLMProvider::AtlasCloud` variant, its `from_str` aliases (`atlascloud` / `atlas-cloud` / `atlas`), the request-builder arm and the `provider_name` arm all survive.

**On verification, so this isn't overstated:** `cargo check --tests` cannot run on this machine. `whisper-rs-sys` builds whisper.cpp through `cmake`, which isn't installed here, and I didn't want to install a system package to route around it. What I did confirm statically is that the restored test still type-checks by construction: `LLMProvider` derives `Debug + PartialEq` so `assert_eq!` applies, `from_str` is an inherent `pub fn` returning `Result<Self, String>`, and `provider_name` is a private module-level fn that the child `tests` module reaches through `super::`. CI, or any machine with cmake, should be the real gate here.

No conflicts remain.

PR #673, opened 2026-07-27
fix(diarization): :bug: speaker identification rework — integrity fixes + measurement harness

Nader-Awad:
## Why

The Speaker Identification pane was flagging 7 of 16 saved voices as "may clash". Investigation found the warnings are largely **correct** — recognition quality had collapsed, and two profiles literally held byte-identical vectors.

Measured on a real profile store (16 profiles, 52 exemplars):

| metric | measured | healthy |
|---|---|---|
| same-speaker cosine | 0.8628 | — |
| different-speaker cosine | 0.8325 | — |
| **margin** | **0.0303** | > 0.3 |
| cosine to global mean | 0.9154 | < 0.5 |
| EER | **~34%** | < 5% |
| false-reject at the 0.60 auto-adopt bar | **97%** | < 10% |

## Root cause

Not the embedding model — that measures healthy (0.85 on guaranteed-same-speaker audio, deterministic, shift-invariant). Three plausible causes were tested and **falsified**: input scaling (`normalize_samples=0` vs f32 — cancels exactly under CMN), mixer clipping (0.00% of samples), and centroid averaging (actually helps).

The real defect is a four-stage amplification:

1. `map_local_speakers_to_profiles` has no exclusivity constraint, so several distinct local clusters get the same name
2. `relabel_and_merge_centroids` merges them into one entry — one recording had a single `Ralf` holding 304 segments, 53% of the meeting
3. that blended centroid is enrolled as the profile
4. blended profiles legitimately collide, and the detector reports it accurately

Confirmed independently: clusters with internal coherence **0.927** (tighter than the same-speaker anchor, so almost certainly one person) carry five different names, and same-label similarity is **perfectly flat** across time gaps (decay −0.009), ruling out voice drift.

## What's in this PR

**Phase 0 — `speaker_eval` harness** (dev-only, behind the `dev-eval` feature). Drives the real diarization modules rather than a reimplementation. Establishes a label-free anchor (one utterance split in half is necessarily the same speaker) that makes every other number interpretable.

**Phase 1 — integrity fixes**
- exemplar provenance migration (nullable; pre-existing rows untouched)
- withdraw the prior enrollment on rename — fixes the confirmed Alice-holds-Camilia defect
- refuse to store one recording under two names
- report cross-profile duplicates for review; delete only on explicit request
- split clash detection: mean-linkage for confusability (count-robust), near-1.0 comparison for corruption

Effect on the settings pane: 7 flagged profiles drop to 2 (genuinely similar), with the two real corruptions moved to a precise duplicates list.

## Still to come

Phase 2 (exclusive Hungarian assignment — the highest-leverage fix), Phase 3 (multi-exemplar enrollment + coherence gate), Phase 4 (rebuild wizard that re-uses the ~361 min of manually labelled audio). Design: `docs/superpowers/specs/2026-07-28-speaker-identification-design.md`.

## Testing

319 pass. `audio::device_detection::tests::test_builtin_mic_detection` fails, but is **pre-existing and unrelated** — that file is byte-identical to HEAD in this branch.

New coverage: rename withdrawal (including a second correction), legacy rows without provenance left alone, cross-profile duplicate detection, self-conflict exclusion, summary refresh on delete, and that clash scores no longer inflate with exemplar count.


Nader-Awad:
Opened against the wrong repository by mistake — this belongs on the fork. Closing.

Issue #399, opened 2026-03-12
[BUG]Imported meeting causing out of sync transcript segment with audio(only the last segment, edge case)


safvanatzack:
## Bug Description
In import meetings
the last transcript segment's timestamp is out of the length of the whole meeting(Not consistent Bug)

## Current Behavior
the last transcript segment's timestamp is out of the length of the whole meeting(Not consistent Bug)

## Expected Behavior
the last speech timestamp should be matching the actual speech in the audio(should not exceeds from the duration of meeting)

## Steps to Reproduce
Try to import a meeting where the the last audio segment didn't have a break(vad ending shouldn't be detected)
Meaning the error is occuring if vad is not detecting the last audio's end

## Environment
- OS: [e.g., macOS, Windows, Linux]
- Browser: [e.g., Chrome, Firefox, Safari]
- Version: [e.g., 1.0.0]
- Node Version: [e.g., 18.0.0]
- npm/pnpm Version: [e.g., 8.0.0]

## Screenshots/Videos
for a 30 minutes long meeting the last transcripts timestamp is 50mins+
<img width="725" height="322" alt="Image" src="https://github.com/user-attachments/assets/18febef6-01c0-4020-aca2-b21d211da029" />

## Error Messages
[If applicable, paste any error messages you're seeing]

## Additional Context
[Add any other context about the problem here]

## Possible Solution
[If you have suggestions on how to fix the issue, please describe them here]

## Checklist
- [x] I have searched for similar issues
- [x] I have provided all required information
- [x] I have included screenshots/videos if applicable
- [x] I have included error messages if applicable

safvanatzack:
Fixed in v0.4.1 through #771.

The VAD final-flush path now keeps timestamps session-absolute and ensures the final segment satisfies:

```text
start <= end <= decoded audio duration
```

Targeted regression coverage was added for the forced-final-segment case that could previously produce a timestamp beyond the imported audio duration.

Closing as resolved in v0.4.1.

Issue #675, opened 2026-07-28
[BUG] Short Whisper transcripts are dropped by text-length confidence heuristic


themaddoxed:
## Bug description

Live Whisper transcription can silently drop short but valid utterances such as acknowledgements, confirmations, and brief replies.

## Current behavior

`WhisperEngine::transcribe_audio_with_confidence` derives a confidence-like value from the UTF-8 byte length of the decoded segment. The live transcription worker then rejects non-empty results below a `0.3` threshold.

Several common English replies fall below that threshold:

- `OK` receives approximately `0.12`.
- `Yes` receives approximately `0.13`.
- `No` receives approximately `0.12`.
- `Sure` receives approximately `0.14`.
- `Got it` receives approximately `0.16`.
- `Thanks` receives approximately `0.16`.

All of these can be skipped even though Whisper returned non-empty text. The derived value is not an acoustic or token probability.

## Expected behavior

A non-empty transcript returned after VAD and Whisper's own no-speech handling should be emitted. The current confidence value may remain available as metadata, but it should not decide whether the transcript is discarded.

## Proposed fix

- Base transcript emission on non-empty content rather than the uncalibrated confidence value.
- Keep empty and whitespace-only results filtered out.
- Add regression coverage for several short English utterances and a whitespace-only result.

## Affected revision

Reproduced on `devtest` at `0bedd8b`.

## Checklist

- [x] Searched for a similar issue
- [x] Identified the root cause in the live transcription worker
- [x] Added a failing regression test before the fix

Hus-Mek:
Confirming this from a second direction, since I nearly shipped a change that would have compounded it.

While diagnosing poor Arabic transcription I added a hallucination filter with a standalone-filler denylist — `"thank you"`, `"okay"`, `"thanks"`, `"bye"`, plus Arabic `شكرا` and its orthographic variants — dropping them when they made up a whole segment. Whisper really does hallucinate those on near-silent input (the meeting I looked at had 14 standalone `شكرا` segments in 26 minutes), so it seemed justified.

[...]

PR #679, opened 2026-07-30
fix(vad): stop fragmenting live speech into sub-4s ASR requests, and filter hallucinations on every engine

Hus-Mek:
> **Rebased onto `devtest` and scoped to the VAD fix** (per review, 2026-08-29). This PR now
> contains only the two VAD/segmentation changes below. The shared cross-provider
> `text_cleanup` rewrite that was previously bundled here (hallucination filtering across
> Whisper, Parakeet, and remote providers) has been removed from this branch and lands as
> follow-up PR #747, so the VAD behavior can be validated without mixing in a
> transcript-output policy change. The `0.3` confidence gate removed by #681 lives in
> `worker.rs`/`whisper_engine.rs` and is untouched here, so this PR does not reintroduce it.
>
> **Updated after live validation (2026-09-01):** the live path now uses **500 ms** redemption
> (the established Meetily Pro live policy); the batch paths keep **2000 ms**. See defect 1.

## What this fixes

Live VAD segmentation quality. Diagnosed from a real 26-minute Arabic meeting; both causes are
provider-agnostic, so local Whisper, Parakeet, and cloud providers are equally affected.

### 1. Live VAD redemption time was 400ms, fragmenting speech at mid-sentence breaths

`pipeline.rs` passed `400`, while `import.rs:58` and `retranscription.rs:52` both define `VAD_REDEMPTION_TIME_MS = 2000`. Supporting evidence that 400 was not deliberate:

- the platform ternary was dead: `if cfg!(target_os = "macos") { 400 } else { 400 }`, under a comment claiming the two platforms differ
- `vad.rs` carried a comment asserting "Use full redemption_time from pipeline (2000ms)" - false for the live path
- `vad.rs` already contains a test named `test_vad_400ms_vs_2000ms_segmentation` whose docstring states that 400ms causes excessive fragmentation

Redemption time decides how long a silence must last before a speech segment closes, and **every segment becomes its own ASR request**. Measured on the real recording at 400ms:

| metric | value |
|---|---|
| requests for 26 min of audio | 322 |
| median request length | 3.49 s |
| requests under 3 s | 42 % |
| requests that reached 25 s | 1 of 321 |
| boundaries in the 0.42-0.75 s range | **152 of 321 (47 %)** |

That last row is the point: nearly half the segment boundaries were mid-sentence breaths that a longer redemption bridges.

Whisper is a fixed **30-second-window** model. Below that it zero-pads the window and leans on its language-model prior, which was trained on web subtitles, so short clips come back as memorised boilerplate instead of speech. The VAD also pads every segment by 300 ms + 400 ms, a fixed cost that dominates short segments:

| segment length | share that is padding silence | output is a single word |
|---|---|---|
| ~1 s | 60 % | 81 % |
| ~2 s | 34 % | 13 % |
| ~3 s | 22 % | 6 % |
| ~5 s | 15 % | 0 % |
| ~7 s and over | 9 % | 0 % |

**Live-path policy (updated after review validation):** an earlier revision of this PR aligned the live path to the batch value (2000 ms). Maintainer live testing with continuous system audio showed that is not viable for the live path: VAD kept one speech segment open for an extended period, no live transcript was emitted, and the accumulated-speech buffer crossed its warning threshold. The live path therefore now uses **`VAD_REDEMPTION_TIME_MS = 500`**, matching the established Meetily Pro live policy, while `import.rs` and `retranscription.rs` keep **2000 ms** - batch has no latency requirement, so live and batch deliberately diverge. 500 ms addresses the fragmentation scope of this PR; it does not by itself guarantee bounded transcript delivery during continuous speech where no qualifying silence occurs - bounded live segments, forced-boundary correctness, and a maximum live-transcript-update interval are tracked separately in **#756**.

### 2. `speech_start_sample` double-counted the session position

The buggy live path was:

```rust
self.speech_start_sample = self.processed_samples + (timestamp_ms * 16000 / 1000);
```

Both operands are already session-absolute. `processed_samples` increments forever, and silero computes `SpeechStart.timestamp_ms` as `processed_duration().saturating_sub(pre_speech_pad)` - verified in `silero-rs/src/lib.rs:302-305`, emitted at `321-324`. Summing them puts the start position at roughly 2x the truth, and the error grows with how late in the session the utterance begins.

The only reader is the force-end branch in `flush()`, so it escaped once per recording, on the final segment. On a 1539.7 s recording that segment was stored with `audio_start_time = 3083.1 s` (ratio 2.0007) and sorted to the end of the transcript. Every other segment tracked wall-clock at ratio ~1.01, which is why it looked like a one-off rather than drift. Relevant to #399.

`devtest` has since applied an equivalent literal fix to the `SpeechStart` path; this branch keeps the same correction expressed against the `VAD_SAMPLE_RATE` module constant, and retains the regression coverage below.

## What changed

- `pipeline.rs` - named `VAD_REDEMPTION_TIME_MS = 500` constant (live policy, per Meetily Pro), documented against the batch paths' 2000 ms; dead platform ternary removed; `test_live_vad_redemption_matches_pro_policy` locks the value
- `vad.rs` - session-absolute timestamp fix; `VAD_SAMPLE_RATE` promoted to a module constant; the false comment corrected; two regression tests added
- `import.rs`, `retranscription.rs` - unchanged at 2000 ms; stale doc comments (which described the live path as 400 ms) updated

## Verification

Focused VAD tests, on the current branch:

```
$ cargo test --lib audio::vad

running 7 tests
test audio::vad::tests::test_flush_segment_timestamps_stay_within_audio_duration ... ok
test audio::vad::tests::test_speech_start_sample_never_exceeds_processed_samples ... ok
test audio::vad::tests::test_vad_continuous_processor_state_across_chunks ... ok
test audio::vad::tests::test_vad_cancellation ... ok
test audio::vad::tests::test_vad_large_file_progress ... ok
test audio::vad::tests::test_vad_chunked_vs_single_processing ... ok
test audio::vad::tests::test_vad_400ms_vs_2000ms_segmentation ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 208 filtered out; finished in 6.30s

$ cargo test --lib audio::pipeline

running 1 test
test audio::pipeline::tests::test_live_vad_redemption_matches_pro_policy ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 214 filtered out; finished in 0.01s
```

The two regression tests requested in review:

- `test_flush_segment_timestamps_stay_within_audio_duration` - a final unfinished utterance satisfies `start <= end <= input duration`.
- `test_speech_start_sample_never_exceeds_processed_samples` - the VAD start position never exceeds processed audio. Asserted directly rather than via silero's segmentation of synthetic audio, because the force-end path only runs when recording stops mid-utterance and the error is only large enough to violate the invariant when speech begins late in the session. The fixture is therefore leading-silence-then-speech; with the bug present it reports `start=41.67s vs processed=22.98s`.

Both were verified in **both directions** - each bug reintroduced to confirm the test fails, then the fix restored to confirm it passes.

Full suite (`cargo test --lib`): 212 passed; the only failure is `audio::device_detection::tests::test_calculate_buffer_timeout_bluetooth`, which **fails identically on unmodified `devtest`** (confirmed by stashing). Unrelated to this PR.

## Deliberately not included

- **The overlapping segment pairs are not a bug.** They are the arithmetic consequence of 700 ms of VAD padding against the redemption window; maximum overlap across all of them was exactly 0.28 s.
- **A minimum-duration floor before dispatch.** Once the timestamp fix lands, the shortest real segment measured was 0.97 s, so a 1 s floor is close to a no-op and risks dropping legitimate one-word replies.
- **Bounded live segments under continuous speech.** 500 ms redemption still requires a qualifying silence to close a segment; forced boundaries and a maximum live-transcript-update interval are tracked in #756.
- **`vad.rs::resample_to_16k`.** Its filter size is `sample_rate / (0.4 * sample_rate)` - the rate algebraically cancels to a fixed 5-tap moving average with no real anti-aliasing at 3:1, which reads like a serious bug. I ported that exact algorithm and re-ran the full meeting against ffmpeg's polyphase resampler with chunking and language held constant: 2229 words vs 2207, median `avg_logprob` -0.293 vs -0.334 (the naive version scored marginally *better*), and both recovered the proper nouns. Whisper's mel front-end weights energy below ~4 kHz heavily, so aliasing folded in above 8 kHz barely reaches the features. Worth fixing for hygiene, but it is not a quality fix and I did not want to bundle it here.
- **The cross-provider `text_cleanup` rewrite** (hallucination filtering across all engines, repetition ratio measured on original text, multi-word boilerplate denylist). Removed from this branch; now under review separately as #747.

### DIFF
--- a/frontend/src-tauri/src/audio/import.rs
+++ b/frontend/src-tauri/src/audio/import.rs
@@ -52,9 +52,10 @@ impl Drop for ImportGuard {
 }
 
 /// VAD redemption time in milliseconds - bridges natural pauses in speech
-/// Batch processing needs longer redemption (2000ms) than live pipeline (400ms)
-/// because the entire file is processed at once by VAD, and 400ms fragments
-/// speech at every natural sentence/topic pause (500ms-2s)
+/// Batch processing needs longer redemption (2000ms) than the live pipeline
+/// (500ms) because the entire file is processed at once by VAD with no
+/// latency requirement, and short redemption fragments speech at every
+/// natural sentence/topic pause (500ms-2s)
 const VAD_REDEMPTION_TIME_MS: u32 = 2000;
 
 /// Maximum file size: 20GB (prevents OOM and excessive processing time)
--- a/frontend/src-tauri/src/audio/pipeline.rs
+++ b/frontend/src-tauri/src/audio/pipeline.rs
@@ -13,6 +13,15 @@ use super::recording_state::{AudioChunk, AudioError, RecordingState, DeviceType}
 use super::audio_processing::{audio_to_mono, LoudnessNormalizer, NoiseSuppressionProcessor, HighPassFilter};
 use super::vad::{ContinuousVadProcessor};
 
+/// How long a silence must last before the VAD closes a speech segment, and
+/// therefore how long the audio clips handed to the ASR engine are.
+///
+/// Kept equal to `import.rs::VAD_REDEMPTION_TIME_MS` and
+/// `retranscription.rs::VAD_REDEMPTION_TIME_MS` so that live recording and
+/// offline re-transcription of the same audio segment identically. Raising this
+/// trades live-transcript latency for longer, more accurate ASR requests.
+const VAD_REDEMPTION_TIME_MS: u32 = 2000;
+
 /// Ring buffer for synchronized audio mixing
 /// Accumulates samples from mic and system streams until we have aligned windows
 struct AudioMixerRingBuffer {
@@ -719,16 +728,34 @@ impl AudioPipeline {
         // For now, we log it for monitoring and potential optimization
         let _ = (mic_device_name, mic_device_kind, system_device_name, system_device_kind);
 
-        // Create VAD processor with balanced redemption time for speech accumulation
-        // The VAD processor now handles 48kHz->16kHz resampling internally
-        // This bridges natural pauses without excessive fragmentation
-        // For mac os core audio, 900ms, for windows 400ms seems good
-
-        let redemption_time = if cfg!(target_os = "macos") { 400 } else { 400 };
-
-        let vad_processor = match ContinuousVadProcessor::new(sample_rate, redemption_time) {
+        // Create VAD processor. The VAD processor handles 48kHz->16kHz resampling
+        // internally.
+        //
+        // Redemption time is how long a silence must last before the VAD closes a
+        // speech segment, so it decides how long the audio clips handed to the ASR
+        // engine are. Conversational speech pauses constantly for breath and
+        // mid-sentence thought, and every pause longer than this becomes a segment
+        // boundary and therefore a separate transcription request.
+        //
+        // This was 400ms, which fragmented a 26-minute meeting into 322 requests with
+        // a median length of 3.5s. Whisper is a fixed 30-second-window model: below
+        // that it zero-pads the window and leans on its language-model prior, which
+        // was trained on web subtitles, so short clips come back as memorised
+        // boilerplate ("subscribe to the channel", "thank you") instead of speech.
+        // Measured on a real recording, 47% of segment boundaries sat in the
+        // 0.42-0.75s range that a longer redemption simply bridges.
+        //
+        // 2000ms matches what the offline paths already use (see
+        // `import.rs::VAD_REDEMPTION_TIME_MS` and
+        // `retranscription.rs::VAD_REDEMPTION_TIME_MS`), so live and batch
+        // transcription now segment identically. It costs up to ~1.6s of extra
+        // live-transcript latency at the end of each utterance.
+        let vad_processor = match ContinuousVadProcessor::new(sample_rate, VAD_REDEMPTION_TIME_MS) {
             Ok(processor) => {
-                info!("VAD-driven pipeline: VAD segments will be sent directly to Whisper (no time-based accumulation)");
+                info!(
+                    "VAD-driven pipeline: segments dispatched per speech burst (redemption_time={}ms)",
+                    VAD_REDEMPTION_TIME_MS
+                );
                 processor
             }
             Err(e) => {
--- a/frontend/src-tauri/src/audio/pipeline.rs
+++ b/frontend/src-tauri/src/audio/pipeline.rs
@@ -16,11 +16,15 @@ use super::vad::{ContinuousVadProcessor};
 /// How long a silence must last before the VAD closes a speech segment, and
 /// therefore how long the audio clips handed to the ASR engine are.
 ///
-/// Kept equal to `import.rs::VAD_REDEMPTION_TIME_MS` and
-/// `retranscription.rs::VAD_REDEMPTION_TIME_MS` so that live recording and
-/// offline re-transcription of the same audio segment identically. Raising this
-/// trades live-transcript latency for longer, more accurate ASR requests.
-const VAD_REDEMPTION_TIME_MS: u32 = 2000;
+/// Live-path policy: 500ms (matches the established Meetily Pro live policy).
+/// The batch paths (`import.rs` / `retranscription.rs`) use 2000ms instead —
+/// they have no latency requirement, so they optimize purely for ASR request
+/// length. The live path cannot: with continuous audio (e.g. a podcast played
+/// as system audio) a 2000ms redemption keeps one VAD segment open
+/// indefinitely, withholds live transcript emission, and overruns the
+/// accumulated-speech-buffer warning threshold. Bounded live segmentation
+/// during continuous speech is tracked separately in #756.
+const VAD_REDEMPTION_TIME_MS: u32 = 500;
 
 /// Ring buffer for synchronized audio mixing
 /// Accumulates samples from mic and system streams until we have aligned windows
@@ -743,13 +747,14 @@ impl AudioPipeline {
         // was trained on web subtitles, so short clips come back as memorised
         // boilerplate ("subscribe to the channel", "thank you") instead of speech.
         // Measured on a real recording, 47% of segment boundaries sat in the
-        // 0.42-0.75s range that a longer redemption simply bridges.
+        // 0.42-0.75s range that a longer redemption bridges.
         //
-        // 2000ms matches what the offline paths already use (see
-        // `import.rs::VAD_REDEMPTION_TIME_MS` and
-        // `retranscription.rs::VAD_REDEMPTION_TIME_MS`), so live and batch
-        // transcription now segment identically. It costs up to ~1.6s of extra
-        // live-transcript latency at the end of each utterance.
+        // 500ms is the live-path policy (see the constant's doc comment). The
+        // batch value (2000ms, `import.rs`/`retranscription.rs`) was tried here
+        // first, but under continuous system audio it kept a VAD segment open
+        // indefinitely and withheld live transcript emission, so live and batch
+        // deliberately diverge. Bounded live segments under continuous speech
+        // are tracked in #756.
         let vad_processor = match ContinuousVadProcessor::new(sample_rate, VAD_REDEMPTION_TIME_MS) {
             Ok(processor) => {
                 info!(
@@ -1103,5 +1108,17 @@ impl AudioPipelineManager {
 impl Default for AudioPipelineManager {
     fn default() -> Self {
         Self::new()
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_live_vad_redemption_matches_pro_policy() {
+        // Live path uses 500ms (Meetily Pro live policy) so transcript
+        // emission stays bounded under continuous audio; the batch paths
+        // (import.rs / retranscription.rs) use 2000ms. See #679 and #756.
+        assert_eq!(VAD_REDEMPTION_TIME_MS, 500);
     }
 }
--- a/frontend/src-tauri/src/audio/retranscription.rs
+++ b/frontend/src-tauri/src/audio/retranscription.rs
@@ -46,9 +46,10 @@ impl Drop for RetranscriptionGuard {
 }
 
 /// VAD redemption time in milliseconds - bridges natural pauses in speech
-/// Batch processing needs longer redemption (2000ms) than live pipeline (400ms)
-/// because the entire file is processed at once by VAD, and 400ms fragments
-/// speech at every natural sentence/topic pause (500ms-2s)
+/// Batch processing needs longer redemption (2000ms) than the live pipeline
+/// (500ms) because the entire file is processed at once by VAD with no
+/// latency requirement, and short redemption fragments speech at every
+/// natural sentence/topic pause (500ms-2s)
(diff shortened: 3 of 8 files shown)


Hus-Mek:
Pushed `d5abf16` — a self-correction, and it interacts with #675, so flagging it rather than letting a reviewer find it.

### What was wrong

The hallucination filter I added in `dbb4e24` included a standalone-filler denylist that dropped `"thank you"`, `"okay"`, `"thanks"`, `"you"`, `"bye"` and Arabic `شكرا` (all orthographic variants) when they constituted a whole segment.

Whisper genuinely does hallucinate those on near-silent input — that is what made the list tempting, and the meeting I diagnosed had 14 standalone `شكرا` segments in 26 minutes. But **nothing in the text distinguishes the two cases.** `شكرا` decoded from 300 ms of silence and `شكرا` decoded from someone actually saying it are byte-identical.

So the list traded a cosmetic problem (a stray "thanks" in the transcript) for a semantic one (a participant's answer silently missing). And **#675 is that exact failure already reported** — "OK", "Yes", "Thanks" dropped by the byte-length confidence heuristic. My filter would have made a filed bug worse from a second, independent direction. Bad trade, now reverted.

### What replaces it

Short-clip hallucination is handled where the evidence actually lives — the provider layer, gating on Whisper's own `avg_logprob` / `compression_ratio` from `verbose_json`. Those are acoustic-ish signals; the text is not. (That part is on the Groq provider in #610, since `groq_provider.rs` does not exist on `devtest`.)

Only **unambiguous multi-word boilerplate** is still filtered by text — `اشتركوا في القناة`, subtitler credits like `ترجمة نانسي قنقر`, `thanks for watching`, `like and subscribe`. No meeting contains those phrases, so there is no genuine reading to lose.

Degenerate repetition is still caught, now language-agnostically: the repetition ratio is measured on the **original** text rather than after collapsing. Measuring after collapsing hid it — a collapsed run leaves one word, and one word has a repetition ratio of zero. `شكرا شكرا شكرا شكرا شكرا` is still discarded, and the code no longer needs to know what the word means to do it.

### Tests

The two inverted tests are replaced by `keeps_short_acknowledgements_issue_675`, asserting that `OK`, `Yes`, `No`, `Sure`, `Got it`, `Thanks`, `Okay.`, `Bye` plus `شكرا`, `شكراً`, `شكرا لك`, `نعم`, `طيب`, `لا` and a tatweel-stretched `شكـــرا` all survive cleaning unchanged. 14 tests in the module, all passing.

Full suite unchanged at **202 passed, 1 failed** — `audio::device_detection::tests::test_calculate_buffer_timeout_bluetooth`, which fails identically on unmodified `devtest` (verified by stashing). Not from this PR.

### Note on #675

This PR no longer makes #675 worse, but it does not fix it either — the byte-length confidence heuristic in `WhisperEngine::transcribe_audio_with_confidence` and the `0.3` threshold in the worker are both untouched here. Happy to take that on separately if it is not already in progress; the proposed fix in #675 (emit on non-empty content, keep the confidence value as metadata only) reads correct to me and would compose cleanly with this PR.

safvanatzack:
Thanks @Hus-Mek  for the thoughtful work on this PR. The investigation into live VAD fragmentation, the 2-second redemption alignment, and the final-segment timestamp path is valuable; this is the strongest current OSS contribution in this area.

Our plan is to address the transcript-loss cluster in focused, independently reviewable pieces:

1. Preserve short valid transcripts: [#681](https://github.com/Zackriya-Solutions/meetily/pull/681) already removes the `0.3` transcript-confidence gate on `devtest`.
2. Stabilize live VAD segmentation: use this PR for the 2-second redemption/hangover behavior so normal conversational pauses do not split one utterance into short ASR requests.
3. Correct end-of-recording VAD behavior: keep the absolute start-timestamp fix so `flush()` cannot emit a trailing segment beyond the audio duration—relevant to [#399](https://github.com/Zackriya-Solutions/meetily/issues/399).
4. Review transcript cleanup independently: shared cleanup across Whisper, Parakeet, and remote providers changes output policy and should have its own focused PR.

Could you please rebase #679 onto the latest `devtest` and resolve the merge conflict?

For this rebased PR, please keep only the VAD-focused work:

- retain live `VAD_REDEMPTION_TIME_MS = 2_000`, aligned with import/retranscription;
- retain the absolute VAD start-timestamp correction;
- retain/add regression coverage proving:
  - a final unfinished utterance has `start <= end <= input duration`;
  - the VAD start position never exceeds processed audio.

Please move the shared cross-provider `text_cleanup` rewrite into a separate follow-up PR. It is useful work, but separating it lets us validate the VAD fix without mixing transcript-output policy changes across providers.

Please ensure the rebased branch integrates cleanly with #681 and does not reintroduce the old `0.3` confidence gate. Once rebased, run the focused Rust VAD tests and add the command/output to the PR description.

Thank you again. Splitting this work will make it substantially easier to review, test, and land the VAD improvements safely.

Hus-Mek:
Thanks @safvanatzack — the four-way split makes sense, and scoping this PR to the VAD fix is the right call. I'll restructure it exactly as you've laid out.

Plan for the rebased #679:

- Rebase onto latest `devtest` and resolve the conflict. I can see #681 already landed there (2026-08-27), so the `0.3` confidence gate is gone on the base — I'll confirm the rebase doesn't drag the old gate back in from my branch's history.
- Keep only the VAD-focused work: live `VAD_REDEMPTION_TIME_MS = 2_000` aligned with the import/retranscription path, and the absolute VAD start-timestamp correction so `flush()` can't emit a segment past the audio duration.
- Keep/add the regression coverage: a final unfinished utterance satisfies `start <= end <= input duration`, and the VAD start position never exceeds processed audio.
- Run the focused Rust VAD tests and add the command plus output to the PR description.

The shared cross-provider `text_cleanup` rewrite (repetition ratio on original text, multi-word boilerplate denylist) will move to its own follow-up PR so the output-policy change is reviewed on its own, independent of the VAD behavior. I'll open that once this one is clean.

Hus-Mek:
Done — #679 is now rebased onto `devtest` and scoped to the VAD fix, and the `text_cleanup` work has moved to #747 as agreed.

- **#679 (this PR):** VAD-only. `VAD_REDEMPTION_TIME_MS = 2000` aligned with the import/retranscription paths, the session-absolute `speech_start_sample` correction, and both regression tests. `cargo test --lib audio::vad` → 7 passed; command and output are in the updated description. The `0.3` gate removed by #681 is untouched.
- **#747 (follow-up):** the shared cross-provider hallucination filtering, reviewable on its own. It does not reintroduce the `0.3` gate.

Both branches integrate cleanly with the current `devtest`.

safvanatzack:
@Hus-Mek I manually tested the rebased branch and found a live-transcription issue with the current 2000 ms redemption setting.

With continuous system audio from a podcast, VAD kept one speech segment open and no live transcript segment was emitted for an extended period. The buffer grew past the 62.5 s warning threshold and began producing repeated accumulation warnings.

For comparison only, I locally changed the live value to 800 ms and repeated the same test. It emitted transcript segments, and the resulting transcript had no material missing content compared with the source video. However, the segments were still large and the accumulated-buffer warning still occurred.

The timestamp correction and its regression coverage still look correct. I am sharing this manual-test finding now; we will complete internal testing and follow up separately on the appropriate live VAD policy/value.

safvanatzack:
Thank you, @Hus-Mek, for following through on the rebase and for keeping this PR carefully focused. The VAD investigation, timestamp correction, and regression coverage are all appreciated.

One update from our live validation: `2000 ms` is appropriate for import and retranscription, but it is not suitable for the live path. With continuous system audio, it kept a VAD segment open for an extended period, withheld live transcript emission, and crossed the accumulated-speech-buffer warning threshold.

Could you please make the following final adjustments to this PR?

- Set the live `VAD_REDEMPTION_TIME_MS` to **`500` ms**, matching the established Meetily Pro live policy.
- Keep import and retranscription at **`2000` ms**, since batch and live paths have different latency requirements.
- Retain the session-absolute `speech_start_sample` correction and its regression coverage for final-segment timestamp bounds.
- Keep the cross-provider text-cleanup work separate, as already agreed.
- Update the PR description so it no longer presents `2000 ms` as the live-path policy.

The `500 ms` change addresses the current live-fragmentation scope. It does not, by itself, guarantee bounded transcript delivery during continuous speech, where VAD may remain active without a qualifying silence. That work is now tracked separately in **#756**, covering bounded live segments, forced-boundary correctness, and a maximum live-transcript-update interval.

Thanks again for the thoughtful work and for working with us to keep these changes safe and independently reviewable.

Hus-Mek:
Done — pushed `0ceee52` with the requested adjustments:

- **Live path is now `VAD_REDEMPTION_TIME_MS = 500`** in `pipeline.rs`, matching the Meetily Pro live policy. The constant's doc comment records why live and batch deliberately diverge (continuous system audio keeping a segment open, withheld emission, the accumulated-buffer warning) and points to #756 for bounded live delivery.
- **Import and retranscription stay at `2000` ms.** Their doc comments also had a stale "live pipeline (400ms)" reference, now corrected — no value change on either path.
- Added `test_live_vad_redemption_matches_pro_policy` locking the live value, alongside the existing `test_vad_redemption_time_constant` that locks batch at 2000.
- The session-absolute `speech_start_sample` correction and both timestamp regression tests are unchanged.
- The PR description no longer presents 2000 ms as the live-path policy; it now documents the 500/2000 split, your live-validation finding, and the #756 follow-up scope.

Verification on the current branch: `cargo test --lib audio::vad` → 7 passed; `cargo test --lib audio::pipeline` → 1 passed; full `cargo test --lib` → 212 passed with only the pre-existing `test_calculate_buffer_timeout_bluetooth` failure that also fails on unmodified `devtest`. Output is in the updated description.

Agreed on the #756 framing — 500 ms fixes the fragmentation scope here, but a qualifying silence is still required to close a segment, so continuous speech needs the forced-boundary/max-update-interval work tracked there. Happy to pick that up if useful.

safvanatzack:
Thank you, @Hus-Mek — I rechecked `0ceee52`. The 500 ms live / 2000 ms batch split is now correct, and the focused tests pass.

One timestamp issue remains: `flush()` counts zero-padding in `processed_samples`. The existing 23-second fixture therefore reports an end of **23,010 ms**. The test passes because it checks `start <= duration` and `end >= start`, but not `end <= duration`.

Adding the missing assertion reproduces the failure:

```text
Segment 0 ends at 23010ms, beyond the 23000ms of audio supplied
```

Please preserve the real post-resampling endpoint while retaining padded VAD inference, then assert:

```rust
seg.end_timestamp_ms <= audio_duration_ms
```

Also, please adjust the two comments implying 500 ms bounds continuous-audio delivery; that remains #756.

The previous live-redemption blocker is resolved. These are the remaining focused corrections.

safvanatzack:
@Hus-Mek Thank you again for the careful VAD investigation, rebase, focused regression coverage, and the live/batch redemption-policy split.

To keep the remaining review item moving, we’ll take over the final follow-up on `adapt/pr-679` and preserve your #679 work.

Your contribution and the existing review context will remain credited. Thank you for the strong work here.

PR #681, opened 2026-07-31
Fix: stop dropping transcripts below confidence threshold (#675)

### LINKED ISSUE #675: [BUG] Short Whisper transcripts are dropped by text-length confidence heuristic

ironmark72:
Removes the confidence-threshold gate that silently discarded non-empty transcripts (e.g. short acknowledgements like "Yes"/"ok") when confidence was below 0.3. Emission now only depends on the transcript text being non-empty, via a new should_emit_transcript helper. Adds unit tests for the helper.

## Description
Previously, transcripts with a confidence score below 0.3 were silently dropped, even when they contained valid short responses such as "Yes" or "ok". This change removes the confidence-based gate entirely and introduces a should_emit_transcript helper that decides emission based solely on whether the transcript text is non-empty (after trimming). Unit tests were added to cover both the "keep short acknowledgements" and "drop empty/whitespace-only" cases.

## Related Issue
Fixes #675

## Type of Change
- [x] Bug fix

### DIFF
--- a/frontend/src-tauri/src/audio/transcription/worker.rs
+++ b/frontend/src-tauri/src/audio/transcription/worker.rs
@@ -23,6 +23,12 @@ pub fn reset_speech_detected_flag() {
     info!("🔍 SPEECH_DETECTED_EMITTED reset to: {}", SPEECH_DETECTED_EMITTED.load(Ordering::SeqCst));
 }
 
+/// Returns true if the transcript text is non-trivial and should be emitted.
+/// Filters empty/whitespace-only text; no confidence gating is applied.
+fn should_emit_transcript(text: &str) -> bool {
+    !text.trim().is_empty()
+}
+
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct TranscriptUpdate {
     pub text: String,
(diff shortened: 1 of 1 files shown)


safvanatzack:
Thanks for the contribution @ironmark72  this correctly removes the length-derived confidence gate that was dropping valid short Whisper transcripts.

### What is the defect

The gate treats output length as recognition confidence:

```text
non-empty "Yes"                → 0.13 → rejected
non-empty "Thank you for joining" → ≥0.3 → accepted
```

That is a length filter, not a quality filter. It discards valid brief replies and applies unevenly across languages because the measurement is UTF-8 bytes rather than characters or model probability.

The issue’s follow-up comment identifies the same product conflict for short Arabic acknowledgements: text alone cannot distinguish a hallucinated short phrase from a spoken short phrase.

The remaining Stop-time recording-persistence gap is an existing lifecycle issue, not introduced by this PR, and should be tracked separately.

**Approved from my side.**

cc. @sujithatzackriya

Issue #662, opened 2026-07-21
[BUG] Download failed for decoder_joint-model.int8.onnx


Noobzik:
## Bug Description
The application meetily is forcing to download two artifacts, the transcription and the summary engine. This is implied as mandatory.
However I got the following error :
- Download failed for decoder_joint-model.int8.onnx with status: 403 Forbidden

## Current Behavior
The downloading of the transcription engine fails at the very end of the progression with the following error message 
```
Download failed for decoder_joint-model.int8.onnx with status: 403 Forbidden
``` 
The summary engine is downloaded successfully

## Expected Behavior
Completed download

## Steps to Reproduce
1. Run the setup
2. Wait for the download to finish
3. Notice the error

## Environment
- OS: Windows 11 25H2 26200.8655 Enterprise
- Browser: Non relevant
- Version: 0.4.0
- Node Version: Non relevant
- npm/pnpm Version: Non relevant

## Screenshots/Videos

<img width="1101" height="733" alt="Image" src="https://github.com/user-attachments/assets/e376086b-70ce-4c30-91ab-018fc6006499" />

## Error Messages
```
Download failed for decoder_joint-model.int8.onnx with status: 403 Forbidden
```

## Additional Context
The setup the run behind a proxy with SSL inspection that cannot be disable. The used proxy is ZScaler

## Possible Solution
[If you have suggestions on how to fix the issue, please describe them here]

## Checklist
- [x] I have searched for similar issues
- [x] I have provided all required information
- [x] I have included screenshots/videos if applicable
- [x] I have included error messages if applicable

Noobzik:
I just saw the following that is related to this issue
https://github.com/Zackriya-Solutions/meetily/issues/529

After troobleshooting, downloading manually works, however the **setup should not redownload the file that are already downloaded** which in this case does the opposite.

safvanatzack:
Hi @Noobzik 
Will address the issue in upcoming releases, Glad that the workaround helped
Thank you

safvanatzack:
Thanks for identifying the retry behavior after the failed decoder download.

This report is a duplicate of [#529](https://github.com/Zackriya-Solutions/meetily/issues/529) for the underlying first-run Parakeet model download failure, including the `decoder_joint-model.int8.onnx` `403` path.

[...]

PR #682, opened 2026-07-31
fix(parakeet): preserve completed files across download retries

VictorMaxWang:
## Description

Preserve completed and resumable Parakeet model artifacts when another artifact is missing, a request fails, or the user cancels a download.

The previous retry preflight validated the whole model directory and deleted every file when any one artifact was missing or undersized. In the reported flow, the large encoder could finish before the decoder returned HTTP 403, but the next retry deleted the encoder and started over near 0%.

This change:

- removes directory-wide cleanup from the retry path and only reads existing artifact sizes for progress accounting;
- leaves each artifact to the existing per-file skip/Range-resume logic;
- makes active-download registration atomic;
- centralizes active-state cleanup across every normal error/return path;
- preserves partial artifacts on cancellation;
- stores cancellation requests per model so parallel downloads cannot overwrite one another; and
- prevents the retry command from forcibly removing an active worker's registration.

## Related Issue

Partially addresses #662.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other

## Testing

- [ ] All tests pass

`cargo test --offline --locked -q -p meetily --lib parakeet_engine::tests:: -- --nocapture`

- 6 passed, 0 failed, 189 filtered out.
- Covers missing-artifact sibling preservation, cleanup after early failure, rejection of concurrent retry, per-model cancellation isolation, preservation of a cancellation request across retry, and retention of resumable files on cancellation.
- `git diff --check` passed.
- Compiler warnings emitted by the targeted test command are pre-existing in unrelated audio/build modules.

## Documentation

- [ ] Documentation updated
- [x] No documentation needed

## Checklist

- [x] Self-reviewed the code
- [x] No merge conflicts

## Screenshots

Not applicable.

## Additional Notes

This PR does not bypass or claim to solve the decoder HTTP 403 itself. That response can originate from the hosting path, proxy, or TLS inspection and needs separate diagnosis. The change addresses the destructive retry behavior: a decoder failure no longer forces the already downloaded encoder and other siblings to be downloaded again.

### DIFF
--- a/frontend/src-tauri/src/parakeet_engine/commands.rs
+++ b/frontend/src-tauri/src/parakeet_engine/commands.rs
@@ -1,8 +1,7 @@
-use crate::parakeet_engine::{ModelInfo, ModelStatus, ParakeetEngine, DownloadProgress};
+use crate::parakeet_engine::{DownloadProgress, ModelInfo, ParakeetEngine};
 use std::path::PathBuf;
-use std::sync::Mutex;
-use std::sync::Arc;
-use tauri::{command, Emitter, AppHandle, Manager, Runtime};
+use std::sync::{Arc, Mutex};
+use tauri::{command, AppHandle, Emitter, Manager, Runtime};
 
 // Global parakeet engine
 pub static PARAKEET_ENGINE: Mutex<Option<Arc<ParakeetEngine>>> = Mutex::new(None);
@@ -508,30 +507,9 @@ pub async fn parakeet_retry_download<R: Runtime>(
         guard.as_ref().cloned()
     };
 
-    if let Some(engine) = engine {
-        // DEFENSIVE: Ensure clean state before retry
-        // This handles any edge cases where error handler didn't complete
-        {
-            let mut active = engine.active_downloads.write().await;
-            if active.contains(&model_name) {
-                log::warn!("Retry: Model {} was still in active downloads, removing", model_name);
-                active.remove(&model_name);
-            }
-        }
-
-        // DEFENSIVE: Force model status to Missing to allow fresh download
-        {
-            let mut models = engine.available_models.write().await;
-            if let Some(model) = models.get_mut(&model_name) {
-                log::info!("Retry: Resetting model {} status from {:?} to Missing", model_name, model.status);
-                model.status = ModelStatus::Missing;
-            }
-        }
-
-        // Rediscover models to refresh state based on disk files
-        let _ = engine.discover_models().await;
-
-        // Call regular download (emits events)
+    if engine.is_some() {
+        // 重试必须复用常规下载入口的原子 active 注册；旧 worker 退出前不能强删标记，
+        // 否则取消后立即重试会让两个任务并发写入同一个 artifact。
         parakeet_download_model(app_handle, model_name).await
     } else {
         Err("Parakeet engine not initialized".to_string())
--- a/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs
+++ b/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs
@@ -2,7 +2,7 @@ use crate::parakeet_engine::model::ParakeetModel;
 use anyhow::{anyhow, Result};
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
-use std::path::PathBuf;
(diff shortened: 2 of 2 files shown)


safvanatzack:
Thanks for the contribution @VictorMaxWang ; PR #682 fixes the important destructive-retry behavior behind #662: a later failed artifact request no longer removes sibling artifacts that have already downloaded successfully. The atomic active-download registration is also the right protection against concurrent writers.

Before merging, a few recovery-path issues still need attention:

1. **Cancelled near-complete files can be treated as complete without validation**  
   A user can cancel near the end of the encoder download, retry, and receive a successful “model available” state even if the retained encoder is truncated. The retry skips a file once it reaches 99% of an *approximate* size, before making a validation request ([skip logic](https://github.com/Zackriya-Solutions/meetily/blob/3fb4ad0d359de1b75c8d2b84699a2919750e2634/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs#L697-L707)); the download then completes and marks the model available ([completion transition](https://github.com/Zackriya-Solutions/meetily/blob/3fb4ad0d359de1b75c8d2b84699a2919750e2634/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs#L959-L970)).  
   **Required:** only skip after exact completion validation from a trusted manifest or server response. Add a Cancel → Retry regression test with a file above 99% but below the true length.

2. **Retry is exposed before cancellation has actually finished**  
   Cancel immediately changes the UI to `Missing`, which exposes Download/Retry ([UI cancellation path](https://github.com/Zackriya-Solutions/meetily/blob/3fb4ad0d359de1b75c8d2b84699a2919750e2634/frontend/src/components/ParakeetModelManager.tsx#L210-L229)). However, the backend correctly retains worker ownership until it exits; a stalled stream can take up to 30 seconds to time out ([worker timeout](https://github.com/Zackriya-Solutions/meetily/blob/3fb4ad0d359de1b75c8d2b84699a2919750e2634/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs#L792-L806)). An immediate retry is therefore rejected as already in progress.  
   **Required:** represent a `cancelling` state, or await worker release before enabling retry. Do not release backend ownership early.

3. **Progress can be overstated when a server ignores `Range`**  
   Retained partial bytes are included in aggregate progress before resume negotiation ([initial accounting](https://github.com/Zackriya-Solutions/meetily/blob/3fb4ad0d359de1b75c8d2b84699a2919750e2634/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs#L665-L669)). If the host or proxy responds with a normal successful response rather than `206`, the file is restarted but the discarded bytes remain counted ([fresh-download fallback](https://github.com/Zackriya-Solutions/meetily/blob/3fb4ad0d359de1b75c8d2b84699a2919750e2634/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs#L724-L734)).  
   **Required:** subtract that artifact’s retained contribution whenever it restarts from zero, with a deterministic no-Range test.

4. **The new preservation test does not cover the reported recovery sequence**  
   The test validates byte discovery/preservation, but does not exercise: completed sibling → later artifact 403 → retry → Range resume.  
   **Required:** add an offline local HTTP fixture that returns a 403 for the later artifact and verifies sibling retention plus retry behavior.

Also one gentle note is noticed some comments in Chinese, it would be much better to have it in English

This PR is a meaningful mitigation for [#662](https://github.com/Zackriya-Solutions/meetily/issues/662)

Thank you

safvanatzack:
Thank you for the contribution, @VictorMaxWang.

PR #682 meaningfully addresses the destructive retry behavior behind #662: a failed artifact download no longer removes sibling artifacts that have already completed, and the active-download ownership changes prevent concurrent writers during retry.

We will handle the remaining recovery-path work in a follow-up after this branch is merged and retargeted to `adapt/pr-682-followup`:

- validate retained artifacts before treating them as complete;
- represent cancellation completion before enabling retry;
- correct progress when a server ignores `Range`; and
- add an offline `403 → retry` regression test covering the reported flow.

The underlying `403` / proxy-hosting failure remains separate from this mitigation and will continue to be tracked independently.

Issue #587, opened 2026-07-08
[BUG] build-gpu.sh passes --features none to llama-helper, breaking CPU-only Linux builds


Jacob-Lasky:
## Bug Description

On a CPU-only machine (no CUDA/ROCm/Vulkan SDK installed), `./build-gpu.sh` aborts early during the llama-helper step with:

```
error: the package 'llama-helper' does not contain this feature: none
```

## Root Cause

When GPU auto-detection finds no accelerator, `TAURI_GPU_FEATURE` is set to the literal string `none`. In `frontend/build-gpu.sh` the llama-helper feature block guards only on the variable being non-empty:

```bash
if [ -n "$TAURI_GPU_FEATURE" ]; then
    LLAMA_FEATURE="$TAURI_GPU_FEATURE"          # = "none"
    ...
    HELPER_FEATURES="--features $LLAMA_FEATURE"  # --features none
fi
(cd "$HELPER_DIR" && cargo build --release $HELPER_FEATURES)
```

Since `none` is a non-empty string it passes the guard, so cargo runs `cargo build --release --features none`, and `none` is not a real feature on the `llama-helper` crate.

`frontend/dev-gpu.sh` already handles this correctly with an extra guard: `if [ -n "$TAURI_GPU_FEATURE" ] && [ "$TAURI_GPU_FEATURE" != "none" ]; then`. The main app build is also unaffected because `scripts/tauri-auto.js` guards `feature && feature !== 'none'`. Only the llama-helper block in `build-gpu.sh` is missing the `none` check.

## Fix

Mirror the guard from `dev-gpu.sh`:

```bash
if [ -n "$TAURI_GPU_FEATURE" ] && [ "$TAURI_GPU_FEATURE" != "none" ]; then
```

## Environment

- CachyOS (Arch Linux), CPU-only build (`TAURI_GPU_FEATURE=none`)
- Release build via `./build-gpu.sh`

## Refs

- Fix location: `frontend/build-gpu.sh` (llama-helper feature block, ~L111)
- Correct guard already present in `frontend/dev-gpu.sh` (~L103)
- Main-app guard: `frontend/scripts/tauri-auto.js` (`feature !== 'none'`)
- Surfaced while building PR #586

Issue #305, opened 2026-01-15
[BUG] Cannot build on linux with amd gpu


mtb-xt:
## Bug Description
I cannot get the app to build under linux

## Current Behavior

```
❯ ./build-gpu.sh
🚀 Meetily GPU-Accelerated Build Script


📦 Building Meetily...

🔍 Detecting GPU features...
🔴 AMD GPU detected with ROCm - using HIPBlas acceleration
✅ Detected GPU feature: hipblas

🦙 Building llama-helper sidecar (release)...
   Building in /home/marat.bakeev/software/meeting-minutes/frontend/../llama-helper with features: --features hipblas
warning: patch for the non root package will be ignored, specify patch at the workspace root:
package:   /home/marat.bakeev/software/meeting-minutes/frontend/src-tauri/Cargo.toml
workspace: /home/marat.bakeev/software/meeting-minutes/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/marat.bakeev/software/meeting-minutes/llama-helper/Cargo.toml
workspace: /home/marat.bakeev/software/meeting-minutes/Cargo.toml
error: the package 'llama-helper' does not contain this feature: hipblas
help: package with the missing feature: meetily
```

## Expected Behavior
Software build works

## Steps to Reproduce
1. Download source (tried both main and v0.2.0
2. cd frontend
3. ./build-gpu.sh

## Environment
- OS: Ubuntu Linux 24.04 LTS
- Browser: Firefox, Safari]
- Version: 0.2.0
- Node Version: v20.19.3
- npm/pnpm Version: npm 10.8.2

## Checklist
- [X] I have searched for similar issues
- [X] I have provided all required information
- [X] I have included screenshots/videos if applicable
- [X] I have included error messages if applicable

i-love-coffee-i-love-tea:
I'm on Ubuntu 26.04 with ROCm 7.1.1 from Ubuntu package sources and I get the same error. 

Shouldn't the feature section in `llama-helper/Cargo.toml` contain something like `llama-cpp2/rocm` for it to support rocm? I noticed it only contains `metal`, `cuda` and `vulcan` features.

khassad:
Same here, building for Vulkan workaround this but fails later.

PR #687, opened 2026-08-03
fix(build): locate CUDA static libs on distro-packaged toolkits

passabilities:
## Description

`./build-gpu.sh` fails on Linux machines that use a distro-packaged CUDA toolkit:

```
error: could not find native static library `cudart_static`, perhaps an -L flag is missing?
```

The build gets as far as `Detected GPU feature: cuda` and then dies while compiling `llama-cpp-sys-2`.

### Root cause

`llama-cpp-sys-2/build.rs` emits `cargo:rustc-link-lib=static=cudart_static` (plus `cublas_static`, `culibos`) and delegates `-L` path discovery to `find_cuda_helper`. That crate's `find_cuda_lib_dirs()` only probes `$CUDA_LIBRARY_PATH`, `/opt/cuda/lib64` and `/usr/local/cuda*/lib64`.

Distro packages such as Ubuntu's `nvidia-cuda-toolkit` install into `/usr`, putting `nvcc` at `/usr/bin/nvcc` and the libraries in `/usr/lib/x86_64-linux-gnu`. None of the probed paths exist, so rustc receives the `-l static=` flags with zero `-L` paths and the link fails.

This reproduces in isolation: a two-line crate with `#[link(name = "cudart_static", kind = "static")]` fails identically without `-L`, and succeeds with `-L native=/usr/lib/x86_64-linux-gnu`.

### Fix

Both `frontend/build-gpu.sh` and `frontend/dev-gpu.sh` now locate the CUDA static libraries before building and export a matching `-L native=` into `RUSTFLAGS`.

The block only runs when `TAURI_GPU_FEATURE = cuda` on `linux-gnu`, and probes in this order:

1. `$CUDA_PATH/lib64`
2. `$CUDA_HOME/lib64`
3. nvcc-relative `../lib64`
4. `/usr/local/cuda/lib64`
5. `/opt/cuda/lib64`
6. `/usr/lib/$(uname -m)-linux-gnu`
7. `/usr/lib64`

The first directory containing `libcudart_static.a` wins. Conventional locations are checked before the multiarch path, so existing `/usr/local/cuda` and `/opt/cuda` installs keep their current behaviour. `RUSTFLAGS` is appended to rather than overwritten.

If no candidate matches, the script prints a warning pointing at `CUDA_PATH` and continues rather than aborting, since the link step may still succeed on setups this heuristic does not cover.

## Related Issue

No existing issue covers this specific link failure. Related Linux build reports: #305, #587.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [x] All tests pass

On Ubuntu 24.04 with `nvidia-cuda-toolkit` (nvcc 12.0 at `/usr/bin/nvcc`, no `/usr/local/cuda*`):

- Before: the build fails at `llama-cpp-sys-2` with the error above.
- After: the full build succeeds and produces both `.deb` and `.AppImage` bundles.
- `ldd` on the resulting binary confirms CUDA is genuinely linked: `libcudart.so.12`, `libcublas.so.12`, `libcublasLt.so.12`, `libcuda.so.1`.

Shell-script only; no effect on macOS or Windows, or on Linux builds that are not using the `cuda` feature.

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

N/A — build script change only.

## Additional Notes

The probe order deliberately prefers the conventional CUDA install locations so that machines with a vendor toolkit under `/usr/local/cuda` are unaffected. The multiarch path is only reached when nothing more specific matched.

### DIFF
--- a/frontend/build-gpu.sh
+++ b/frontend/build-gpu.sh
@@ -88,6 +88,41 @@ else
     echo -e "${YELLOW}⚠️ No specific GPU feature detected or forced${NC}"
 fi
 
+# Locate CUDA static libraries.
+# llama-cpp-sys-2 links cudart_static/cublas_static/culibos and relies on
+# find_cuda_helper, which only probes $CUDA_LIBRARY_PATH/lib64, /opt/cuda/lib64
+# and /usr/local/cuda*/lib64. Distro packages (e.g. Ubuntu's nvidia-cuda-toolkit)
+# install into /usr/lib/<triple> instead, so rustc gets no -L path and fails with
+# "could not find native static library `cudart_static`".
+if [ "$TAURI_GPU_FEATURE" = "cuda" ] && [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    CUDA_LIB_DIR=""
+    NVCC_LIB_DIR=""
+    if command_exists nvcc; then
+        NVCC_LIB_DIR="$(dirname "$(command -v nvcc)")/../lib64"
+    fi
+    for candidate in \
+        "${CUDA_PATH:+$CUDA_PATH/lib64}" \
+        "${CUDA_HOME:+$CUDA_HOME/lib64}" \
+        "$NVCC_LIB_DIR" \
+        /usr/local/cuda/lib64 \
+        /opt/cuda/lib64 \
+        "/usr/lib/$(uname -m)-linux-gnu" \
+        /usr/lib64; do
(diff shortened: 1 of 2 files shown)


Dixilea:
The changes suggested to `build-gpu.sh` indeed resolved the issue for me on Ubuntu.

PR #694, opened 2026-08-04
fix(summary): handle Claude thinking blocks in Anthropic response parsing

davertor:
### Problem

Selecting `claude-sonnet-5` (or `claude-opus-5`) as the Claude model makes summary generation fail with:

```
Failed to parse LLM response: missing field `text`
```

The models are offered in the picker — it is populated live from `/v1/models` (`src/anthropic/anthropic.rs:74`) and `is_chat_model` accepts anything starting with `claude-` — so they are selectable but unusable.

### Root cause

These models enable adaptive thinking by default when the request omits a `thinking` parameter, so the response `content` array leads with a `thinking` block that carries no `text` field. `ClaudeChatContent` declared `text` as a required `String`, which fails deserialization of the entire response. Even with lenient parsing, `content.get(0)` would have returned the thinking block rather than the summary.

Older models (`claude-sonnet-4-5` and earlier) do not think unless asked, so `content[0]` was always a text block and the original assumption held.

### Fix

- `text` becomes `Option<String>` so non-text blocks deserialize.
- Select the first block that actually carries text, instead of block 0.
- `max_tokens` 2048 → 8192. It is a shared budget covering thinking *and* the answer, so 2048 truncated summaries mid-sentence on these models.

Backwards compatible: models without thinking return a single leading text block, which is still what gets selected.

### Testing

Three unit tests added in `llm_client.rs` covering the thinking-first, text-only, and no-text-block response shapes.

**Transparency:** I could not run `cargo test` or an end-to-end summary — I had no Rust toolchain available while preparing this. The change is small and hand-reviewed, but please let CI and a reviewer confirm rather than taking my word for it. Happy to push fixups.

### Related, not fixed here

The Python backend (`backend/app/transcript_processor.py`) hits the same wall from a different angle: `pydantic-ai==0.2.15` asserts that every non-text block is a `ToolUseBlock` (`pydantic_ai/models/anthropic.py:257`), so a thinking block raises `AssertionError`. That needs a dependency bump and belongs in its own PR.

### DIFF
--- a/frontend/src-tauri/src/summary/llm_client.rs
+++ b/frontend/src-tauri/src/summary/llm_client.rs
@@ -60,7 +60,17 @@ pub struct ClaudeChatResponse {
 
 #[derive(Deserialize, Debug)]
 pub struct ClaudeChatContent {
-    pub text: String,
+    // Only `text` blocks carry this field. Models that enable thinking by
+    // default (Sonnet 5, Opus 5) also return `thinking` blocks, which don't.
+    pub text: Option<String>,
+}
+
+impl ClaudeChatResponse {
+    /// First block that carries text. With thinking enabled the leading block
+    /// is a `thinking` block, so `content[0]` is not necessarily the answer.
+    fn first_text(&self) -> Option<&str> {
+        self.content.iter().find_map(|block| block.text.as_deref())
+    }
 }
 
 /// LLM Provider enumeration for multi-provider support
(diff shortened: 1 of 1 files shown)


safvanatzack:
@sujithatzackriya, reviewed and approved from my side.

PR #694 safely handles thinking-first Claude response blocks while retaining text-only compatibility. No source-backed regressions were found in the review.

Please note:
> max_tokens 2048 → 8192. It is a shared budget covering thinking and the answer, so 2048 truncated summaries mid-sentence on these models.

Good to merge.

PR #720, opened 2026-08-20
fix(audio): resolve real PulseAudio/PipeWire monitor sources for Linux system audio capture

### LINKED ISSUE #701: Linux: System Audio device is never resolved at recording start — system audio silently missing
## Description

On Linux, selecting any "System Audio" device in Settings and starting a recording never actually captures system audio, and no error is shown anywhere in the UI. The recording proceeds normally with microphone-only audio, silently missing the system/meeting audio track.

## Root cause

- `frontend/src-tauri/src/audio/devices/platform/linux.rs` (`configure_linux_audio`) enumerates PulseAudio/PipeWire monitor sources visible to cpal's ALSA host and decorates the display name for the picker:
  ```rust
  devices.push(AudioDevice::new(
      format!("{} (System Audio)", name),
      DeviceType::Output
  ));
  ```
  e.g. a raw device named `my_monitor` becomes `"my_monitor (System Audio)"`. This decorated string is what gets shown in Settings and persisted to `recording_preferences.json`.

- At recording start, `frontend/src-tauri/src/audio/devices/configuration.rs::get_device_and_config()` resolves the actual cpal device by exact string comparison against the raw device name:
  ```rust
  #[cfg(target_os = "linux")]
  {
      if let Ok(pulse_host) = cpal::host_from_id(cpal::HostId::Alsa) {
          for device in pulse_host.input_devices()? {
              if let Ok(name) = device.name() {
                  if name == audio_device.name { ... }
              }
          }
      }
  }
  ```
  `audio_device.name` still carries the `" (System Audio)"` suffix — only `"(input)"`/`"(output)"` suffixes are stripped elsewhere, in `AudioDevice::from_name`. So `name == audio_device.name` (`"my_monitor" == "my_monitor (System Audio)"`) never matches, and the function returns `Err("Device not found: ...")`.

- That error is swallowed in `frontend/src-tauri/src/audio/stream.rs::AudioStreamManager::start_streams()`:
  ```rust
  Err(e) => {
      warn!("⚠️ Failed to create system audio stream: {}", e);
      // Don't fail if only system audio fails
  }
  ```
  It's only logged via `warn!()` (invisible without `RUST_LOG` set) and the recording proceeds anyway — nothing surfaces to the user.

Net effect: on Linux, **no** "System Audio" device can ever be successfully opened for recording, regardless of which entry is picked in Settings — this affects 100% of Linux users trying to record system/meeting audio, not just a specific device.

## Steps to reproduce

1. On a Linux desktop with PipeWire or PulseAudio, open Settings → Default Audio Devices.
2. Pick any entry under "System Audio" (populated by `configure_linux_audio`).
3. Start a recording while system audio is playing (e.g. a video call).
4. Stop the recording and inspect the file: microphone audio is present, system audio is not. No warning or error was shown during the process.

## Suggested fix

Strip the `" (System Audio)"` display suffix before comparing against the raw cpal device name, mirroring the existing `"(input)"`/`"(output)"` stripping already done in `AudioDevice::from_name`:

```rust
#[cfg(target_os = "linux")]
{
    let raw_name = audio_device
        .name
        .trim_end_matches(" (System Audio)")
        .to_string();

    if let Ok(pulse_host) = cpal::host_from_id(cpal::HostId::Alsa) {
        for device in pulse_host.input_devices()? {
            if let Ok(name) = device.name() {
                if name == raw_name {
                    let default_config = device
                        .default_input_config()
                        .map_err(|e| anyhow!("Failed to get default input config: {}", e))?;
                    return Ok((device, default_config));
                }
            }
        }
    }
}
```

I've applied and locally built this exact change and confirmed it resolves the issue end-to-end (recorded file now contains system audio). Happy to open a PR with it if that's useful.

## Related usability gap (secondary, worth a mention)

Separately: cpal's ALSA host on a stock PipeWire/PulseAudio desktop generally does **not** surface monitor sources as ALSA devices out of the box (`snd_device_name_hint` / `arecord -L` won't list them), so `configure_linux_audio`'s `name.contains("monitor")` filter may simply find nothing on a fresh install — the "System Audio" list can be empty with no guidance for the user. Getting a monitor source to appear currently requires manually registering it as a named ALSA PCM via `~/.asoundrc` (`type pulse`, with an explicit `hint { show on }` block, and a device name that itself contains "monitor" since the filter matches on the cpal-reported name, not the description). Might be worth documenting this Linux setup step, or better, capturing system audio on Linux via PipeWire/libpulse directly instead of going through the ALSA hint enumeration.

## Environment

- Meetily 0.4.0, self-built from `main` HEAD
- Arch Linux (EndeavourOS), PipeWire 1.6.8, KDE Plasma / Wayland

AnsCodeLab:
## Description

On Linux, cpal has no ALSA loopback backend, so system audio was never actually captured: recording silently fell back to microphone-only, with no error surfaced anywhere.

- `configure_linux_audio()` tried to find PulseAudio/PipeWire monitor sources via `cpal::host_from_id(HostId::Alsa)` -- the same host cpal already defaults to on Linux (there's no separate Pulse `HostId`). ALSA's `snd_device_name_hint` enumeration never exposes monitor sources (confirmed live: `arecord -L` lists none even when `pactl list sources` shows several), so no usable system-audio device was ever offered in the device picker.
- `default_output_device()` (the path used whenever no device is explicitly selected -- the common case) returned a playback sink, not a capture source. `get_device_and_config()` then tried to open it as a cpal input stream and failed. The failure was only logged as a `warn!`, so recording silently continued mic-only.
- The generic "any leftover cpal device -> Output" fallback in `discovery.rs` also made the Settings permission-check's `hasSystemAudio` flag a false positive on Linux, so users got no warning either.

## Related Issue

Fixes #701

## Fix

Adds a Linux-only `audio::capture::pulse` module that shells out to `pactl`/`parec` (from `pulseaudio-utils`, also provided by PipeWire's `pipewire-pulse` compatibility layer -- present on essentially every modern Linux desktop) to enumerate real monitor sources, resolve the default sink's monitor, and stream captured system audio into the existing `AudioCapture` pipeline. This mirrors the existing macOS Core Audio backend split already present in `stream.rs`. Also drops the generic Linux fallback in `discovery.rs` that was producing unusable pseudo-devices.

Two issues caught in an independent code review before merge and fixed:
- `pactl`'s field labels (`Name:`, `Description:`, `Monitor of Sink:`) are gettext-translated per desktop locale; without forcing `LC_ALL=C` the parser silently found nothing on any non-English desktop, reproducing this exact bug under a different trigger. Fixed, with a regression test that sets the test process's own `LANG`/`LC_ALL` to German and confirms sources are still found (verified it actually fails without the fix).
- The `pactl`-based monitor resolution ran synchronously on the tokio runtime instead of via `spawn_blocking`, inconsistent with the pattern already used for the `parec` reader loop in the same function. Fixed.

## Type of Change
- [x] Bug fix

## Testing
- [x] Unit tests added/updated (`frontend/src-tauri/src/audio/capture/pulse.rs`: pactl output parser test + locale-override regression test)
- [x] All tests pass

Verified live end-to-end on a GNOME/PipeWire Linux desktop, not just compiled:
- Enumerated real monitor sources via the new module; resolved the default sink's monitor; round-tripped display name -> technical source name.
- Played a test tone through `paplay` while capturing from the resolved monitor: captured samples showed clear non-silence (max amplitude 0.41, 87% non-zero samples), proving real loopback capture, not silence or mic bleed.
- Built and installed a full release `.deb`, launched the packaged app, started a real recording with no explicit device selection (the common path), and confirmed via `/proc/<pid>/io` on the app's own spawned `parec` child that it sustained ~384 KB/s -- the exact expected rate for 48kHz/2ch/f32 -- for the duration of the recording.
- Confirmed the same install correctly resolves and captures via the packaged binary's own logs (`Using PulseAudio/PipeWire monitor capture for system audio`, `Recording manager started successfully with 2 active streams`).

## Documentation
- [x] No documentation needed (internal implementation fix, no user-facing config/API change)

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

### DIFF
--- a/frontend/src-tauri/src/audio/capture/mod.rs
+++ b/frontend/src-tauri/src/audio/capture/mod.rs
@@ -7,6 +7,9 @@ pub mod backend_config;
 #[cfg(target_os = "macos")]
 pub mod core_audio;
 
+#[cfg(target_os = "linux")]
+pub mod pulse;
+
 // Re-export capture functionality
 pub use system::{
     SystemAudioCapture, SystemAudioStream,
@@ -16,6 +19,9 @@ pub use system::{
 
 #[cfg(target_os = "macos")]
 pub use core_audio::{CoreAudioCapture, CoreAudioStream};
+
+#[cfg(target_os = "linux")]
+pub use pulse::PulseMonitorSource;
 
 // Re-export backend configuration
 pub use backend_config::{
--- a/frontend/src-tauri/src/audio/capture/pulse.rs
+++ b/frontend/src-tauri/src/audio/capture/pulse.rs
@@ -1,0 +1,226 @@
+// PulseAudio/PipeWire system-audio (loopback) capture for Linux.
+//
+// cpal's ALSA backend cannot enumerate or open PulseAudio/PipeWire "monitor"
+// sources: ALSA's `snd_device_name_hint` only reports genuine ALSA PCM
+// devices, and PulseAudio/PipeWire never register per-sink monitors as ALSA
+// hardware devices (confirmed: `arecord -L` never lists them). WASAPI
+// (Windows) and Core Audio (macOS) both support native loopback capture
+// through cpal/cidre; ALSA has no equivalent, so system audio devices
+// selected via cpal on Linux (mistakenly labelled as monitors, or as the
+// raw default *playback* device) always fail to open as an input stream and
+// recording silently falls back to microphone-only.
+//
+// Monitor sources are only visible through PulseAudio's own client
+// protocol. Rather than linking libpulse directly (the `-devel`/`-dev`
+// headers are frequently missing on end-user systems), this module shells
+// out to `pactl`/`parec`, which ship in `pulseaudio-utils` and are also
+// provided by PipeWire's `pipewire-pulse` compatibility layer -- i.e.
+// present on essentially every modern Linux desktop.
+
+use anyhow::{anyhow, Result};
+use std::process::{Child, Command, Stdio};
+
+/// A PulseAudio/PipeWire source that monitors a sink's output, i.e. usable
+/// for system-audio/loopback capture.
+#[derive(Debug, Clone)]
+pub struct PulseMonitorSource {
+    /// Technical PulseAudio source name, e.g. "alsa_output.pci-....monitor"
+    pub name: String,
+    /// Human-readable description, e.g. "Monitor of Built-in Audio Speaker"
+    pub description: String,
+    /// Name of the sink this source monitors.
+    pub monitor_of_sink: String,
+}
+
+/// List every PulseAudio/PipeWire source that monitors a sink (i.e. every
+/// source usable for system-audio capture). Requires `pactl`.
+pub fn list_monitor_sources() -> Result<Vec<PulseMonitorSource>> {
+    let output = Command::new("pactl")
+        .args(["list", "sources"])
+        .output()
+        .map_err(|e| {
+            anyhow!(
+                "Failed to run `pactl` ({e}); install `pulseaudio-utils` (or `pipewire-pulse`) for system audio capture"
+            )
+        })?;
+
+    if !output.status.success() {
+        return Err(anyhow!(
+            "`pactl list sources` exited with {}: {}",
+            output.status,
+            String::from_utf8_lossy(&output.stderr)
+        ));
+    }
+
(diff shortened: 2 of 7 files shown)


AnsCodeLab:
Closing per request.

PR #721, opened 2026-08-20
feat(widget): floating always-on-top recording widget

### LINKED ISSUE #718: [FEATURE] Floating always-on-top recording widget (mini window)
## Description

Add an optional floating, always-on-top "recording widget" window: a small, borderless, draggable pill showing recording status (elapsed time, a live mic/system level indicator, pause/stop controls) that stays visible above whatever app the user is actually working in during a meeting, instead of requiring them to switch back to the full Meetily window to see/control recording state.

## Motivation

Meetily is used *during* a meeting while the user is in a video-call app, browser, notes app, etc. Today the only always-visible recording indicator is the system tray icon, which requires opening a menu to see status or stop recording. A small persistent on-screen widget (the pattern used by e.g. screen recorders and other meeting-note tools) makes recording state and quick controls visible without alt-tabbing away from the meeting.

## Proposed approach

- New Tauri window (`tauri.conf.json` currently defines a single `main` window — see `frontend/src-tauri/tauri.conf.json`), created/shown/hidden via a new Tauri command, not part of `app.windows` at startup:
  - `decorations: false`, `transparent: true`, `alwaysOnTop: true`, `skipTaskbar: true`, `resizable: false`, small fixed size (e.g. ~260x64).
  - Custom drag region (`data-tauri-drag-region`) so it can be repositioned; position persisted (reuse the existing `recording_preferences`/settings storage pattern already used for mic/system device prefs).
- Reuses existing recording state/events end-to-end — no new Rust recording logic. It listens to the same `transcript-update`/recording-state events the main window already listens to (`frontend/src/contexts` + existing Tauri event listeners), and calls the existing `pause_recording`/`resume_recording`/`stop_recording` commands from `recordingService.ts`.
- A small new React route/window entry (Tauri supports multiple HTML entry points, or a single-page app keyed off a `window.label`/query param) rendering just the pill UI.
- Toggle: a setting ("show floating widget while recording") plus a tray menu item to show/hide it on demand.
- Platform note: `alwaysOnTop`/`transparent`/`decorations:false` are supported the same way on X11 and Wayland (native GTK window flags) — no OS-specific code needed, unlike system-audio-loopback style work.

## Acceptance criteria

- [ ] Floating widget window can be shown/hidden via a Tauri command and a UI toggle.
- [ ] Shows live elapsed recording time and updates when paused/resumed.
- [ ] Stop button actually stops the active recording (verified against the real recording state, not just closes the window).
- [ ] Window stays on top of other applications and has no title bar/borders.
- [ ] Position is draggable and persists across app restarts.
- [ ] Does not regress the existing main-window recording flow (main window recording controls keep working when the widget is open or closed).

## Out of scope

- Click-through / ignore-mouse-events mode (widget stays fully interactive for now).
- Mobile — this only applies to the Tauri desktop app.

AnsCodeLab:
## Description

Adds an optional floating, always-on-top "recording widget" window: a small, borderless, draggable pill showing recording status (elapsed time, pause/resume/stop) that stays visible above whatever app you're actually working in during a meeting, instead of requiring a switch back to the full Meetily window.

## Related Issue

Fixes #718

## Design notes

- New `widget` window declared statically in `tauri.conf.json` (`visible:false`, `decorations:false`, `transparent:true`, `alwaysOnTop:true`, `skipTaskbar:true`), toggled via a new `toggle_widget_window` command from the tray menu and a Settings checkbox.
- The widget page (`app/(widget)/widget/page.tsx`) needed its own Next.js route-group root layout (`app/(widget)/layout.tsx`) -- the existing single root `layout.tsx` unconditionally wraps every route with the full Sidebar/onboarding/8-provider tree, which would have rendered the entire app UI squeezed into a 260x70 pill. Existing routes were moved into `app/(main)/` to keep their URLs and provider tree unchanged.
- The widget's Stop button uses a new `stop_recording_from_widget` command that mirrors `tray.rs`'s `stop_recording_handler` exactly (computes `save_path`, calls `stop_recording`, emits `recording-stop-complete`) rather than a bare `stop_recording` invoke -- the meeting-finalization listener (`RecordingPostProcessingProvider`) only lives in the main window's React tree, so a naive Stop button would have stopped the recording without ever saving it.
- Position persisted via a new `widget_preferences.json` store (mirrors `recording_preferences.rs`'s pattern), written from a debounced `WindowEvent::Moved` handler in Rust so the widget's own JS needs no filesystem permission.

An independent plan review before implementation caught both of the design points above; an independent code review after implementation caught one more required fix (below).

## Type of Change
- [x] New feature

## Testing
- [x] All tests pass (no new unit-testable logic beyond existing patterns)

Verified, not just compiled:
- `cargo check` clean; `pnpm run build` (`next build`, `output:'export'`) succeeds -- `/widget` compiles and statically exports alongside all existing routes, confirming the `(main)`/`(widget)` route-group split; inspected `out/widget.html` directly and confirmed it contains no Sidebar/onboarding markup.
- Built a full release `.deb`, installed it, launched it against a real X11/XWayland display, and used `wmctrl`/`xdotool` to inspect real OS window state: confirmed the widget stays hidden by default, and that writing `show_widget:true` + a position to the real store file and relaunching shows the widget window at exactly that position (proves the startup-restore path executes end to end against the real store format).
- A code review before merge caught that the `widget` capability was missing `core:window:allow-start-dragging`, which the drag-to-reposition mechanism depends on -- every drag attempt would have been silently rejected at the ACL layer. Fixed. A follow-up manual drag test under an XWayland compatibility shim (this app is native Wayland; `wmctrl`/`xdotool` need X11) was inconclusive -- the window didn't visibly move, but synthetic `xdotool` pointer events may not exercise the same native drag-begin path as a real user gesture, and this same compat shim already showed an unrelated GTK-specific quirk (the widget renders 400px tall instead of the requested 70px on this GTK stack -- noted below, not something fixable purely through Tauri's declarative window config). **A human doing a real click-and-drag check on an actual desktop session is the recommended final confirmation for the drag interaction specifically.**
- Not verified: clicking the actual pause/resume/stop buttons or a real end-to-end "record → widget shows elapsed time → stop → confirm meeting saved" pass, since no interactive click automation was available in this environment beyond `wmctrl`/`xdotool` window-geometry inspection -- the stop-flow correctness argument rests on the code-level match to `tray.rs`'s already-shipped, working stop handler (verified line-by-line) plus `app.emit` reaching all windows (confirmed in source).

## Known limitation
On this Linux/GTK/WebKitGTK stack, the widget window renders at 400px tall instead of the requested 70px on first map, even with `resizable:false` and explicit min/max size hints -- appears to be a wry/webkit2gtk minimum-content-size floor, not fixable purely through declarative Tauri config. A post-show `set_size()` re-assertion is in place but doesn't override it. Width sizing is correct; macOS/Windows aren't expected to hit this same floor (untested, not buildable on this machine).

## Documentation
- [x] No documentation needed (opt-in feature, self-explanatory UI)

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts


AnsCodeLab:
Withdrawing: the floating widget doesn't work well in practice (drag-to-reposition doesn't function correctly, plus a known GTK sizing issue). Deleted the source branch, which is what auto-closed this.

PR #722, opened 2026-08-20
feat(dictation): live dictation mode via AT-SPI (Linux)

### LINKED ISSUE #719: [FEATURE] Live dictation mode: transcribe speech directly to the OS cursor in any app
## Description

Add an optional "live dictation" mode: instead of only recording audio to a file and transcribing it into the meeting's transcript panel, continuously transcribe speech and insert the recognized text directly at the current text-cursor position in whatever application currently has OS focus (e.g. an email client, a chat app, a document editor) — effectively system-wide voice dictation, not tied to Meetily's own window.

## Motivation

Meetily already runs local streaming transcription (`frontend/src-tauri/src/audio/transcription/`, Whisper/Parakeet). The only output today is the in-app transcript panel + saved meeting. Reusing that same local transcription pipeline as a system-wide dictation tool (type-what-I-say into any app) is a natural extension and avoids depending on a separate dictation tool.

## Investigated feasibility

This machine/session is GNOME on **Wayland** (`XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=GNOME`, Mutter compositor). That rules out the naive approach:

- `xdotool` (X11/XTest) can't reliably target or type into native Wayland client windows — Wayland deliberately blocks synthetic input across app boundaries for security. Only works for legacy XWayland-backed apps.
- `wtype` (the Wayland analog of `xdotool`) only works on wlroots-based compositors (Sway, etc.), **not** Mutter (GNOME) or KWin (KDE) — the two most common desktop compositors.
- `ydotool` works everywhere (it's a kernel-level virtual keyboard via `/dev/uinput`), but requires one-time, per-machine root setup: install the package, a udev rule granting the user's group access to `/dev/uinput` (currently `crw-------`, root-only), and running the `ydotoold` daemon as a service. Not something a desktop app can silently self-provision without an explicit, admin-consented setup step.
- The AT-SPI accessibility bus (`org.a11y.Bus`) is already running with no extra permissions needed. AT-SPI's `EditableText` interface lets a client find the currently-focused accessible text field (across GTK/Qt apps, LibreOffice, many Electron apps, etc.) and insert text directly into it — this is the same mechanism real dictation tools (e.g. Talon, nerd-dictation) use on Linux, and it's the only path here that works today with zero extra system configuration. Coverage isn't 100% (some apps under-implement AT-SPI, e.g. certain Electron/browser text fields), so it needs a graceful "couldn't find a focused text field" fallback.

## Proposed approach

- **Primary path — AT-SPI text insertion**: new Rust module under `frontend/src-tauri/src/audio/` (or a sibling module, e.g. `dictation/`) that, when live-dictation mode is on, queries the AT-SPI bus for the currently focused accessible object, and calls its `EditableText::insert_text` (or `Text`/`Value` interfaces as appropriate) with each finalized transcription segment as it's produced by the existing transcription worker — reusing the same VAD-segmented transcription stream already wired for the transcript panel, not a second transcription pipeline.
- **Fallback when no focused editable text object is found**: don't fail silently — surface a clear, visible notification/toast ("no active text field detected — copied to clipboard instead") and put the segment on the clipboard as a safe fallback the user can paste manually.
- **Opt-in "works everywhere" mode**: document `ydotool` as an optional advanced path (a settings-page note with the exact `dnf`/`apt` package + udev rule steps), not something the app attempts to configure itself. Not part of the default/required implementation.
- **Global toggle + safety**: a clearly visible, easy-to-hit global hotkey and/or the floating widget from #718 to start/stop dictation mode, since text is being injected into whatever has focus — accidental mis-typing into the wrong window is the main risk to design against (e.g. show a persistent, unmissable on-screen indicator whenever dictation mode is actively injecting text).
- Scope this to Linux first (matches the investigated environment); Windows/macOS have their own native accessibility APIs (UI Automation / Accessibility API) that would need separate follow-up work, not assumed here.

## Acceptance criteria

- [ ] A "Live Dictation" mode can be started/stopped independently of normal meeting recording.
- [ ] While active, finalized transcription segments are inserted at the OS text cursor in the currently focused app (verified against a real focused text field in a real other application, not just the Meetily window).
- [ ] When no focused editable text field is found, the app does not silently drop the text — it notifies the user and falls back to clipboard.
- [ ] A persistent, hard-to-miss visual indicator is shown whenever dictation mode is actively able to inject text.
- [ ] Documented (README/settings help text) that this requires Linux + AT-SPI (default) with `ydotool` documented as an optional broader-coverage path.

## Out of scope

- Windows/macOS native accessibility integration (separate follow-up).
- Auto-installing/configuring `ydotool`/udev rules on the user's behalf.

AnsCodeLab:
## Description

Adds an optional "Live Dictation" mode: continuously transcribes speech and inserts the recognized text directly at the current text cursor in whatever application currently has OS focus (email client, chat app, editor, etc.), via AT-SPI2 on Linux -- reusing the app's existing local transcription pipeline rather than a separate dictation engine.

## Related Issue

Fixes #719

## Feasibility investigation (Linux-first, matches the issue's scope)

This machine is GNOME on Wayland (Mutter). `xdotool`/`wtype` can't reliably inject into native Wayland windows or don't support Mutter; `ydotool` works everywhere but needs one-time root/udev setup this app should never silently perform. AT-SPI (`org.a11y.Bus`) is already running with zero extra permissions and is the same mechanism real Linux dictation tools (Talon, nerd-dictation) use -- chosen as the only path that works today with no admin setup. `ydotool` is documented as an optional advanced path, not auto-configured.

## Design notes

An independent plan review before implementation required several safety fixes, and an independent code review after implementation required one more -- all implemented:

- **No `is_partial` filtering.** The plan initially assumed `is_partial` meant "wait for a final revision" -- it's actually a chunk-duration flag (`whisper_engine.rs`), and filtering on it would have silently dropped nearly all real dictation output. Every finalized `transcript-update` segment is injected.
- **Password-field protection.** Checks `Role::PasswordText` on the focused object before every injection (the atspi crate has no `State::PasswordText` as originally assumed -- verified against the actual published crate source; AT-SPI2 represents password fields via role, the same mechanism screen readers use). Skips injection and falls back to clipboard+notification if the focused field is a password field.
- **Fresh re-validation, append-only.** Immediately before every `insert_text` call, focus is re-confirmed and the insertion offset is the text's current end, queried fresh -- never a remembered/predicted caret position, to avoid corrupting either the dictated text or concurrent user typing.
- **Decoupled from transcription.** Injection never runs inline in the sole serial transcription worker (`worker.rs`, `NUM_WORKERS=1`, deliberately serial for zero chunk loss) -- a bounded, drop-oldest queue + separate task owns the AT-SPI connection.
- **Clipboard fallback preserves prior content**, restored after 10s -- but only if the clipboard still holds exactly the fallback text at restore time (a code-review fix: the initial version would have clobbered anything the user copied in between).
- **`start()`/`stop()` are now serialized** end-to-end via a lifecycle lock (a code-review fix: concurrent toggles -- tray double-click, hotkey racing a Settings save -- could otherwise both pass a stale `is_active()` check and each spawn an independent AT-SPI connection, orphaning the first one's D-Bus subscription with no way for a later `stop()` to reach it).

A lower-severity finding (queue-overflow drops are logged but not surfaced to the user, unlike the injection-failure path) is left as a known residual gap, noted here rather than silently dropped -- unlikely to trigger under normal dictation pacing given the small number of D-Bus round trips per segment.

## Type of Change
- [x] New feature

## Testing
- [x] Unit tests added/updated (`dictation::queue`: FIFO order, drop-oldest overflow, close/reopen semantics -- 5 tests)
- [x] All tests pass

Verified live end-to-end against real other applications on this machine, not mocked:
- `cargo check`/`cargo test --lib` clean (191 passing; the one pre-existing unrelated failure, a floating-point precision assertion in Bluetooth buffer-timeout code, predates and is untouched by this change).
- Wrote and ran real, non-mocked integration tests (`tests/dictation_atspi_live.rs`) against this machine's actual live GNOME Wayland session: spawned a real `gnome-text-editor`, drove the exact production `connect()`/focus-tracker/`inject_segment()` path to type a marker string, and confirmed it landed by reading the buffer back over an independent AT-SPI call -- **passed**. Spawned a real `zenity --password` dialog, confirmed injection is skipped every time it's focused, and confirmed the field was never typed into -- **passed**. Re-ran 3 consecutive times after the code-review fixes above, 2/2 passing each run (one earlier isolated run had a single flaky failure that didn't reproduce on retry, consistent with transient desktop-focus interference on a long-lived shared session, not a regression).
- `npx tsc --noEmit` clean for the new/touched frontend files.
- Not independently verified: the full `DictationManager` orchestration wired inside a fully running packaged app with its Whisper/Parakeet models and window (tray toggle, global-hotkey registration on this Wayland session specifically, settings save/load round-trip through the packaged app) -- verified via `cargo check`/`cargo test` and code review against the same live-verified `atspi_injector` primitives, not via a full running app session. Recommend a human/GUI smoke check: toggle dictation from the tray, dictate into a real app, confirm the fallback notification/clipboard-restore timing, and confirm whether `tauri-plugin-global-shortcut` actually registers on the target Wayland/GNOME session (support varies by compositor; the tray toggle is the documented, always-working fallback).

## Documentation
- [x] Documentation updated (Settings page notes the Linux/AT-SPI requirement; `ydotool` documented as an optional broader-coverage path, never auto-configured)

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,6 +198,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "arboard"
+version = "3.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0348a1c054491f4bfe6ab86a7b6ab1e44e45d899005de92f58b3df180b36ddaf"
+dependencies = [
+ "clipboard-win",
+ "image",
+ "log",
+ "objc2",
+ "objc2-app-kit",
+ "objc2-core-foundation",
+ "objc2-core-graphics",
+ "objc2-foundation",
+ "parking_lot",
+ "percent-encoding",
+ "windows-sys 0.60.2",
+ "wl-clipboard-rs",
+ "x11rb",
+]
+
+[[package]]
 name = "array-init"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -377,6 +398,57 @@ name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
+name = "atspi"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6bf601cccedfffec598ec2db1f9d6745885458bccc0e8916d7023f017c94b3d0"
+dependencies = [
+ "atspi-common",
+ "atspi-connection",
+ "atspi-proxies",
+ "zbus",
+]
+
+[[package]]
+name = "atspi-common"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8a79bed3f5b408ce3152f36e07327a845e6ed5d7e2821a89264037dbcc11daf"
+dependencies = [
+ "enumflags2",
+ "serde",
+ "static_assertions",
+ "zbus",
+ "zbus-lockstep",
+ "zbus-lockstep-macros",
+ "zbus_names",
+ "zvariant",
+]
+
+[[package]]
+name = "atspi-connection"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8fab8e4f574f5a7d3af280b38eff25fb6f47a537dac9ae39ce152f52b19fb10b"
+dependencies = [
+ "atspi-common",
+ "atspi-proxies",
+ "futures-lite",
+ "zbus",
+]
+
+[[package]]
+name = "atspi-proxies"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "53403acd3ab2fdb5914f6558da22e540fc07656fce5510f8c02be0e6ef68413e"
+dependencies = [
+ "atspi-common",
+ "serde",
+ "zbus",
+]
 
 [[package]]
 name = "atty"
@@ -636,6 +708,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
+name = "byteorder-lite"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
+
+[[package]]
 name = "bytes"
 version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -760,7 +838,7 @@ version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
 dependencies = [
- "nom",
+ "nom 7.1.3",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


AnsCodeLab:
Withdrawing alongside #721 -- abandoning both of these for now. Deleted the source branch, which is what auto-closed this.

Issue #612, opened 2026-07-14
Feature request: optional remote/networked transcription backend


johnstamly:
## Problem

Meetily's transcription runs entirely in-process inside the Tauri app — WhisperEngine/ParakeetEngine load a model and transcribe locally, with no way to point the app at a transcription backend running on a different machine. This is a real limitation for anyone whose recording machine (e.g. a laptop) doesn't have a strong enough CPU/GPU to run Whisper comfortably, but who owns another machine on their network (a desktop with a real GPU) that could do it much faster.

## Proposed solution

Add a third transcription provider option, e.g. remoteWhisper, alongside the existing localWhisper and parakeet. When selected, Meetily would still capture and mix audio locally exactly as it does today — only the transcription step would be sent over the network to a small standalone server running elsewhere.

This fits cleanly into the existing TranscriptionProvider trait (frontend/src-tauri/src/audio/transcription/provider.rs), which already abstracts over Whisper and Parakeet with a simple transcribe(audio, language) -> TranscriptResult interface. Since that call is already one chunk in, one text result out (the worker pool handles chunking before calling into whichever provider is configured), a remote provider can implement the same trait by making one HTTP request per chunk — no new streaming protocol or worker-pool changes needed.

Concretely, this would mean, as three pieces: first, a new standalone server binary (its own crate, alongside llama-helper) that loads a ggml whisper.cpp model via whisper-rs, the same engine the app already uses, and exposes it over HTTP (POST /transcribe, GET /health), with the same GPU feature flags (cuda/vulkan/hipblas/metal) as the rest of the project. Second, a RemoteWhisperProvider implementing TranscriptionProvider, POSTing audio to a configured server address. Third, a remoteWhisperUrl (plus optional shared-secret) field in transcript settings, alongside a new option in the provider dropdown.

## Why this matters

It keeps Meetily's core privacy promise (nothing leaves infrastructure you control) while letting people split recording and transcription across two machines they own — useful for underpowered laptops, or for teams wanting a shared GPU transcription box on their LAN.

Hus-Mek:
This is already implemented across three open, mergeable PRs:

- **#528** — a pluggable, vendor-neutral HTTPS-backed remote ASR provider, so the app can point at a transcription backend running on a different machine (cloud GPU, self-hosted inference server, etc.).
- **#610** — surfaces the remote provider in **Settings → Transcription Provider**.
- **#533** — documents the `RemoteProvider` contract plus a sample worker.

[...]

Issue #301, opened 2026-01-06
[FEATURE] Custom OpenAI Compatible API Option for TRANSCRIPTIONS


Nels2:
## Feature Description
Want the ability to setup a locally hosted OpenAI Compatible API Server for Transcriptions.

## Problem Statement
Currently you can only choose between `Local Whisper` and `Parakeet`

## Proposed Solution
Implement the ability for users to input their own transcription model, much like how it is currently done with the 'Custom Server (OpenAI)' Option under **Summary** tab.

## User Story
As a power user,
I want to be able to set up Meetily with my locally hosted servers,
So that I do not need to run the models on my local PC or VM, which obviously do not have the firepower to handle running an LLM, nor do I want it to.

## Acceptance Criteria
- [ ] Allow user to input model, endpoint, and API key - just like how it is done in the Summary tab for 'Custom Server (OpenAI)'

## Technical Considerations
 I know the source code has custom providers commented out here, so it seems someone started this journey but stopped.

## Alternatives Considered
None, currently.... besides attempting to rebuild this from source with the added integration, but that's beyond my scope.

## Additional Context
[Add any other context, screenshots, or mockups about the feature request here]

## Checklist
- [ ] I have searched for similar feature requests
- [ ] I have provided all required information
- [ ] I have included any relevant screenshots/mockups
- [ ] I have described the problem and proposed solution clearly

heavyghost:
Can I also add that this would be very useful for enterprise laptops with integrated npu's (IE the Dell Pro 14 with AMD Ryzen AI 7 350 SOC), which are starting to show up more and more as LLM tech matures. Currently meetily runs fairly well but consumes a fair amount of CPU (approx 40%) for Transcription when using the Parakeet Lighting model. 

Using the Custom Server (OpenAI) service you've already implemented for the Summary service and pointing it to a localhost url running the [FastflowLM ](https://github.com/FastFlowLM/FastFlowLM )framework to utilize the onboard NPU, with llama3.2:3b, uses 90-100% NPU when generating the summary but is able to process and complete the report with a minute or so (based off a 1 minute test call).

Implementing a Customer Server (OpenAI) option for the Transcription Service could offload significant load from the CPU to provide better performance and utilize available NPU tech.

Nels2:
@heavyghost I forked this repo and have added all that, I also have speaker diarization set up (although the server doing this must provide speaker segments, etc) See: https://github.com/Nels2/meeting-minutes-cus 

[...]

Issue #225, opened 2025-10-28
[FEATURE] - Transcription managed by separate machine


Norm2023:
## Feature Description
Transcription task is very demanding and slow task. Would be great to have the option to have it run on another remote machine (similarly with AI summary where remote Ollama is possible)

sujithatzackriya:
Did you try the parakeet model?

<img width="1125" height="724" alt="Image" src="https://github.com/user-attachments/assets/89cb2d7c-f623-41ef-b91a-553fbcfa34ca" />

Norm2023:
yes but i find it not so accurate

tcharbonneaumemtronik:
This feature could be very usefull a bit like we can setup the summary to run remotely

Hus-Mek:
This is addressed by open PR **#528** — a pluggable HTTPS-backed remote ASR provider so transcription can run on a separate machine, the same model as the remote-Ollama summary path @tcharbonneaumemtronik mentioned. **#610** surfaces it in Settings and **#533** documents the contract with a sample worker. All three are open and mergeable.

PR #731, opened 2026-08-25
Feature/remote whisper provider

### LINKED ISSUE #637: [BUG] Recording blocked with "Transcription model not ready" when using Whisper provider (hardcoded Parakeet pre-check)

GermaniU:
## Description

Adds a third transcription provider, `remoteWhisper`, that sends audio to a
self-hosted OpenAI-compatible server (`POST /v1/audio/transcriptions`) instead of
loading a model in-process. Audio still never leaves infrastructure the user
controls — it just runs on a machine with a stronger GPU.

Four pieces, one per commit:

1. **The provider** — talks to the server, stores the URL in the existing `model`
   column of `transcript_settings` (no schema change), needs no API key.
2. **Onboarding** — a collapsed "Already running your own AI servers?" disclosure
   lets you pick the remote backend, or defer the summary model, without
   downloading ~670 MB you will never use. The default path is unchanged.
3. **The recording gate** — `useRecordingStart.ts` checked for a local Parakeet
   model regardless of the configured provider, which made the feature unusable.
4. **Settings persistence** — the Server URL field in Settings only updated React
   state; nothing called `api_save_transcript_config`, so a changed URL was lost on
   reload and the recording engine kept using the previous value.

They ship together because 1 without 3 is a dead end: the gate blocks every
recording attempt and offers a download dialog that has no option for a remote
server. And 1 without 4 means the server address can only ever be set during
onboarding — changing it later silently did nothing.

### Notes for reviewers

- **Language sentinels.** The picker emits `auto` / `auto-translate`, which are not
  ISO-639-1 codes; forwarding them makes faster-whisper answer `500`. They are
  dropped, which is exactly how that API is asked to auto-detect.
- **`auto-translate` is hidden for this provider.** `/v1/audio/transcriptions`
  transcribes in the source language — translation lives behind
  `/v1/audio/translations`. Rather than return untranslated text as if it had been
  translated, the option is hidden and the stored preference is coerced on its way
  to the backend, so switching back to Local Whisper restores the user's choice.
- **`/health` proves reachability, not readiness.** The green check in onboarding
  means the server answered; it cannot promise the next transcription succeeds.
  Observed in practice: a server returning `200 {"status":"ok"}` while inference
  was broken.
- **Where the settings persistence lives.** Local Whisper and Parakeet persist from
  inside their model managers (`autoSave`), not from the panel that hosts them.
  None of the three consumers of `TranscriptSettings` saves on its own: the settings
  page has no handler, the modal footer button only closes, and `SettingTabs` passes
  `onSaveTranscript` but the prop is commented out in the JSX. So the remote
  provider persists from inside the component too, matching the model managers and
  fixing all three consumers at once rather than wiring a save handler three times.
- **Save is explicit, not debounced.** A `Save` button writes the URL, with blur and
  Enter as a safety net. Saving per keystroke would write partial URLs such as
  `http://192.168.6` to the database. `Test connection` calls
  `remote_whisper_check_health`, deliberately the same probe
  `validate_transcription_model_ready` runs before recording, so what the button
  verifies is what is checked later. `provider` is forced to `remoteWhisper` on
  save, which also covers re-selecting the provider without retyping the URL.
- **`complete_onboarding` signature changed** (`model: String` → `Option<String>`);
  `None` means the user brings their own summary provider. `ModelStatus` gains two
  optional fields; both use serde defaults, so statuses written by earlier versions
  still load — covered by tests.

## Related Issue

Fixes #637
Implements #612, #301, #225, #527

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Performance improvement
- [x] Code refactoring
- [x] Other

## Testing
- [x] All tests pass

7 unit tests: language sentinel handling, WAV header shape, and onboarding status
backward compatibility. Manually tested on macOS (Apple Silicon) against a
faster-whisper server running `large-v3-turbo` on an RTX 3060 — onboarding with
the remote backend, recording, and live transcription over several sessions.

The settings persistence commit was verified from a locally built macOS bundle:
changing the Server URL, saving, reopening Settings and confirming the new value
survives, plus `Test connection` against a reachable and an unreachable address.
The engine re-reads `transcript_settings` on every validation and every start, so
the change applies to the next recording without restarting the app.

The branch was rebased onto the current `devtest` tip; `tsc --noEmit` and
`next build` are green on the rebased tree.

## Documentation
- [x] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

## Screenshots 
image

## Additional Notes

The recording-gate fix (#637) affects users who never touch remote transcription:
it also unblocks anyone using Local Whisper, which is what that issue reports.

### DIFF
--- a/frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
+++ b/frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
@@ -1,0 +1,288 @@
+// audio/transcription/remote_whisper_provider.rs
+//
+// Remote Whisper transcription provider: sends audio over HTTP to a
+// self-hosted OpenAI-compatible `/v1/audio/transcriptions` endpoint
+// (e.g. a faster-whisper server running on a machine with a dedicated GPU),
+// instead of loading a model in-process on this machine.
+//
+// The endpoint is expected to accept a multipart/form-data POST with a
+// `file` field (WAV audio) and an optional `language` field, and to return
+// JSON shaped like `{"text": "..."}` — the same contract OpenAI's
+// `/v1/audio/transcriptions` uses. Any faster-whisper / whisper.cpp server
+// built against that convention works without changes.
+
+use super::provider::{TranscriptionError, TranscriptionProvider, TranscriptResult};
+use async_trait::async_trait;
+use reqwest::multipart;
+use log::warn;
+use serde::Deserialize;
+use std::time::Duration;
+
+/// Sample rate the rest of the transcription pipeline hands us audio at.
+/// (See `TranscriptionProvider::transcribe` doc comment: 16kHz mono f32.)
+const SAMPLE_RATE_HZ: u32 = 16_000;
+
+/// How long we allow a single transcription request to run before giving up.
+/// Large chunks on a cold GPU (model not yet resident) can take a while.
+const REQUEST_TIMEOUT_SECS: u64 = 60;
+
+/// How long we allow a liveness probe to run. Deliberately much shorter than a
+/// transcription: a probe gates interactive actions (the onboarding "Test"
+/// button, and every recording start), so it must fail fast. A host that is off
+/// or behind a DROP firewall never sends a reset, so without this the probe
+/// would inherit the 60s transcription budget and freeze the UI for a minute.
+const HEALTH_TIMEOUT_SECS: u64 = 5;
+
+/// Meetily's language picker emits these sentinels for "let the engine decide".
+/// They are not ISO-639-1 codes: an OpenAI-compatible server rejects them
+/// (faster-whisper answers 500 Internal Server Error). Omitting the field
+/// entirely is exactly how that API is asked to auto-detect, so the sentinels
+/// must be dropped rather than forwarded.
+const AUTO_LANGUAGE_SENTINELS: [&str; 2] = ["auto", "auto-translate"];
+
+#[derive(Debug, Deserialize)]
+struct RemoteTranscriptionResponse {
+    text: String,
+}
+
+/// Transcription provider that delegates to a remote whisper-compatible HTTP server.
+pub struct RemoteWhisperProvider {
+    /// Base URL of the remote server, e.g. "http://192.168.1.100:8093".
+    /// No trailing slash expected; it is stripped defensively in `new`.
+    base_url: String,
+    client: reqwest::Client,
+    model_label: String,
+}
+
+impl RemoteWhisperProvider {
+    pub fn new(base_url: String) -> Self {
+        let base_url = base_url.trim_end_matches('/').to_string();
+        let client = reqwest::Client::builder()
+            .timeout(Duration::from_secs(REQUEST_TIMEOUT_SECS))
+            .build()
+            .unwrap_or_else(|_| reqwest::Client::new());
+
+        let model_label = format!("remote ({})", base_url);
+
+        Self {
+            base_url,
+            client,
+            model_label,
+        }
+    }
+
+    fn endpoint(&self) -> String {
+        format!("{}/v1/audio/transcriptions", self.base_url)
+    }
+
+    /// Map Meetily's language selection onto what the remote API accepts:
+    /// `Some(iso_code)` to force a language, `None` to let the server detect it.
+    fn normalize_language(language: Option<String>) -> Option<String> {
+        let language = language?;
+        let trimmed = language.trim();
+        if trimmed.is_empty() {
+            return None;
+        }
+
+        let lowered = trimmed.to_ascii_lowercase();
+        if AUTO_LANGUAGE_SENTINELS.contains(&lowered.as_str()) {
+            if lowered == "auto-translate" {
+                // `/v1/audio/transcriptions` transcribes in the source language;
+                // translation lives behind a separate endpoint this provider does
+                // not target. Be explicit rather than silently returning
+                // untranslated text as if it had been translated.
+                warn!(
+                    "Remote Whisper: 'auto-translate' is not supported by \
+                     /v1/audio/transcriptions; falling back to auto-detected transcription"
+                );
+            }
+            return None;
+        }
+
+        Some(trimmed.to_string())
+    }
+
+    /// Encode 16kHz mono f32 PCM samples as a WAV byte buffer (16-bit PCM).
+    /// Self-contained — no extra crate dependency needed for such a simple header.
+    fn encode_wav_pcm16(samples: &[f32]) -> Vec<u8> {
+        let num_samples = samples.len() as u32;
+        let byte_rate = SAMPLE_RATE_HZ * 2; // mono, 16-bit
+        let data_size = num_samples * 2;
+        let riff_size = 36 + data_size;
+
+        let mut buf = Vec::with_capacity(44 + data_size as usize);
+        buf.extend_from_slice(b"RIFF");
+        buf.extend_from_slice(&riff_size.to_le_bytes());
+        buf.extend_from_slice(b"WAVE");
+
+        buf.extend_from_slice(b"fmt ");
+        buf.extend_from_slice(&16u32.to_le_bytes()); // fmt chunk size
+        buf.extend_from_slice(&1u16.to_le_bytes()); // PCM
+        buf.extend_from_slice(&1u16.to_le_bytes()); // mono
+        buf.extend_from_slice(&SAMPLE_RATE_HZ.to_le_bytes());
+        buf.extend_from_slice(&byte_rate.to_le_bytes());
+        buf.extend_from_slice(&2u16.to_le_bytes()); // block align
+        buf.extend_from_slice(&16u16.to_le_bytes()); // bits per sample
+
+        buf.extend_from_slice(b"data");
+        buf.extend_from_slice(&data_size.to_le_bytes());
+        for &sample in samples {
+            let clamped = sample.clamp(-1.0, 1.0);
+            let pcm = (clamped * i16::MAX as f32) as i16;
+            buf.extend_from_slice(&pcm.to_le_bytes());
+        }
+
+        buf
+    }
+}
+
+#[async_trait]
+impl TranscriptionProvider for RemoteWhisperProvider {
+    async fn transcribe(
+        &self,
+        audio: Vec<f32>,
+        language: Option<String>,
+    ) -> std::result::Result<TranscriptResult, TranscriptionError> {
+        if audio.is_empty() {
+            return Err(TranscriptionError::AudioTooShort {
+                samples: 0,
+                minimum: 1,
+            });
+        }
+
+        let wav_bytes = Self::encode_wav_pcm16(&audio);
+
+        let mut form = multipart::Form::new().part(
+            "file",
+            multipart::Part::bytes(wav_bytes)
+                .file_name("chunk.wav")
+                .mime_str("audio/wav")
+                .map_err(|e| TranscriptionError::EngineFailed(e.to_string()))?,
+        );
+        if let Some(lang) = Self::normalize_language(language) {
+            form = form.text("language", lang);
+        }
+
+        let response = self
+            .client
+            .post(self.endpoint())
+            .multipart(form)
+            .send()
+            .await
+            .map_err(|e| {
+                TranscriptionError::EngineFailed(format!(
+                    "Cannot reach remote whisper server at {}: {}",
+                    self.base_url, e
+                ))
+            })?;
+
+        if !response.status().is_success() {
+            let status = response.status();
+            let body = response.text().await.unwrap_or_default();
+            return Err(TranscriptionError::EngineFailed(format!(
+                "Remote whisper server returned {}: {}",
+                status, body
+            )));
+        }
+
+        let parsed: RemoteTranscriptionResponse = response
+            .json()
+            .await
+            .map_err(|e| TranscriptionError::EngineFailed(format!("Invalid response JSON: {}", e)))?;
+
+        Ok(TranscriptResult {
+            text: parsed.text,
+            confidence: None, // faster-whisper server doesn't return a confidence score
+            is_partial: false,
+        })
+    }
+
+    async fn is_model_loaded(&self) -> bool {
+        // "Loaded" here means "reachable" — the remote process owns model lifecycle.
+        // Note this only proves the server answers; it cannot promise the next
+        // transcription succeeds.
+        let health_url = format!("{}/health", self.base_url);
+        matches!(
+            self.client
+                .get(&health_url)
+                .timeout(Duration::from_secs(HEALTH_TIMEOUT_SECS))
+                .send()
+                .await,
+            Ok(resp) if resp.status().is_success()
+        )
+    }
+
+    async fn get_current_model(&self) -> Option<String> {
+        Some(self.model_label.clone())
+    }
+
+    fn provider_name(&self) -> &'static str {
+        "RemoteWhisper"
+    }
+}
+
+/// Tauri command: probe a remote Whisper server before committing to it.
+///
+/// Onboarding and settings both need to tell the user "this URL works" *before*
+/// it is persisted, otherwise the first failure only surfaces mid-recording.
+/// Deliberately reuses `TranscriptionProvider::is_model_loaded` so that what we
+/// validate here is exactly what `validate_transcription_model_ready` checks later.
+#[tauri::command]
+pub async fn remote_whisper_check_health(base_url: String) -> Result<bool, String> {
+    if base_url.trim().is_empty() {
+        return Err("Server URL is empty".to_string());
+    }
+
+    let provider = RemoteWhisperProvider::new(base_url);
+    Ok(provider.is_model_loaded().await)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn auto_sentinels_are_dropped_so_the_server_detects_the_language() {
+        // Forwarding these produced "500 Internal Server Error" from faster-whisper.
+        for sentinel in ["auto", "auto-translate", "AUTO", "Auto-Translate"] {
+            assert_eq!(
+                RemoteWhisperProvider::normalize_language(Some(sentinel.to_string())),
+                None,
+                "sentinel {sentinel} must not reach the server"
+            );
+        }
+    }
+
+    #[test]
+    fn blank_and_absent_languages_are_dropped() {
+        assert_eq!(RemoteWhisperProvider::normalize_language(None), None);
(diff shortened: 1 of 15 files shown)


GermaniU (commented):
## Self-review (agent-assisted) — Remote Whisper provider

Notes on my own PR before asking for review. The four commits are cleanly separated, the messages explain the *why* rather than the *what*, and the change fits the existing abstractions (`TranscriptionProvider` trait, reuse of the `model` column, serde-defaulted status fields with round-trip tests). Nothing here breaks the existing local flow.

One issue I'd fix before merge, plus a few non-blocking follow-ups. Details inline.

### 🔴 Health check couples to a contract the provider doesn't promise
`is_model_loaded()` probes `GET /health`, but transcription uses `POST /v1/audio/transcriptions`. `/health` is **not** part of the OpenAI `/v1/audio/transcriptions` convention — it's specific to whisper.cpp-server / faster-whisper-server. A server that correctly implements `/v1/audio/transcriptions` but doesn't expose `/health` (OpenAI's own API, a minimal proxy, a custom gateway) fails onboarding validation and the recording gate **even though it transcribes fine** — and since Continue is gated on the probe passing, that user is stuck in the wizard with no obvious way out. Prefer `GET /v1/models` (part of the OpenAI contract), or fall back to it on 404, or at minimum document that the server MUST expose `/health`.

### 🟡 Warnings
- `encode_wav_pcm16`: `data_size`/`riff_size` are `u32` with unchecked arithmetic — silent wrap on absurdly long buffers in release. Theoretical for second-long chunks; a `debug_assert!`/`saturating_add` documents the invariant for free.
- No retry on a transient network blip: the local path couldn't lose a chunk this way; the remote one drops the whole POST on a momentary LAN hiccup. One short-backoff retry would avoid gaps in live transcripts. Robustness regression vs local, not a blocker.
- `REQUEST_TIMEOUT_SECS = 60` is fixed and not configurable — a large chunk on a busy server is cut and lost.

### 🟢 Well done
Language coercion lives in `ConfigContext` (not the settings panel) so provider switches without opening the modal stay consistent; the user's real preference is preserved and restored; the double-write in `handleSetSelectedLanguage` was removed to avoid the race; serde backward-compat is covered by two tests; `hasLoadedStatusRef` guards the debounced save and is released in `finally`; the three call sites are unified so they can't drift.

No HTTP-path test for `transcribe()` (would need a mock server) — acceptable, but a `wiremock`/`httpmock` fake would close the gap on the multipart + `{"text": ...}` parsing.

*Reviewed by Hermes (agent) — model: claude-opus-4-8 (Anthropic), via the Hermes gateway — on Germani's behalf.*

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
```diff
@@ -0,0 +1,288 @@
...
+            confidence: None, // faster-whisper server doesn't return a confidence score
+            is_partial: false,
+        })
+    }
+
+    async fn is_model_loaded(&self) -> bool {
+        // "Loaded" here means "reachable" — the remote process owns model lifecycle.
+        // Note this only proves the server answers; it cannot promise the next
+        // transcription succeeds.
+        let health_url = format!("{}/health", self.base_url);
```
GermaniU: 🔴 **Main concern:** this probes `GET {base_url}/health`, but the provider's contract is `POST /v1/audio/transcriptions`. `/health` is **not** part of the OpenAI-compatible convention this file claims to target — it's a whisper.cpp-server / faster-whisper-server extension. A server that implements `/v1/audio/transcriptions` correctly but has no `/health` (OpenAI's own API, a minimal proxy, a custom gateway) will fail this liveness check and therefore fail onboarding validation *and* the pre-recording gate, even though transcription would succeed. Because onboarding blocks Continue on this probe, that user gets stuck with no obvious way out.

Suggest probing `GET /v1/models` (part of the OpenAI contract), or falling back to it when `/health` returns 404, or at minimum documenting in the URL placeholder/tooltip that the server MUST expose `/health` so the promise and the implementation match.

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        Some(trimmed.to_string())
+    }
+
+    /// Encode 16kHz mono f32 PCM samples as a WAV byte buffer (16-bit PCM).
+    /// Self-contained — no extra crate dependency needed for such a simple header.
+    fn encode_wav_pcm16(samples: &[f32]) -> Vec<u8> {
+        let num_samples = samples.len() as u32;
+        let byte_rate = SAMPLE_RATE_HZ * 2; // mono, 16-bit
+        let data_size = num_samples * 2;
```
GermaniU: 🟡 `data_size = num_samples * 2` and `riff_size = 36 + data_size` are `u32` with unchecked arithmetic. In release (wrapping) a buffer above ~2 Gsamples silently produces a corrupt WAV header. Meetily's chunks are seconds long so this is theoretical, but a `debug_assert!` or `saturating_add` costs nothing and documents the invariant.

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
```diff
@@ -0,0 +1,288 @@
...
+                .mime_str("audio/wav")
+                .map_err(|e| TranscriptionError::EngineFailed(e.to_string()))?,
+        );
+        if let Some(lang) = Self::normalize_language(language) {
+            form = form.text("language", lang);
+        }
+
+        let response = self
+            .client
+            .post(self.endpoint())
```
GermaniU, line 171: 🟡 A single transient network blip drops the whole chunk here — the local engine had no such failure mode. For live transcription, one retry with short backoff would avoid gaps in the transcript. Non-blocking, but it's a robustness regression relative to the local path worth a follow-up.

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
```diff
@@ -0,0 +1,288 @@
...
+use serde::Deserialize;
+use std::time::Duration;
+
+/// Sample rate the rest of the transcription pipeline hands us audio at.
+/// (See `TranscriptionProvider::transcribe` doc comment: 16kHz mono f32.)
+const SAMPLE_RATE_HZ: u32 = 16_000;
+
+/// How long we allow a single transcription request to run before giving up.
+/// Large chunks on a cold GPU (model not yet resident) can take a while.
+const REQUEST_TIMEOUT_SECS: u64 = 60;
```
GermaniU, line 27: 🟡 `REQUEST_TIMEOUT_SECS = 60` is hard-coded. Defensible for a cold GPU, but a large chunk on a saturated server is cut at 60s and lost with no recourse. Consider making it configurable if you see timeouts in practice.

frontend/src/contexts/ConfigContext.tsx
```diff
@@ -96,6 +96,25 @@ interface ConfigContextType {
...
+ * deliver — translation lives behind a separate `/v1/audio/translations` endpoint
+ * this app does not call.
+ *
+ * Rust owns the value transcription actually reads (`LANGUAGE_PREFERENCE`, whose
+ * own default is `auto-translate`), and it is written from this context. So the
+ * coercion belongs here, not in a settings panel that only exists while its modal
+ * is open — otherwise a user who switches provider without visiting that panel
+ * keeps requesting a translation that silently never happens.
+ */
+function resolveLanguageForProvider(language: string, provider: string): string {
```
GermaniU, line 111: 🟢 Putting the `auto-translate` coercion here (write path to Rust) rather than in the settings panel is the right call — the comment nails why: a user who switches provider without opening that modal would otherwise keep requesting a translation that silently never happens. And preserving the real preference in localStorage so switching back restores it is exactly right.

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
GermaniU: Good catch — this was wrong for the general case, not just incomplete.

Checked it empirically: OpenAI's own API returns 404 on `/health` and 401 on `/v1/models`; a faster-whisper server is the reverse (200 and 404). So neither endpoint covers the field alone, and probing only one strands a whole family of servers on a false negative — with onboarding blocking Continue on that verdict, exactly as you describe.

Fixed: both endpoints are now tried in turn. 401/403 count as alive, since an endpoint that demands credentials has still proven it exists and answers — treating it as dead would reintroduce the same false negative in a different shape.

<sub>Reviewed and addressed with Claude Code (Opus 5).</sub>

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
GermaniU: Agreed. Switched to `u32::try_from` with `saturating_mul`/`saturating_add`, plus a test covering it.

Unreachable with real chunks (2^31 samples is roughly 37 hours of audio in a single request), but the guard costs nothing and the header can no longer silently disagree with the payload if chunking ever changes.

<sub>Reviewed and addressed with Claude Code (Opus 5).</sub>

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
GermaniU: Correct, and worth stating plainly: this is a robustness regression against the local path, which had no network failure mode at all.

Leaving it as a follow-up rather than folding it in here. Retry with backoff changes the timing of live transcription — how long a chunk can hold up the stream before it is dropped — and that deserves its own change with its own testing, not a ride-along in a PR this size.

<sub>Reviewed and addressed with Claude Code (Opus 5).</sub>

frontend/src-tauri/src/audio/transcription/remote_whisper_provider.rs
GermaniU: Noted. Making it configurable means UI plus persistence plus a migration for the stored value, which is more surface than this PR should carry.

Your framing is the right one — worth doing if timeouts actually show up in practice. Keeping it as a follow-up.

<sub>Reviewed and addressed with Claude Code (Opus 5).</sub>

GermaniU:
## Code Review Summary

**Verdict: Reviewed 💬** (1 issue, 3 suggestions)

**PR:** #731 — Feature/remote whisper provider
**Author:** @GermaniU
**Files changed:** 15 (+1,381 −214)

### ⚠️ Warnings

- **`frontend/src/components/TranscriptSettings.tsx:170-174`** — El payload de `invoke('api_save_transcript_config', {...})` omite el cuarto parámetro `_auth_token: Option<String>` de la firma Rust (`api/api.rs:650-656`). Los parámetros Tauri optionales pueden omitirse del payload, así que esto funciona, pero la firma también declara `_app: AppHandle<R>` — y el orden del payload (provider, model, apiKey) coincide con la firma. Sin embargo, cuando el usuario cambia el provider a `remoteWhisper` desde el dropdown (`onValueChange`), solo se llama `setUiProvider(provider)` y no se persiste el provider — la persistencia solo ocurre vía `saveServerUrl`. Si el usuario selecciona remoteWhisper, NO escribe URL y sale de Settings, el backend queda con el provider anterior. No es un bug del flujo onboarding (que sí persiste todo), pero Settings permite un estado intermedio incoherente entre UI (`uiProvider=remoteWhisper`) y DB (provider anterior). Sugerencia: al cambiar el dropdown, o persistir inmediatamente el provider o revertir `uiProvider` al salir sin guardar.

### 💡 Suggestions

- **`remote_whisper_provider.rs:124` (REQUEST_TIMEOUT_SECS = 60)** — Un chunk de audio largo en GPU fría puede exceder 60s. El comentario del propio código reconoce el caso ("Large chunks on a cold GPU can take a while"). Si el pipeline divide en chunks de N segundos, 60s es holgado; pero un usuario transcribiendo una reunión de 2h con resúmenes parciales podría ver timeouts en el primer chunk. Considera 90-120s o hacer el timeout configurable.

- **`remote_whisper_provider.rs:311-327` (is_model_loaded)** — El sondeo `/health` o `/v1/models` con timeout de 5s es correcto, pero un servidor que responde `200 {"status":"ok"}` durante el arranque del modelo no está listo para transcribir (el propio PR body lo documenta: "a server returning 200 while in..."). El comentario en el código lo admite honestamente. Para onboarding está bien; solo considera un tooltip en Settings → Test connection que diga "reachable ≠ ready" para evitar tickets.

- **`onboarding.rs:548-565`** — `complete_onboarding` acepta `transcription_provider: Option<String>` sin validar el valor contra un enum conocido (`parakeet` | `remoteWhisper`). Un typo persistiría un provider inválido que caería al fallback `"localWhisper" | _` en `engine.rs`. El frontend solo envía valores válidos hoy, pero un validador `match` en el comando Rust cerraría el contrato.

### ✅ Looks Good

- **Diseño de `normalize_language`** — Manejo impecable de los sentinelas `auto`/`auto-translate`: drop en vez de forward, warn explícito, tests cubriendo case-variants. El bug de faster-whisper respondiendo 500 está correctamente neutralizado.
- **`encode_wav_pcm16`** — Header WAV construido a mano con `saturating_mul` para evitar wrap en buffers imposibles; con test que verifica el header byte a byte (RIFF/WAVE/mono/16kHz/16-bit). Sólido.
- **Doble probe `/health` + `/v1/models`** con 401/403 contando como "alive" — cubre ambas familias de servidores whisper sin falsear reachability.
- **`resolveLanguageForProvider` en ConfigContext** — La coerción vive donde está la única write-path hacia Rust, con comentario explicando por qué NO en el settings panel. Razonamiento correcto y documentado.
- **Guard `hasLoadedStatusRef` en OnboardingContext** — Previene que el auto-save debounced sobreescriba una config persistida con defaults de parakeet antes de leer el estado. Sutil y bien cazado.
- **`reportTranscriptionNotReady`** — Un solo lugar para el feedback de usuario, con razón discriminada (`downloading` vs `remote-unreachable` vs `unsupported-provider`) y el model-selector cerrado cuando descargar no es el remedio. Los 3 entry points ya no pueden divergir.
- Los tests Rust (5 en remote_whisper_provider, 2 en onboarding) cubren los contratos reales: roundtrip de status, retrocompatibilidad de deserialización, header WAV.

### 📋 Notas de verificación

- CI: sin checks configurados en la rama (`no checks reported`) — **build y tests no verificables remotamente**.
- Revisión basada en lectura completa del diff (2,271 líneas) + firma real del comando Rust en `api/api.rs` (base `devtest`) confirmada vía API.
- Nota: la línea `apiKey: null` del diff aparece redactada como `apiKey: ***` en herramientas de extracción con redacción automática de secretos; el código real en la rama es `apiKey: null,` — verificado byte a byte. El payload es correcto: `remoteWhisper` no usa API key.

**Sugerencia del revisor: Approve con comentarios.** Los hallazgos no bloquean el merge — son robustez para Settings y contratos más estrictos. Decisión final del maintainer upstream.

<sub>🤖 Revisado por **Hermes Agent** · modelo `glm-5.3` (provider `ollama-cloud`) vía Hermes gateway · skill `github-code-review` v1.4.0 · revisión asistida, no sustituye la aprobación humana.</sub>

PR #737, opened 2026-08-26
fix(transcription): complete post-merge follow-ups

### LINKED ISSUE #734: [Follow-up] Repair historical HE-AAC duration metadata on retranscription
## Follow-up

Follow-up to #608 and #607.

## Problem

For a meeting imported before #608, user-initiated **Enhance / Retranscribe** corrects transcript rows and timing but historically left `metadata.json.duration_seconds` at the old half-duration value.

## Product decision

Repair the metadata only as part of explicit retranscription. Do not scan, migrate, or target historical meetings automatically because existing metadata has no reliable HE-AAC discriminator.

## Acceptance criteria

- Retranscription overwrites existing `metadata.json.duration_seconds` with the just-decoded duration.
- Existing metadata fields remain intact; existing cleanup of `detected_summary_language` remains unchanged.
- A regression test seeds the pre-fix half duration and verifies the full duration after retranscription.
- Packaged-desktop QA retranscribes a historical HE-AAC import and verifies transcript timing plus `metadata.json` duration.

## Release-note copy

Fixed recovery for HE-AAC recordings imported before the duration fix. Open the affected meeting and choose **Enhance / Retranscribe** to regenerate the transcript and correct the meeting metadata duration.

### LINKED ISSUE #735: [Follow-up] Keep Local Whisper download status authoritative during discovery
## Follow-up

Follow-up to #639 and #637.

## Problem

During the early phase of a first Local Whisper model download, a recording-start preflight can rediscover models from disk before a model file exists. The UI then shows missing-model setup even though the native transfer is active.

## Acceptance criteria

- Native Whisper model discovery reports `Downloading` while the model name is present in the active-transfer registry, including absent, zero-byte, and partial files.
- Cached download progress is preserved when available; the initial status is `Downloading { progress: 0 }`.
- Once no transfer is active, existing disk classification remains in force (`Missing`, `Corrupted`, or `Available`).
- Invalid names and all download error, cancellation, and completion paths remove active-transfer membership.
- Packaged-desktop QA verifies home, sidebar-direct, and auto-start show download-in-progress during a first Local Whisper download, and return to model setup after cancellation or failure.

## Scope

No new frontend IPC command or UI state is required. The existing provider-aware readiness flow consumes the corrected native model status.

safvanatzack:
## Description

Completes the post-merge follow-ups from PRs #608 and #639.

- Corrects historical `metadata.json.duration_seconds` during an explicit retranscription, while preserving unrelated metadata.
- Keeps active Local Whisper downloads in `Downloading` state before a usable file exists, including zero-byte and partial downloads.
- Clears native active-download state on cancellation, failure, unsupported models, and success so users are offered setup again after a terminal failure.
- Adds Rust regressions and frontend readiness coverage for initial Whisper download state.

## Related Issues

Refs #734 and #735.

Related: #607, #608, #637, and #639.

## Testing

- [x] Targeted Rust regression tests
- [x] Frontend readiness tests
- [x] Production frontend build
- [ ] Packaged desktop matrix

## Release notes

The user-facing historical HE-AAC recovery note is tracked in #734.

### DIFF
--- a/frontend/src-tauri/src/audio/retranscription.rs
+++ b/frontend/src-tauri/src/audio/retranscription.rs
@@ -736,6 +736,7 @@ fn write_retranscription_metadata(
         let existing = std::fs::read_to_string(&metadata_path)?;
         let mut value: serde_json::Value = serde_json::from_str(&existing)?;
         if let Some(obj) = value.as_object_mut() {
+            obj.insert("duration_seconds".to_string(), serde_json::json!(duration_seconds));
             obj.insert("retranscribed_at".to_string(), serde_json::json!(now));
             obj.insert("status".to_string(), serde_json::json!("completed"));
             obj.insert("transcript_file".to_string(), serde_json::json!("transcripts.json"));
@@ -1012,5 +1013,41 @@ mod tests {
         // Non-audio formats
         assert!(!AUDIO_EXTENSIONS.contains(&"txt"));
         assert!(!AUDIO_EXTENSIONS.contains(&"pdf"));
+    }
+
+    #[test]
+    fn retranscription_metadata_replaces_stale_duration_without_losing_existing_fields() {
+        let dir = tempfile::tempdir().unwrap();
+        let metadata_path = dir.path().join("metadata.json");
+        std::fs::write(
+            &metadata_path,
+            serde_json::to_string_pretty(&serde_json::json!({
+                "meeting_id": "existing-meeting",
+                "duration_seconds": 2523.2,
+                "audio_file": "recording.m4a",
+                "summary_language": "fr",
+                "detected_summary_language": "en",
+                "custom_field": "preserve me"
+            }))
+            .unwrap(),
+        )
+        .unwrap();
+
+        write_retranscription_metadata(
+            dir.path(),
+            "ignored-for-existing-metadata",
+            5046.4,
+            "recording.m4a",
+        )
+        .unwrap();
+
+        let metadata: serde_json::Value =
+            serde_json::from_str(&std::fs::read_to_string(metadata_path).unwrap()).unwrap();
+        assert_eq!(metadata["duration_seconds"], 5046.4);
+        assert_eq!(metadata["meeting_id"], "existing-meeting");
+        assert_eq!(metadata["audio_file"], "recording.m4a");
+        assert_eq!(metadata["summary_language"], "fr");
+        assert_eq!(metadata["custom_field"], "preserve me");
+        assert!(metadata.get("detected_summary_language").is_none());
     }
 }
--- a/frontend/src-tauri/src/whisper_engine/commands.rs
+++ b/frontend/src-tauri/src/whisper_engine/commands.rs
@@ -1,4 +1,4 @@
-use crate::whisper_engine::{is_download_cancelled, ModelInfo, WhisperEngine};
+use crate::whisper_engine::{is_download_cancelled, CancelDownloadOutcome, ModelInfo, WhisperEngine};
 use std::sync::{Arc, Mutex};
 use std::path::PathBuf;
 use tauri::{command, Emitter, Manager, AppHandle, Runtime};
@@ -492,7 +492,7 @@ pub async fn whisper_download_model(
 }
 
 #[command]
-pub async fn whisper_cancel_download(model_name: String) -> Result<(), String> {
+pub async fn whisper_cancel_download(model_name: String) -> Result<CancelDownloadOutcome, String> {
     let engine = {
         let guard = WHISPER_ENGINE.lock().unwrap();
         guard.as_ref().cloned()
--- a/frontend/src-tauri/src/whisper_engine/commands.rs
+++ b/frontend/src-tauri/src/whisper_engine/commands.rs
@@ -1,4 +1,4 @@
-use crate::whisper_engine::{ModelInfo, WhisperEngine};
+use crate::whisper_engine::{is_download_cancelled, ModelInfo, WhisperEngine};
 use std::sync::{Arc, Mutex};
 use std::path::PathBuf;
 use tauri::{command, Emitter, Manager, AppHandle, Runtime};
@@ -463,6 +463,13 @@ pub async fn whisper_download_model(
                 ) {
                     log::error!("Failed to emit download complete event: {}", e);
                 }
+                Ok(())
+            }
+            Err(e) if is_download_cancelled(&e) => {
+                // Cancellation is handled by the existing cancel command/UI path. Do not
+                // emit a terminal event or reject the original request, which would
+                // otherwise overwrite that path's Missing state with an error.
+                log::info!("Download cancelled for {}", model_name);
                 Ok(())
             }
             Err(e) => {
--- a/frontend/src-tauri/src/whisper_engine/whisper_engine.rs
+++ b/frontend/src-tauri/src/whisper_engine/whisper_engine.rs
@@ -935,6 +935,32 @@ impl WhisperEngine {
 
         if result.is_ok() && !active_download.cancellation.is_cancelled() {
             result = self.validate_model_file(file_path).await;
+
+            if result.is_ok() {
+                let expected_min_size = WHISPER_MODEL_CATALOG
+                    .iter()
+                    .find(|model| model.0 == model_name)
+                    .map(|model| ((model.2 as f64 * 0.9) as u64) * 1024 * 1024);
+
+                result = match expected_min_size {
+                    Some(expected_min_size) => match fs::metadata(file_path).await {
+                        Ok(metadata) if metadata.len() >= expected_min_size => Ok(()),
+                        Ok(metadata) => Err(anyhow!(
+                            "Downloaded model file is too small: {} bytes (expected at least {} bytes)",
+                            metadata.len(),
+                            expected_min_size
+                        )),
+                        Err(e) => Err(anyhow!(
+                            "Failed to read downloaded model file metadata: {}",
+                            e
+                        )),
+                    },
+                    None => Err(anyhow!(
+                        "Unsupported model for download validation: {}",
+                        model_name
+                    )),
+                };
+            }
         }
 
         if result.is_err() && !active_download.cancellation.is_cancelled() && file_path.exists() {
@@ -1472,21 +1498,23 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn downloaded_model_sends_user_agent_and_validates_header() {
+    async fn downloaded_model_sends_user_agent_and_rejects_undersized_valid_header() {
         let dir = tempfile::tempdir().unwrap();
         let engine = WhisperEngine::new_with_models_dir(Some(dir.path().to_path_buf())).unwrap();
         engine.discover_models().await.unwrap();
         let (url, request, server) = response_http_server(8, b"ggml\0\0\0\0").await;
 
-        engine.download_model_from_url("tiny", &url, None).await.unwrap();
+        let error = engine.download_model_from_url("tiny", &url, None).await.unwrap_err();
         let request = String::from_utf8(request.await.unwrap()).unwrap();
         server.await.unwrap();
 
+        assert!(error.to_string().contains("too small"));
         assert!(request.to_ascii_lowercase().contains(&format!(
             "user-agent: meetily/{}",
             env!("CARGO_PKG_VERSION")
         )));
         assert!(!engine.active_downloads.lock().await.contains_key("tiny"));
+        assert!(!dir.path().join("ggml-tiny.bin").exists());
         assert!(matches!(
             engine
                 .available_models
@@ -1495,7 +1523,7 @@ mod tests {
                 .get("tiny")
                 .expect("tiny should remain cached")
                 .status,
-            ModelStatus::Available
+            ModelStatus::Missing
         ));
     }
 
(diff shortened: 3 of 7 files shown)


safvanatzack:
Founded issue: cancel → retry loses download ownership: cancelling returns the model to retryable state before the prior worker is guaranteed stopped. A retry can reuse the same model state; the old worker can then clear the retry’s active marker, causing users to see setup/missing while the retry is still downloading.

sujithatzackriya:
## Review: ship after two small fixes

Did a deep pass on this one (correctness, dataflow, KPI, silent-failure, git history, over-engineering) and also cross-checked it against the parallel download-reliability work in Pro #514. Bottom line: **0 blockers, 3 should-fix, 5 worth-noting**, and every finding was reverified against the code. All six of the stated claims hold up.

Nice work on this:

- The cancellation rewrite is not over-engineered. The old flag + 100ms-sleep design could not interrupt a stalled transfer; the per-download `CancellationToken` + biased `select!` + awaited completion latch is the right primitive and matches how cancellation is already done elsewhere in the codebase. No new dependency (`tokio-util` was already there).
- `c756848` correctly closes the cancel-then-retry ownership race, and the net diff also fixes a pre-existing `active_downloads` leak on mid-stream chunk errors (the old `?` returned without cleaning the set). Good catch.
- The cancel UX is now honest: the new `is_download_cancelled` arm in `commands.rs` suppresses the spurious red error toast on a user cancel, and I verified it can only match a real user cancellation (typed `DownloadCancelled`, never a wrapped genuine failure), and that the cancel path genuinely resets state to `Missing`. That is correct, not a swallowed error.

### Should fix (P1)

**1. Progress bar renders `NaN%` on the nested wire shape.**
`ModelStatus::Downloading { progress: u8 }` has a plain derive, so it serializes nested as `{"Downloading":{"progress":N}}`, but the frontend types it flat and reads `model.status.Downloading` as a number (`WhisperModelManager.tsx:522-524`, rendered at `:693`). When the nested shape reaches the card (via `deleteModel`'s raw `setModels` and on initial load when the model is not in localStorage) the readout becomes `NaN%` / `[object Object]`. This is pre-existing, but the new discover "Downloading before a file exists" branch widens how often it happens. It self-heals in about 300ms once a progress event delivers a flat number.
Fix: make the read accept both shapes, for example `typeof s.Downloading === 'object' ? s.Downloading.progress : s.Downloading`, and widen the type in `whisper.ts`. A nested-only retype would break the common path, since the frontend also produces the flat shape internally.

**2. `discover_models` can regress a just-completed model to `Downloading{0}`.**
The worker sets `Available` (`whisper_engine.rs:1140-1142`) while the model is still in `active_downloads`. If `discover_models` runs before `finish_download` removes the key (`:936`), it rebuilds `Downloading{0}` and clears+rewrites `available_models` (`:262-264`), and the success path never re-asserts `Available`. A `load_model` in that window returns Err (`:342-343`). Narrow race, self-heals next poll.
Fix: re-assert the computed final status in `finish_download`'s success branch (one write).

**3. Download client sends no User-Agent.**
`Client::new()` at `whisper_engine.rs:1051` sends no UA, and all whisper models pull from huggingface.co, which returns 403 to blank-UA requests behind Cloudflare. This is pre-existing (base is the same), but since this PR rewrites this exact client it is a natural place to close it. Pro already solved this with a shared `download_http_client()` + `APP_USER_AGENT`. It is not silent (the 403 surfaces as a `model-download-error` toast), so this is a visible-but-opaque-cause first-run failure rather than data loss.
Fix: add `.user_agent(concat!("Meetily/", env!("CARGO_PKG_VERSION")))` on a `ClientBuilder`.

### Worth noting (P2)

- GGML magic-byte validation runs at `discover_models:219` but not on the fresh-download `Available` write, so there is a brief "success then corrupted" window. Cheap to close alongside fix 3.
- `cancel_download` could hang if the worker aborted or panicked before `send_replace`. Not reachable today (no panic sites in the worker, and the download is awaited inline), but a short safety comment would help.
- On a non-cancel download error the partial file and stale `Downloading` status linger until the next `discover_models` self-heals them to `Corrupted`. The error is surfaced; just an untested residue.
- The added frontend test asserts `{ Downloading: 0 }`, which the IPC boundary never actually emits; it passes only via the key-presence check, so the real nested shape stays untested. Ties to fix 1.
- Unrelated to this PR, but the old racy flag+sleep cancel pattern still lives in `parakeet_engine` and the summary `model_manager`. Worth a follow-up so the engines stay consistent.

### Optional cleanup
The two entry/in-loop `is_cancelled()` guards are redundant with the biased `select!`, and `finish_download`'s second map-removal block plus both `_ =>` owner-mismatch warn branches are unreachable (concurrent inserts are already refused while the key is present). Collapsing those is about 18 lines and does not change behavior.

Happy to approve once fixes 1 and 2 land (both are one-liners); fix 3 and the sibling-engine consistency can be a fast follow.

sujithatzackriya:
## Round 2: the fixes landed cleanly. One new deadlock to fix before merge.

Thanks for the fast, thorough turnaround on the round-1 comments. I re-reviewed the 3 new commits (`c756848` to `f7e3440`) deeply and verified each fix against the code. Every round-1 finding is genuinely resolved, and with the User-Agent and content-validation changes the whisper download path now matches the hardening we track elsewhere. Nice work.

One important caveat: the "sync model download state" commit introduced a lock-order deadlock that can wedge model management. It is a small, safe fix. Details and a ready-to-build task breakdown below.

### Round-1 findings: all resolved (verified)

| Finding | Status | Proof |
|---|---|---|
| P1-1 wire-shape `NaN%` | Fixed | nested `{Downloading:{progress}}` migration complete across all 10 whisper sites |
| P1-2 discover regresses to `Downloading{0}` | Fixed | worker no longer sets Available; `finish_download` re-asserts it |
| P1-3 no User-Agent | Fixed | `Client::builder().user_agent("Meetily/{version}")`, asserted on the wire |
| P2-1 GGML validate at download | Fixed | `validate_model_file` before Available; failure to Missing + delete |
| P2-2 cancel hang | Fixed | `cancel_download_with_timeout(5s)` |
| P2-3 asymmetric cleanup | Fixed | any error deletes the partial + sets Missing |
| P2-4 misleading test | Fixed | test asserts the real nested shape |

### New issues found this round

**P0 (merge blocker) - AB-BA lock-order deadlock.**
`discover_models` now acquires `active_downloads` then `available_models` (whisper_engine.rs:242-243). `finish_download`'s cancel branch acquires them in the opposite order: it holds the `available_models` write guard (`let mut models`, :1002, a Drop guard live to the end of the `if cancellation_won` block) and then acquires `active_downloads` at :1008. Both are tokio async locks, and `whisper_get_available_models`, `whisper_download_model`, `whisper_cancel_download` are independent command tasks, so a model-list refresh racing a user cancel can park both futures permanently. `discover_models` is a hot path (import, retranscription, transcription startup, UI), so the collision is reachable; recovery needs an app restart. This is also the only way the `active_downloads` entry can stay permanently reserved (the deadlocked task never signals completion). devtest did not have this (the old `discover_models` took `available_models` alone).
Fix: reorder the cancel branch so it takes `active_downloads` first (matching `discover_models`), i.e. `remove_file` then remove the `active_downloads` entry then set `available_models = Missing`, never holding `available_models` across the `active_downloads` lock. Add a concurrency regression test. Note: simply dropping the guard while keeping `available_models` first does NOT work - it re-opens P1-2.

**P2 - cancel-timeout leaves the card stuck "Downloading".**
On the 5s cleanup timeout, `cancel_download` returns Err, the command maps it to "Failed to cancel download", and the frontend cancel `catch` (WhisperModelManager.tsx:283-289) only toasts. It never clears `downloadingModels` or sets Missing, so the card stays wedged and the message is misleading (the token was cancelled). Reachability is narrow - the biased `select!` makes a normal cancel prompt, so the 5s path realistically only fires under the P0 above or pathological I/O - and it self-heals on a manual refresh. Fix: reconcile the card on cancel-error (refetch / clear state) and reword.

**P2 - Parakeet has the same `NaN%` class the whisper fix resolved (pre-existing).**
`parakeet_engine.rs:31` serializes the nested `{progress}` shape, but `parakeet.ts:20` types it flat and `ParakeetModelManager.tsx:449` reads `.Downloading` as a number, so the Parakeet card renders `NaN%` on the discover path (masked on the live event path). Out of this PR's scope, but a good fast-follow now that the whisper pattern is settled.

**P2 (low) - undersized-download floor is brittle-coupled to the catalog.**
The `size_mb * 0.9` floor is safe for all 12 current models (I checked live HuggingFace `Content-Length` - every model clears by at least ~11%). The residual risk is future drift: a model whose real file drops below 90% of its hand-maintained catalog MB would become permanently un-installable (reject, delete, Missing, re-reject). Cheap hardening: add an exact `downloaded == total_size` truncation check when `Content-Length` is present, and keep the catalog floor only as the fallback when the length is absent (never reject on unknown length).

**Verified non-issues (so nobody re-chases them):** `Corrupted` is not dead code (`discover_models:206/215` still emits it for pre-existing bad files on disk); the `Corrupted -> Missing` change on a failed download is a net UX improvement (Error card + Retry; whisper downloads are not resumable); and the undersized check does not brick any current model.

### Suggested fix plan (each row = one PR)

| Task | Pri | Closes | What | Files | Depends |
|---|---|---|---|---|---|
| **T1** | P0 | deadlock | Unify download lock order (reorder cancel branch) + AB-BA regression test. Owns the canonical invariant (`active_downloads` before `available_models`). | `whisper_engine.rs` (cancel branch + tests) | none - MERGE FIRST |
| **T2** | P1 | cleanup | Collapse the finalize state machine (~-33 lines of dead/duplicated guards), gated by T1's test | `whisper_engine.rs` (disjoint dead regions) | T1 |
| **T3** | P2 | cancel UX | Reconcile the whisper card on cancel-timeout + reword the misleading toast | `WhisperModelManager.tsx` | after T1 (independent) |
| **T4** | P2 | parakeet | Parakeet nested wire-shape migration (mirror of the whisper P1-1 fix) | `parakeet.ts`, `ParakeetModelManager.tsx` | after T1 (independent) |
| **T5** | P2 | download integrity | Exact truncation check when `Content-Length` is present; catalog floor as fallback (never reject on absent length) | `whisper_engine.rs` (loop + size floor) | T2 |
| **T6** | - | cleanup | Decision: warn-only `remove_file` cleanup = keep as-is (self-heals on next discover); a clarifying comment can ride T2 | none (no PR) | - |

**Merge order:** `T1 -> T2 -> T5`; T3 and T4 can merge independently any time after T1. T2 must not land before T1 (shipping the cleanup without the reorder could re-ship the deadlock with no test to catch it).

Only T1 blocks merge; happy to approve once it lands with its regression test. The rest are good fast-follows.

safvanatzack:
## Follow-up — scoped Whisper fixes pushed

| Review item | Status | Resolution / rationale |
|---|---|---|
| **P0: AB-BA lock-order deadlock** | **Resolved** | Cancellation finalization now acquires `active_downloads` before `available_models`, preserves matching ownership through the `Missing` transition, then releases it. Added a deterministic cancellation-vs-discovery regression test. |
| **P2: cancel timeout leaves card stuck `Downloading`** | **Resolved** | Timeout now returns `pending`. The UI shows `Cancelling…`, retains reservation and disables retry, then polls backend discovery once per second until cleanup reaches a terminal state. |
| **Late cancel after successful completion** | **Resolved** | The UI reconciles backend state before changing local status, so a model that completed immediately before Cancel remains `Available` rather than being reset to `Missing`. |
| **P2: Parakeet nested progress / `NaN%`** | **Deferred — out of scope** | This is pre-existing Parakeet behavior. This PR is intentionally limited to Whisper cancellation/finalization; no Parakeet files were changed. |
| **P2: catalog-size floor coupled to future model metadata** | **Deferred — separate hardening** | The current catalog floor is valid for existing models and remains the intentional plausibility policy. Content-Length/exact truncation policy changes need separate scope and tests. |
| **T2: finalize-state-machine cleanup** | **Deferred — avoid refactor** | Not required for the deadlock or timeout correctness fixes. Kept this merge-blocker follow-up behaviorally narrow. |

### Verification

| Check | Result |
|---|---|
| Focused Whisper suite | **13 passed, 0 failed** |
| Scoped TypeScript check | **Passed** |
| `git diff --check` | **Passed** |

The P0 blocker and scoped cancellation-timeout issue are addressed. Remaining items are intentional fast-follows.

sujithatzackriya:
Re-reviewed `eeb4566` deeply (backend concurrency, the new frontend reconcile loop, over-engineering) and verified the lock ordering by hand.

- **P0 AB-BA deadlock: confirmed fixed.** The cancel branch now takes `active_downloads` before `available_models`, matching `discover_models`; those are the only two co-hold sites and both use the same canonical order, so the cycle is gone. The double active-lock is sequential (no self-deadlock), there is no `Downloading{0}` flash, and the `#[cfg(test)]` Notify-barrier interleave test is a real regression guard that fails on any re-inversion and compiles out of release.
- **Cancel-timeout: resolved.** The `CancelDownloadOutcome` contract lines up across Rust/TS, only the whisper manager consumes it, and reconcile-instead-of-force-Missing also closed the late-cancel-after-completion race.
- **All prior fixes preserved**, delta is lean.

One low-reachability residual (the reconcile poll has no ceiling, so a truly wedged worker could stick the card on "Cancelling...") is a good hardening follow-up but not a blocker; the biased `select!` interrupts hung reads and finalize does only fast local I/O, so it is near-unreachable in practice.

The deferred Parakeet nested-progress `NaN%` item is now tracked as a standalone follow-up: #742.

Blocker cleared. Merging.

PR #665, opened 2026-07-21
fix(ui/summary): meeting-details layout and Ollama reasoning strip (#592)

aliCamargo:
## Description
Fixes meeting-details layout and Ollama thinking-model summary handling for #592.

**UI / layout**
- Adds a resizable transcript/summary split view (`MeetingDetailsSplitView`) with Settings-style tabs
- Keeps summary actions in the top toolbar when the summary is empty
- Uses container queries so action label sizing follows panel width
- Minor Logo / MainContent adjustments for the new layout

**Summary / Ollama**
- Leaves Ollama reasoning enabled (does not send `reasoning_effort: none`) so thinking models can reason
- Strips reasoning from the saved notes summary
- Surfaces `reasoning_stripped` / `reasoning` to the frontend and notifies the user when reasoning was filtered out
- Folds separate Ollama `reasoning` into `<think>` so existing cleaners can extract it

## Related Issue
Fixes #592

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe)

## Testing
- [x] Unit tests added/updated
- [x] Manual testing performed
- [ ] All tests pass

### Manual test plan
- [ ] Open meeting details: transcript/summary split is resizable; tabs match Settings-style behavior
- [ ] Empty summary: generate actions stay in the top toolbar
- [ ] Narrow/wide panels: action labels size correctly via container queries
- [ ] Generate summary with a non-thinking Ollama model: summary looks normal; no “reasoning filtered” toast
- [ ] Generate summary with a thinking Ollama model (e.g. qwen3): notes exclude thinking text; toast notes reasoning was filtered
- [ ] Confirm saved summary in the meeting does not retain `<think>` / raw reasoning blocks

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Code follows project style
- [x] Self-reviewed the code
- [x] Added comments for complex code
- [ ] Updated README if needed
- [x] Branch is up to date with devtest
- [x] No merge conflicts

## Screenshots (if applicable)
[Add before/after screenshots of the meeting-details split layout]

## Additional Notes
Commits on this branch:
- `6e2a8ef` — `fix(ui): meeting-details split layout, toolbars, and container queries`
- `2f28bb8` — `fix(summary): keep Ollama reasoning enabled and strip it for notes`

Base branch: `devtest` (not `main`), per CONTRIBUTING.md.

PR #603, opened 2026-07-12
fix(summary): preserve transcript coverage across chunks

ShiroKSH:
## Description

Keep long local-summary chunks contiguous when sentence or word boundaries shorten a window. The next chunk now starts from the boundary that was actually emitted, preserving the configured overlap instead of skipping the gap up to the original fixed-size cursor.

Boundaries that cannot leave forward progress after overlap fall back to the next usable word boundary or the full window. Focused regressions cover the original omission, Unicode text, and overlap equal to the chunk window.

## Related Issue

Fixes #602

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Testing

- [x] Added focused Rust unit tests for coverage, Unicode boundaries, and progress
- [x] Extracted `chunk_text` from the changed source and ran its three regressions with `rustc --test`: 3 passed
- [x] Ran existing frontend library tests with `bun test tests/lib`: 8 passed
- [x] Ran `git diff --check`

## Documentation

- [x] No documentation needed

## Checklist

- [x] Code follows project style
- [x] Self-reviewed the code
- [x] Branch is up to date with `devtest`
- [x] No merge conflicts

## Additional Notes

The hierarchical summary path uses this chunker for long Ollama and BuiltInAI transcripts. Cloud-provider and short-transcript single-pass behavior is unchanged.

Issue #592, opened 2026-07-09
[BUG] Summary panel hidden in normal layout and Ollama/Qwen reasoning text appears in generated summary


TYD-Bruce:
## Bug Description

I found two related issues while testing Meetily v0.4.0 on Windows.

First, after generating an AI summary, the summary may not be visible in the normal desktop layout. The app says the summary is ready, but the summary panel does not appear unless I resize the window into a narrow vertical layout. When I do that, the summary becomes visible, but the transcript view disappears.

Second, when using an Ollama model such as `qwen3.5:9b`, the generated summary may include the model's reasoning/debug text, such as `Thinking`, `Self-Correction`, `Decision Strategy`, and `Strict Interpretation`. This makes the copied/exported meeting summary unusable as clean meeting notes.

I also noticed that the `Enhance` feature can sometimes preserve or introduce transcript errors, especially with proper nouns, technical terms, and mixed English/Chinese content.

## Current Behavior

1. Meetily generates a summary and shows that the AI summary is ready, but the summary panel is not visible in the normal desktop layout.
2. If I resize the app window into a narrow vertical layout, the summary appears, but the transcript view disappears.
3. When using Ollama with `qwen3.5:9b`, the generated/copied summary can include reasoning/debug text such as:
   - `Thinking`
   - `Self-Correction`
   - `Decision Strategy`
   - `Strict Interpretation`
   - `Wait...`
   - `Actually checking the provided text again carefully...`
4. The `Enhance` feature can preserve or introduce incorrect transcript text.

Examples of transcript issues after enhancement:

| Intended text | Enhanced transcript output |
|---|---|
| abnormal outliers | normal outliers |
| different indicators | different指標 |
| Adi | Eddy |
| May data | main data |
| high score | high square |

## Expected Behavior

1. After a summary is generated, the summary panel should always be visible or accessible in the normal desktop layout.
2. Resizing the window should not make one of the main views disappear without a clear tab or toggle.
3. The transcript and summary should be accessible through a stable layout, such as side-by-side panels or clear `Transcript` / `Summary` tabs.
4. When using Ollama models, Meetily should display only the final summary, not the model's reasoning/thinking/debug output.
5. If a model returns reasoning text, Meetily should filter it out before rendering, saving, copying, or exporting the summary.
6. The `Enhance` feature should avoid changing proper nouns, technical terms, or mixed-language transcript text unless confidence is high.

## Steps to Reproduce

[...]

PR #744, opened 2026-08-29
fix(ui/summary): follow up #665 reasoning isolation and sidebar logo

safvanatzack:
## Description
Follow-up to #665. This branch includes its meeting-details layout and Ollama summary work, then hardens the reasoning boundary and restores the expanded-sidebar wordmark to the `devtest` appearance.

**UI / layout**
- Adds the resizable transcript/summary split view with Settings-style tabs.
- Keeps summary actions in the top toolbar when the summary is empty.
- Uses container queries so action labels follow panel width.
- Restores the compact blue **Meetily** pill in the expanded sidebar; the collapsed logo remains unchanged.

**Summary / Ollama follow-up**
- Requests `reasoning_effort: "none"` for Ollama-compatible chat requests, then retries without it when a legacy endpoint rejects that parameter.
- Strips leading `<think>` / `<thinking>` envelopes at each summary-pipeline boundary and rejects responses with no visible markdown after stripping.
- Persists final visible markdown plus the `reasoning_stripped` flag only; redacts legacy persisted reasoning for affected summaries.
- Keeps the desktop and mobile summary/transcript panes on the same accessible tabs implementation, with clamped persisted split ratios for pointer and keyboard resizing.

## Related PR
Follow-up to #665.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Verification
- [x] Manual dev-build test: generated a summary with Ollama `qwen3.5:4b`; the saved final markdown contained no `<think>` block.
- [ ] Automated tests run

## Manual test plan
- [ ] Open meeting details: transcript/summary split is resizable; tabs match Settings-style behavior.
- [ ] Empty summary: generate actions remain in the top toolbar.
- [ ] Narrow and wide panels: action labels follow panel width.
- [x] Generate a summary with Ollama `qwen3.5:4b`: saved notes contain final markdown only, with no `<think>` block.
- [ ] Confirm an Ollama endpoint that rejects `reasoning_effort` retries successfully without it.
- [ ] Confirm a reasoning-only response is rejected rather than saved as a completed summary.
- [x] Expanded sidebar: the compact blue Meetily pill is shown; collapsed sidebar retains its icon logo.

## Documentation
- [x] No documentation needed.

## Additional Notes
Follow-up commits:
- `dc90358` — `refactor(summary): isolate reasoning from persisted meeting summaries`
- `4330504` — `fix(ui): restore compact expanded sidebar logo [skip ci]`

Base branch: `devtest`.


athulchandroth:
## Code Review — reasoning isolation + sidebar logo

Base `devtest` -> `adapt/pr-665-followups`. 19 files (+961/-648), Rust summary pipeline + React/TypeScript meeting-details UI.

**Bottom line: no merge blockers.** Reasoning isolation is well-built — reasoning text is never persisted (only a `reasoning_stripped` bool), it is locked in by a dedicated test, and empty/reasoning-only model output becomes a hard error instead of a blank summary. Strong unit-test coverage. Findings below are worth folding in while the code is open; none strictly block merge.

### P0 — Merge blockers
None. No crashes on common input, no data loss, no security exposure.

### P1 — Should fix

**1. Ollama `reasoning_effort:"none"` is now sent on every request; the safe-retry heuristic is too narrow, opening a new hard-fail path**
`frontend/src-tauri/src/summary/llm_client.rs` (build body ~72, `ollama_rejects_reasoning_effort` ~136-190 + regex 11-16, retry ~408-443)
Requests that never carried this param now always do. The param-less retry only fires on HTTP status exactly 400/422 AND a structured `param`/`field` match or a fully-anchored message regex. An Ollama server that rejects `"none"` with a different status (e.g. 500) or differently-worded/trailing-prose text is not detected, and summary generation hard-fails where it previously worked. Ollama is a first-class local provider, so the blast radius is real.
Suggested fix: retry on any 400/422 whose body mentions `reasoning_effort` (case-insensitive substring). The retry is idempotent and cheap.

**2. Chunk summaries silently dropped -> incomplete summary with no user signal**
`frontend/src-tauri/src/summary/processor.rs:401-405`
Each chunk failing the visible-content guard is logged then dropped; the summary proceeds from surviving chunks and only hard-fails if every chunk drops. A model emitting reasoning-only output for half the chunks yields a plausible-but-incomplete summary, with the loss visible only in the log.
Suggested fix: `warn!` with the drop count, and ideally carry a "partial" flag to the UI like the existing reasoning-stripped toast.

### P2 — Worth noting

Correctness / UX (cheap, in code this PR touched):
- **Timeout message says "60 seconds" but the real timeout is 300s** — `llm_client.rs:388,402`. Misleads triage; one-line fix derives it from `REQUEST_TIMEOUT_DURATION` and also clears a clippy `useless_format`.
- **English-normalization empty result silently discarded with no log** — `processor.rs:118-123`. Safe fallback to pass-1 markdown, but a permanently no-op pass leaves zero trace. Add a `warn!`.
- **Auto-switch effect overrides the user's tab on the narrow (<768px) layout** — `page-content.tsx:144-149`. Force-switches to Summary on every `aiSummary`/`summaryStatus` change; a mobile user reading Transcript is pulled to Summary. Fire once on the transition (a `hasAutoSwitchedRef` guard); desktop unaffected.
- **`.expect()` panic in the Ollama retry path** — `llm_client.rs:422-425`. Unreachable today, latent panic in a network path; use `if let Some(obj)` instead.
- **HTTP status dropped from LLM failure errors** — `llm_client.rs:417, 444-450`. 401 vs 429 vs 500 is what triage needs.
- **`as unknown as Summary` double-cast** — `useSummaryGeneration.ts:277-282`. Honestly commented, but defeats type-checking for all `aiSummary` consumers; a discriminated union fixes it.

Maintainability / structure (non-blocking; most pre-existing but relocated or worsened here):
- **Duplicate divergent Tailwind config** — `tailwind.config.js` (live) vs `tailwind.config.ts` (dead). The PR added the container-queries plugin to both, which is the tell that it was unclear which is live. Delete the `.ts` (fold its unique tokens in first). Same smell in the two PostCSS configs.
- **`generate_meeting_summary` now takes 19 positional args** (this PR added 3); a mis-ordered `Option<&str>` type-checks silently. Bundle stable inputs into a params struct.
- **`processor.rs` grew to four responsibilities** (807 lines): processor + reasoning stripper + language translate/normalize + chunker. Split into `markdown_cleanup.rs` + `language_pipeline.rs` (both already pure and unit-tested).
- **`database` <-> `summary` coupling** (pre-existing, not caused here) — the persistence layer imports `summary::CustomOpenAIConfig`. Compiles fine; relocate the type to a neutral module.
- **`SummaryPanel` Save-dirty highlight reads a ref during render** -> stale until an unrelated re-render — `SummaryPanel.tsx:283`.
- Also: `LazyLock` vs `once_cell` mix in the same module, `localStorage` access unguarded (`MeetingDetailsSplitView.tsx`), `chunk_text` O(n^2) byte-offset recompute, provider boolean-forks vs exhaustive `match`, ~38-prop `SummaryPanel` prop-drilling, dead title props, and `redact_flagged_reasoning` guarding a key nothing writes.

### Architecture health

| Check | Status | Notes |
|-------|--------|-------|
| No circular dependencies | Pass | Zero functional cycles; test code gated behind `#[cfg(test)]`. |
| Clean layer boundaries | Fail (pre-existing) | `database` depends up on `summary::CustomOpenAIConfig`. No new violations; frontend is a strict tree. |
| No god modules | Fail | `service.rs` 1001 lines, `processor.rs` 807. |
| Consistent patterns | Fail | Two Tailwind configs, `LazyLock` vs `once_cell`, two `Summary` types. Container-query migration is the one strongly-consistent new pattern. |
| Accessibility | Pass | Resizer a11y solid (role/aria/keyboard); Logo gained `type`, `aria-label`, `alt`. |
| Concurrency safety | Pass | No `unsafe`, no async-mutex-across-await; cancellation threaded correctly. |
| Performance | Pass (minor) | `chunk_text` O(n^2) offsets; per-request `api_url`/`headers` clones. Slot composition avoids drag jank. |

### Strengths
- Reasoning isolation is sound and safe — reasoning text never persisted; enforced by the `result_json_never_persists_reasoning_text` test.
- Unusually well-tested — every new pure function has focused unit tests.
- Explicit, honest user feedback — a toast reports when reasoning was filtered; `require_visible_markdown` turns empty output into hard errors rather than blank summaries.
- `MeetingDetailsSplitView` is a clean low-coupling seam — panels injected as `ReactNode` slots, stable refs keep separator-drag re-renders off the heavy transcript/editor subtrees.
- The container-query migration (`@container` + `@[Nrem]:` replacing `lg:`/`xl:`) is applied consistently.
- `mod.rs` re-export narrowing reduces the module's public surface with no dangling references.
- Good incidental fix — dropping the `transcripts?.length > 0` guard lets a summary render before paginated transcripts load (previously a blank pane).

athulchandroth:
## Manual test findings

Ran the PR's manual test plan on a local dev build of `adapt/pr-665-followups`. One issue found (Test 2); details below.

| # | Test | Result |
|---|------|--------|
| 1 | Resizable transcript/summary split + tabs | Pass |
| 2 | Empty summary: generate actions in top toolbar | **Fail** — see F-1 |
| 3 | Action labels follow panel width (container queries) | Pass |
| 4 | Reasoning stripped from saved summary (Ollama `qwen3.5:4b`) | Pass |
| 5 | Ollama endpoint rejecting `reasoning_effort` retries without it | Deferred (needs a rejecting/mock endpoint to trigger) |
| 6 | Reasoning-only response rejected, not saved | Deferred (needs a mock endpoint to trigger deterministically) |
| 7 | Expanded sidebar shows compact blue Meetily pill | Pass |

### F-1 — Stopping generation on an empty summary leaves an empty "has-summary" action bar

**Severity:** Medium (UI state inconsistency; can lead to saving an empty summary)

**Steps to reproduce**
1. Open a meeting that has transcripts but no summary.
2. Start generation using the inline generate button in the empty-state body.
3. While it is generating, click Stop.

**Observed**
- After Stop, a summary action bar (Save / Copy / Find / Open Folder) appears over an empty editor. The panel has left its empty state even though no summary was produced.

**Expected**
- After stopping with no summary produced, the panel should return to the empty state (inline generate button + top generate toolbar), with no summary action bar and no empty editor.

**Likely cause**
- `handleStopGeneration` (`frontend/src/hooks/meeting-details/useSummaryGeneration.ts:643-669`) cancels the backend, stops polling, and resets `summaryStatus` to `idle`, but does not reset `aiSummary` and does not guard against a polling result that is already in flight.
- The polling-completion branch (`useSummaryGeneration.ts:265-283`) can still run after Stop and call `setAiSummary(...)` + `setSummaryStatus('completed')`.
- In `SummaryPanel.tsx`, once `aiSummary` is non-null the panel leaves the empty state (`!aiSummary`, line 305) and renders the has-summary layout: the updater action bar (line 279) plus an empty editor (line 311+).

**Concrete risk**
- The user can press Save on this empty action bar and persist a blank summary for the meeting.

**Suggested fix**
- In `handleStopGeneration`, bump a generation/version guard (or set a `cancelled` ref) that the polling-completion branch checks before applying its result, so a late completion is ignored. Do not transition to `completed` or set `aiSummary` for a generation that was cancelled while `aiSummary` was still null.

### Notes
- Tests 5 and 6 are deferred: a current Ollama server ignores unknown params (returns 200) and will not exercise the retry, and it will not reliably emit reasoning-only output, so both need a mock endpoint to test deterministically. Happy to add one if useful.

safvanatzack:
Follow-up to the review:

Addressed in this PR:

- Kept `reasoning_effort: "none"` for modern Ollama, while adding one field-less compatibility retry only for verified 400/422 legacy `reasoning_effort` rejections, including the v0.12 `think value "none" is not supported` response.
- Preserved HTTP status codes in LLM failures, corrected timeout diagnostics to the actual 300-second limit, and removed the retry-path `expect`.
- Made long-transcript chunk processing complete-or-failed: each failed or empty/reasoning-only chunk retries once; a second failure aborts the summary rather than persisting a partial completed result.
- Added visibility for empty English-normalization fallback.
- Prevented repeated mobile tab switching after a user selects Transcript.
- Replaced unsafe Markdown-to-legacy summary casts with the shared `MeetingSummary` union, made editor dirty state explicit, removed dead title-editing wiring, and made split-pane storage failures non-fatal.

Deferred intentionally:

- The remaining performance/architecture/configuration follow-ups are tracked in #745: quadratic chunk offset calculation, the large summary-generation parameter list, processor module boundaries, duplicate frontend build configuration, cross-layer configuration access, and UI prop drilling/provider branching.
- These were not folded into this PR because they require broader design and regression work than the focused correctness fix.

Verification:

- `cargo check --tests -j 1` completed successfully without running tests.
- Manual desktop smoke with Ollama `qwen3.5:4b` completed successfully: one normal request, no compatibility warning, 1/1 chunk processed, and the final Markdown summary was saved.

safvanatzack:
@athulchandroth Thanks for the F-1 report. I manually re-ran the Stop flow on the current branch with a meeting that had transcripts and no summary, and I could not reproduce the empty has-summary action bar.

Observed in the manual run:

- Stop cancelled the backend generation and restored the backup.
- Subsequent `api_get_summary` results reported `has_data: false`.
- The UI remained in the expected empty state: inline Generate Summary and generation controls only; no Save / Copy / Find / Open Folder bar and no empty editor.

The finding is still a potential race rather than a normal Stop-path result: clearing the polling interval does not cancel an already-awaiting `api_get_summary` invocation. A late completed polling result could still reach the callback after Stop and set `aiSummary`.

We will address that stale-polling-result risk in a follow-up, but the reported empty action-bar state was not reproduced in the current manual validation.

safvanatzack:
I found a P2 in the legacy Ollama compatibility-retry path: cancellation is not observed while the compatible error body is read, and the subsequent retry can be polled after Stop.\n\nI’m fixing it by racing error-body consumption with cancellation and prioritizing cancellation before the retry request.

athulchandroth:
## 🔍 Deep Review — reasoning isolation + sidebar logo

Automated multi-agent deep review of head `cd9c57b` vs base `devtest` (merge-base `7c94aa6`) — 23 files, +1133/−727. Findings re-ranked by real-world risk (P0/P1/P2), not each reviewer's in-domain severity.

**Bottom line:** careful, well-tested PR. **No P0 blockers** — no crashes, corruption, or security issues; the reasoning-isolation core is clean, centralized, and unit-tested. Three P1s worth fixing before merge (the top one flagged by two reviewers independently).

---

### 🔴 P0 — Merge blockers
None.

---

### 🟠 P1 — Should fix

**1. Reasoning stripper only handles the *leading* `<think>` envelope — leaks reasoning + can discard a valid summary** `[NEW]`
`processor.rs:23-57` (regexes anchored `^\s*<think…>`), run at every pipeline boundary. Two failure modes:
- **Leak:** a `<think>` block not at the very start (nested, mid/trailing, an early literal `</think>`, or an unclosed tag whose reasoning contains a `#`) is persisted and displayed verbatim in the saved notes — while still stamped `reasoning_stripped: true`, mislabeling it as filtered.
- **Data loss:** an unclosed leading `<think>` before a valid but heading-less summary collapses to `""` → `require_visible_markdown` rejects it → the summary is marked **failed**. Chunk/combine prose summaries are most exposed.

MEDIUM in practice (prompts forbid reasoning, Ollama gets `reasoning_effort:"none"`, reasoning models emit thinking first), but the finding to fix given the privacy-first framing. **Fix:** also strip *paired* `<think>…</think>` anywhere (same code size), keep the leading-unclosed heuristic, add a mid-document test.

**2. `completed` status + unparseable stored JSON → infinite spinner, no error** `[PRE-EXISTING, in scope]`
`commands.rs:262-272` (PR touched the sibling arm at :264). A corrupted/legacy stored `result` returns `data: None` while `status: "completed"`; the poller stops (completed is terminal), the frontend's `completed && data` guard fails with no matching error branch → the UI spins **"Generating…" forever** with no toast, only a Rust `log_error!`. **Fix:** return `status: "error"` when a completed row fails to parse so the existing error toast fires.

**3. First-time generation failure surfaces no persistent error** `[PRE-EXISTING, in scope]`
`SummaryPanel.tsx:290-378` (PR restructured this tree). When the *first* generation fails (`aiSummary === null`), the render falls to `EmptyStateSummary`, which ignores `summaryError`; the error banner lives only in the has-a-summary branch. The user's only signal is an auto-dismissing toast, then a blank empty state. **Fix:** render the banner (or pass `summaryError`) in the `!aiSummary` branch too.

---

### 🟡 P2 — Worth noting

**Reasoning isolation / cleanup**
- **`redact_flagged_reasoning` protects nothing it claims** `[NEW]` (`commands.rs:226-237`) — no code path ever writes a `"reasoning"` key; any leaked legacy reasoning is in the `markdown` field, untouched. Dead/misleading; fix or remove and fold the real cleanup into finding #1 (re-strip `markdown` on read).
- **`content: null` → confusing parse error** `[NEW]` (`llm_client.rs:96-97`) — `#[serde(default)]` covers absent, not present-null (a common reasoning-only shape). Use `Option<String>`.
- **Ollama `reasoning_effort` rejection not memoized** `[NEW]` (`llm_client.rs:79,431-469`) — every pipeline call re-probes+retries on older Ollama; localhost/fast-reject so latency-only. Cache the verdict.

**Frontend / config**
- **Dual Tailwind config — `tailwind.config.ts` is dead** `[NEW]` — only `.js` loads; PR added container-queries to both (works via `.js`). Side effect: `@tailwindcss/typography` lives only in the dead `.ts`, so the notes page's `prose` is **unstyled**. Delete `.ts`, merge into `.js`, verify `prose`.
- **localStorage write inside a `setState` updater** `[NEW]` (`MeetingDetailsSplitView.tsx:80-87`) — impure, double-fires under StrictMode. Persist from a ref.
- **Expanded sidebar wordmark not keyboard-operable** `[PRE-EXISTING]` (`Logo.tsx:34-38`) — a `<span>` DialogTrigger; Tab/Enter can't open About (collapsed variant is a real button).
- **Auto-generate effect stale deps** `[PRE-EXISTING]` (`page-content.tsx:156-177`) — can auto-generate with stale config/transcripts.
- Split-view polish (NaN clamp, `aria-valuetext`, per-frame re-render, mobile flash); `reasoning_stripped` only via the `any` escape hatch.

**Structure (mostly pre-existing)**
- `service.rs` god-module (1001 lines) — extend-not-create; extract helpers post-merge.
- 12–19-arg LLM-config bundle through 5 fns → a borrowed `LlmCallCtx<'a>` struct.
- Dead `summaryResponse`/`isRecording` overlay; unguarded `console.log` of meeting data; `chunk_text` O(n²); wasted full-transcript clone.

---

### Architecture Health

| Check | Status |
|---|---|
| No circular dependencies | ✅ Pass (both Rust & React graphs are clean DAGs) |
| Clean layer boundaries | ✅ Pass (one-way reasoning slice; split-view via ReactNode composition) |
| No god modules | ❌ Fail — `service.rs` 1001 lines (pre-existing, non-blocking) |
| Consistent patterns | ❌ Fail — dual Tailwind config trap |
| Scalable structure | ✅ Pass (strip cheaper than before; retries capped) |
| Accessibility | ❌ Fail — expanded Logo not keyboard-operable; missing `aria-valuetext` |
| Concurrency safety | ✅ Pass (no Mutex across await; biased cancellation proven; no ReDoS; UTF-8-safe) |
| Performance | ✅ Pass (notes: un-memoized Ollama rejection; minor clones/re-renders) |
| Platform conventions | ✅ Pass (static-export/CSR handled; listeners auto-cleaned; localStorage try/catch) |

### Strengths
- Reasoning-strip is **centralized and correct in the common case** — one shared `clean_llm_markdown_detailed`, one `reasoning_stripped` bool OR-combined identically at all 4 boundaries; old whole-string cleaner removed with no dangling refs.
- Reasoning-only/empty responses are **correctly rejected and surfaced** as a real error (not a silent blank).
- **Split-view is robust** — ratio clamped, localStorage try/catch both ways, persisted only on drag-end/keydown, full keyboard a11y, stable-element props (no reconcile mid-drag), zero feature coupling.
- **Strong Rust unit-test coverage** (stripping, retry detection, Ollama-rejection gating, redaction, "never persists reasoning").
- Clean cleanup — dead title-editing wiring removed, triplicated button-group JSX de-duplicated, prop surface reduced.

<sub>🤖 Generated with automated deep review (code · errors · architecture · rust · ts-frontend). Findings are advisory; classifications are for attribution.</sub>

safvanatzack:
Thanks for the follow-up review. We rechecked the findings against **v0.4.0**, the last user-facing release, rather than using `devtest` alone as the baseline. `devtest` contains unreleased work, and this release candidate also includes #665, so “pre-existing in `devtest`” is not by itself a reason to dismiss a finding.

## Addressing in this PR

| Item | Why it belongs here | Planned change / check |
|---|---|---|
| Expanded sidebar wordmark is not keyboard-operable | The expanded About trigger is a non-focusable `<span>`. This behavior also exists in v0.4.0, so it is a baseline accessibility defect rather than a #744 regression, but `Logo.tsx` is already in scope and the semantic correction is isolated. | Replace the expanded trigger with a real `button`, preserve its visual treatment, and verify Tab, Enter, and Space open the About dialog. |
| Stop can become failed during a legacy-Ollama compatibility retry | This is in the #744 retry path. Initial error handling and the retry request observe cancellation, but reading a failed retry body still does not. A delayed retry failure can therefore arrive after Stop and replace `cancelled` with `failed`. | Make the failed-retry body read cancellation-aware using the existing helper. Verify against a controlled endpoint that rejects the first request, delays the retry failure, and preserves `cancelled` after Stop. |

## Intentionally not expanding this PR

| Item | Disposition | Reason |
|---|---|---|
| Non-leading `<think>` / `<thinking>` content | No change | The PR contract is deliberately **leading-envelope** removal. Later tags may be literal Markdown, and current tests preserve them intentionally. We will keep the wording scoped to leading envelopes rather than claiming universal tag removal. |
| Unclosed leading `<think>` with heading-less visible prose | Provider-fixture follow-up | Current behavior can reject that shape. We do not have a normal supported-provider reproduction, and changing it requires an explicit privacy-versus-content policy rather than a speculative parser rule. |
| `content: null` OpenAI-compatible responses | Provider-fixture follow-up | The parser currently treats explicit `null` as invalid. This predates the current change, and no supported provider fixture has been observed yet. We will cover it when a real provider response establishes the expected shape. |
| Completed status with malformed stored JSON | Baseline hardening | This is a recovery path for corrupt or legacy persisted data, not a #665/#744 behavior change. It needs a dedicated fixture and a clear status-repair policy. |
| First-generation failure has no persistent inline error | Baseline UX work | The temporary toast behavior predates this candidate. Improving the empty-state error presentation is worthwhile, but is separate from the reasoning/sidebar changes. |
| Ollama rejection capability is not memoized | Performance follow-up | Older endpoints can incur one bounded rejection/retry per request or chunk. This is latency overhead, not a correctness or persistence failure. |
| Dual Tailwind/PostCSS configuration | Separate configuration cleanup | We confirmed the active Tailwind config lacks the typography plugin while the unused TypeScript config contains it; the Notes `prose` styling issue is therefore a real existing UX defect. It should be tracked accurately, but config consolidation is outside this focused PR. |
| Split-view persistence purity, stale auto-generation dependencies, optional `aria-valuetext`, drag-cost concerns, mobile-tab polish | No change / tracked cleanup | These are baseline polish or maintenance items without a demonstrated release-blocking user path. |
| Large modules, long argument lists, prop drilling, provider branching, and related debt | #745 cleanup track | These need broader design and regression coverage than is appropriate for this focused correctness fix. |

This keeps the PR focused on the two changes that are both concrete and low-risk, while retaining the remaining findings as explicit follow-up work rather than treating unreleased `devtest` behavior as a user-facing baseline.

sujithatzackriya:
## Review — reasoning isolation + sidebar logo

Thanks for the thorough iteration on this PR; the cancellation, chunk-abort, and stop-flow work is genuinely solid. Below is a consolidated review of everything I found, re-verified against the current head, so you can plan and implement the fixes in whatever order suits you. Every finding has a file:line anchor and a short repro/direction.

**Bottom line: recommend HOLD.** The headline goal (reasoning isolation) currently regresses the behavior on `devtest`, and a few reachable user-facing failure paths surface no error. None require a rewrite — all are bounded and testable.

### Priority matrix

| Priority | Count | Theme |
|---|---|---|
| **P0 — blocks merge** | 2 | Private reasoning leaks into saved notes; first-run failures are invisible |
| **P1 — should fix before merge** | 4 | Data-loss on truncated reasoning; infinite spinner on a corrupt row; dishonest "filtered" toast; no failure telemetry |
| **P2 — worth fixing (not blocking)** | 7 | Retry gate, `content:null`, empty-completed, races, blank-save, UX nits |

---

## P0 — merge blockers

### P0-1 · Non-leading `<think>` reasoning leaks into saved notes and the UI

**Where:** `processor.rs:10-13` (both regexes are `^`-anchored, `(?is)^\s*<think(?:ing)?>…`), `processor.rs:23-57` (`clean_llm_markdown_detailed`), test at `processor.rs:834-838`.

**What's wrong:** On `devtest`, `clean_llm_markdown_output` stripped **every** closed `<think>…</think>` block anywhere in the output (a global `replace_all`). This PR replaces that with a stripper that only removes a block anchored at the **start** of the text, and removes the global function. As a result, any `<think>` block that is not the very first token — after a line of preamble, a trailing block, or a second block — now survives into the persisted markdown (`service.rs:145`) and is rendered in the Notes pane. The PR's own test at `processor.rs:834-838` asserts a non-leading block is preserved, so this is currently intended behavior.

**Why it matters:** This is a privacy regression for a privacy-first product, and it re-opens the live behavior reported in #592. Reasoning models commonly emit a short preamble before `<think>`, so this is reachable for typical Ollama / OpenAI-compatible reasoning models.

**Repro:** Model returns `"Here is the summary:\n<think>secret</think>\n# Notes\n..."` → the `<think>secret</think>` is saved and displayed.

**Direction:** Restore a global (unanchored) strip of **closed** `<think>/<thinking>` blocks at every pipeline boundary (this is the `devtest` behavior; a lazy `(?s)<think(?:ing)?>.*?</think(?:ing)?>` `replace_all` is enough). Keep a heuristic for the unclosed-leading case (see P1-2). Also delete `redact_flagged_reasoning` (`commands.rs:226-237`) — it only strips a `"reasoning"` JSON key that no code path ever writes, so it protects nothing; the real fix is stripping at write time.

**Suggested test:** `clean(...)` on non-leading / trailing / two-block inputs returns markdown containing no `<think`/`</think`; an input with a valid block **between** two `<think>` blocks keeps the middle text (lazy match, no over-strip); flip the `:834-838` assertion.

---

### P0-2 · First-generation failure surfaces no persistent error

**Where:** `SummaryPanel.tsx:290-302` (the `!aiSummary` branch), `EmptyStateSummary.tsx:13-19`, `useSummaryGeneration.ts:324-337` (the `allEmpty` branch).

**What's wrong:** When a fresh meeting's **first** summary fails (`aiSummary === null`), the panel renders `EmptyStateSummary`, which has no `error` prop — the error banner and status box are only mounted in the has-a-summary branch. So on first-run failure the only signal is a toast (~4s), after which the UI returns to the pristine "No Summary Generated Yet" state as if nothing happened. The legacy `allEmpty` branch shows nothing at all.

**Why it matters:** This is the highest-frequency failure moment (first Generate: provider down, model missing, empty output, chunk abort) and it looks like success-then-nothing. The user has no way to know it failed or to retry.

**Direction:** Give `EmptyStateSummary` an optional `error` + `onRetry` (+ `onOpenSettings` for a model-required error) and render a persistent inline error card with a **Retry** action when `summaryError` is set and `aiSummary` is null. Add a visible state (+ toast) to the `allEmpty` branch. Reset `summaryError` on retry.

**Suggested test:** rendering the panel with `aiSummary=null, status='error', summaryError='X'` puts the text `X` **and** a Retry control in the DOM (not just a toast).

> Note: the frontend currently has no test runner (no vitest/jest/RTL, no `test` script, no `*.test.*` files). A P0 like this one is a DOM-presence invariant, so you may want to add a lightweight harness as part of this work so the fix is provable rather than manual-only.

---

## P1 — should fix before merge

| ID | Finding | Where | Why it matters | Direction |
|---|---|---|---|---|
| **P1-1** | A `completed` row whose stored result JSON is unparseable/absent returns `data:null` while status stays `completed`; the poller stops and no frontend branch handles `completed && !data`, so the UI spins **"Generating…" forever** with no error and no telemetry. | `commands.rs:262-272` (`api_get_summary`), `useSummaryGeneration.ts:281`, `SidebarProvider.tsx:232` | Latent (needs a corrupt/legacy row) but is an unrecoverable hang when hit. | On an unparseable completed result, return `status:"error"` with a diagnostic (Rust), and add a `completed && !data → error` branch (frontend). Gate the downgrade on `status=='completed'` only. |
| **P1-2** | An **unclosed** leading `<think>` before a heading-less body collapses to `""` (the recovery requires an ATX `#` heading via `unwrap_or_default()`), so `require_visible_markdown` rejects it and the whole run is marked **failed** — discarding a valid summary. | `processor.rs:31-38`, `:59-67` | Data loss on large-transcript (chunked) runs; the chunk prompt asks for heading-less output, so the ATX requirement is mismatched. `devtest`/#665 fell back to the original text here; this PR removed that. | Recover from the first non-empty content line instead of requiring a heading. The recovery must **re-run** the closed-block strip so a later block can't leak — assert no-leak AND no-loss in the same test. |
| **P1-3** | The "Model reasoning was filtered out of the notes" success toast is driven by the backend `reasoning_stripped` flag, which is true whenever *any* leading strip happened — **not** whether reasoning actually remains. So with P0-1 it shows leaked reasoning **and** claims it was filtered. | `useSummaryGeneration.ts:295,302-304` | Actively misleads the user about a privacy guarantee. | Fix with P0-1: show the "filtered" copy only when the persisted markdown provably has no `<think` residual (a `hasResidual(markdown)` check the frontend already has the data for); define `reasoning_stripped` as "removed something AND no residual remains". |
| **P1-4** | Several terminal paths fire a `started` analytics event but **no terminal event**: user Stop, backend cancel, the P1-1 hang, and the English-normalization fallback (which seals as a plain success). The event also has no field to distinguish a cancel from a failure. | `useSummaryGeneration.ts:176-213`, `:659-686`; `processor.rs:123-149`; analytics event shape at `analytics.ts` (`success/duration/error_message` only) | A user-Stop is indistinguishable from a silent crash in analytics, corrupting the started→completed metric. | Emit exactly one terminal event per generation (a per-generation dedup ref avoids double-emit on Stop-then-poll). Add an `outcome` property so `cancelled`/`fallback`/`error`/`empty`/`ok` are separable; mark the normalization fallback as a distinct `fallback` (still a success, but warned). |

---

## P2 — worth fixing (not blocking)

| ID | Finding | Where |
|---|---|---|
| **P2-1** | Ollama `reasoning_effort` retry is gated to HTTP `400/422`; a proxy rejecting with `500/404` hard-fails a summary that previously worked. Gate the param-less retry on the body match, not the status. | `llm_client.rs:168` |
| **P2-2** | An OpenAI-compatible response with `"content": null` hard-fails deserialization (`#[serde(default)]` only covers an absent key) — and `content:null` is exactly what reasoning-only turns send. Use `Option<String>`. | `llm_client.rs:96-97` |
| **P2-3** | A title-only output (`"# Meeting Title"` with no body) passes the visible-markdown guard (checked pre-strip) but `strip_title` returns `""`, so a `completed` row is persisted with empty markdown. Reject/flag when the post-strip markdown is empty. | `service.rs:145`, `processor.rs:497` |
| **P2-4** | The cancel and complete paths both write the terminal row with no compare-and-set, so a Stopped summary can be persisted mislabeled `cancelled` while holding the generated content, and re-surface on reload. | `commands.rs:440`, `service.rs:568` |
| **P2-5** | The cancellation registry is keyed only by `meeting_id` with unguarded cleanup; two overlapping regenerations make one uncancellable. Reachable only if the Generate/Stop state machine can be bypassed. | `service.rs:199-227` |
| **P2-6** | The empty-summary guard is bypassed by two earlier `return`s (`summary_json:[]` and whitespace `markdown`), so a previously-persisted empty payload can still render a has-summary action bar and let the user Save a blank. Apply the same emptiness check to both. | `page.tsx:212,233` |
| **P2-7** | UX nits: mobile tab auto-switch can yank a manually-selected Transcript tab the first time a summary appears (`page-content.tsx:145`); the resize separator has no `aria-valuetext` (announces a bare number) (`MeetingDetailsSplitView.tsx:149`); the empty-models hint recommends `gemma2:2b` in one place and `gemma3:1b` in another (`SummaryGeneratorButtonGroup.tsx:204` vs `useSummaryGeneration.ts:514`). | various |

---

## What's already solid (please keep as-is)

- **Cancellation is thorough** — `await_or_cancel` wraps every await/IO point with cancel-first selection, proven by a test. Nice.
- **Chunk retry-then-abort** is complete-or-fail and preserves cross-chunk synthesis (a real improvement over the base's silent chunk drop).
- **The stop / save-blank race is closed** via the monotonic `generationIdRef` guard.
- **Timeout diagnostic corrected (60s → 300s), HTTP status now in error strings, `.expect()` removed.**
- **Sidebar About is a real keyboard-operable button**, and the split-view is drag-resizable with proper `role="separator"` keyboard handling and non-fatal storage.
- No collateral regressions to #694 (Claude thinking-block parsing) or #603 (chunk coverage).

---

## Two open decisions worth settling in your plan

1. **Fenced `<think>` in the strip fix (P0-1):** should a `<think>` inside a legitimate ``` code block in a summary be stripped too, or preserved? Stripping everywhere (matching `devtest`) is simplest and closes the leak; preserving needs a fence-aware scanner. Recommend strip-everywhere unless you expect summaries to legitimately contain `<think>` in code.
2. **Crash-orphan / corrupt rows:** P1-1 covers the corrupt-row hang; a related edge is a row left in `generating` after an app quit mid-run. Worth deciding whether to fold a guarded stale→error recovery in now or track it separately.

---

## Suggested sequencing

1. **P0-1 + P1-3** together (same "no residual" invariant) — the privacy blocker and its honest toast.
2. **P1-2** (pair its test with P0-1's no-leak assertion).
3. **P0-2** (first-run error surface; consider adding a test harness here).
4. **P1-1** (corrupt-row hang) and **P1-4** (terminal telemetry) — do these together since they share the failure paths.
5. P2s as capacity allows.

Happy to dig into any of these further.

athulchandroth:
## Deep Review — reasoning isolation + sidebar logo

**Verdict:** Well-structured, unusually well-tested change (Rust summary files are ~40–45% tests; cancellation, hydration, and stale-async guards were independently verified sound). **No P0 merge-blockers** — no crash, data-corruption, or security defect. Three P1 findings cluster on one theme: the reasoning-isolation parsing is too brittle for the exact reasoning providers this PR targets. Two of the three were flagged independently by two reviewers each.

---

### 🟠 P1 — Should fix before merge

**1. `content: null` from a reasoning provider fails the entire summary request**
`llm_client.rs:95-102` · [NEW]
`MessageContent.content` uses `#[serde(default)]`, which only fills an **absent** key — a present `"content": null` fails with *"invalid type: null, expected a string"*. The OpenAI schema declares assistant `content` as `string | null`, and the providers that null out `content` (answer in `reasoning`/`reasoning_content`) are **exactly the reasoning models this PR exists to support**.
- **Failure mode:** every summary against such a provider dies with "Failed to parse LLM response" — the shipped feature is broken for a whole provider class. The new test only covers the absent-key case.
- **Fix:** `content: Option<String>` + `.as_deref().unwrap_or("").trim()`, and add a `null`-case test.

**2. An unclosed leading `<think>` tag throws away the whole answer**
`processor.rs:32-39` · [NEW]
When output opens with an unclosed `<think>` and the real answer has **no `#` heading**, `MARKDOWN_HEADING_REGEX.find(...).unwrap_or_default()` yields `""` and discards everything — silently, with `reasoning_stripped=true`. Chunk/combine prompts request concise **prose** (usually heading-less) and run only for Ollama/BuiltInAI — the local reasoning models this PR targets.
- **Failure mode:** one truncated `<think>` turns recoverable noise into a hard abort — *"transcript section N of M failed after 2 attempts."* Good content is dropped and nothing logs it.
- **Fix:** fall back to the post-tag remainder (`.unwrap_or(after_open_tag)`) + `warn!` when discarding; let the empty-content guard catch only genuinely empty output.

**3. Non-leading `<think>` reasoning leaks into the saved summary**
`processor.rs:10-13, 27-39` · [NEW]
Both strip regexes are anchored `^\s*<think…>`, so reasoning is removed **only** when it's the very first content. `Sure, here's the summary:\n<think>…</think>\n# Report` persists the reasoning verbatim — the exact leak the PR set out to prevent, for a privacy-first product. This is a **deliberate trade-off** vs. the removed global-strip, so it needs a conscious decision.
- **Fix:** also strip **closed** `<think>…</think>` envelopes anywhere (only when `</think>` actually closes), keeping the leading-open-tag heuristic for the unclosed case — or explicitly accept the trade-off in a comment.

---

### 🟡 P2 — Worth fixing (no immediate failure mode)

- **Misleading "no content" error** (`processor.rs:59-67`, [NEW]) — same message whether the model returned nothing or the cleaner (P1 #2) dropped valid content. A parser bug looks like a model problem; user is told to retry a model that worked. Log raw `content.len()` at the failing sites.
- **Dual Tailwind + dual PostCSS config** (`tailwind.config.{js,ts}`, [NEW-amplified]) — `.js` wins resolution, so `.ts` is dead config, yet the PR edited both to add `@tailwindcss/container-queries`. A future one-sided edit silently kills every `@container`/`@[40rem]:` label. Sharpest bit: `@tailwindcss/typography` lives **only** in the dead `.ts` — worth verifying `prose` in `notes/[id]/page.tsx` is styled today. Collapse to one authoritative config.
- **`llm_client.rs` doubled (397→918); `generate_summary` is a ~257-line monolith** ([NEW]) — no failure mode, but it's now the code most likely to be re-edited per provider. Decompose into `resolve_endpoint`/`build_*_body`/`execute_with_retry`/`parse_completion`.
- **Failed chunk is now fatal to the whole run** (`processor.rs`, [NEW]) — more correct than the old silent skip, but one flaky chunk now aborts a long transcript. Confirm this is intended.
- **Pervasive `any` on the reworked data flow** (`useSummaryGeneration`, `useMeetingData`, `useCopyOperations`, `page.tsx`, [PRE-EXISTING]) — Tauri response shapes and `meeting` are `any` exactly where this PR threads `.markdown`/`.reasoning_stripped`/`.status`. The new `MeetingSummary` union is good but stops at the hook boundary. Add typed response interfaces; reasonable as a follow-up.
- **`SummaryPanel` 32-prop drilling** ([PRE-EXISTING]) — the PR modified the contract without shrinking it; group into `summary`/`modelConfig` objects or consume context.

Smaller [NEW] hardening worth a quick pass: legacy `reasoning`/`reasoning_content` redaction only fires when `reasoning_stripped==true` (`commands.rs:226-238`); error-body read failure discards the real `reqwest::Error` for `"Unknown error"` with no log (`llm_client.rs:432-449`); stringly-typed `error.contains("cancelled")` cancellation classification; superseded-generation `catch` blocks swallow errors with zero trace (`useSummaryGeneration.ts`).

---

### Architecture Health

| Check | Status | Notes |
|-------|--------|-------|
| No circular dependencies | Pass | No build-breaking/init-order cycles. Only `summary`↔`database` bidirectional coupling — pre-existing, benign intra-crate, untouched. |
| Clean layer boundaries | Pass | No NEW violations; new `MeetingDetailsSplitView` is props-only, all new edges app→feature. |
| No god modules | ⚠️ Fail | `service.rs` 1001, `llm_client.rs` 918 (**doubled this PR**), `useSummaryGeneration.ts` 696, `SummaryPanel` ~32 props. Mostly inherited; Rust ~40–45% tests. |
| Consistent patterns | Pass | Strong positive drift (named structs, `pub(crate)`, container queries, consolidated `MeetingSummary`). Deviation: dual config. |
| Scalable structure | Pass (watch) | Growth risks: provider addition is shotgun surgery, Tauri command double-registration. Reasoning isolation is data-driven, scales well. |
| Accessibility | Pass | Real `<button>`+`aria-label`+focus ring on Logo; `role="separator"` + keyboard on the resizer. One LOW `aria-labelledby` nuance. |
| Concurrency safety | Pass | `await_or_cancel` biased-select verified via real TCP-socket tests; `generationIdRef` stale-async guards sound. |
| Performance | Pass (one note) | Ollama `reasoning_effort` probe re-paid every pass (2× round-trips) against legacy servers — LOW, self-healing. |
| Platform conventions | Pass | Next.js 14 static-export + Tauri handled correctly; hydration-safe init. Rust idioms clean. |

### Strengths

- **Reasoning isolation is data-driven** — `LlmCompletion { content, reasoning_stripped }` computed once from Anthropic `thinking` or OpenAI-compat `reasoning`/`reasoning_content`; adding a provider needs no new reasoning branch.
- **Net-positive error handling** — `require_visible_markdown` hard-fails empty/reasoning-only output instead of saving blank summaries; HTTP status codes now surfaced; `page.tsx` stops rendering data-less processes as empty `{}`.
- **`MeetingDetailsSplitView`** is a clean isolated extraction that *reduces* layout coupling.
- **Strong Rust test coverage**, including real TCP-socket cancellation tests.
- **Frontend done right** — hydration-safe init, verified stale-async guards, a11y upgrades, `any`→typed tightening in `useCopyOperations`.

---

The three P1s share one root: the `<think>`/`content` parsing assumes well-formed, leading-only, heading-structured output from models that routinely violate all three. Fixing #1 and #2 is small and high-value; #3 needs a design call. Everything else is safe to land or defer.

<sub>🤖 Generated with Claude Code deep-review (9 specialist agents: code, error-handling, 5× architecture, Rust, TS-frontend).</sub>

safvanatzack:
@athulchandroth Thanks for the deep review. I pushed `03dfb7b` with the bounded lifecycle/reasoning fixes.

### Resolved

- **`content: null` from OpenAI-compatible reasoning providers** — `MessageContent.content` is now `Option<String>` and absent/null content normalizes to an empty string before visible-content validation. Added null-content coverage.
- **Non-leading, trailing, multiple, and fenced closed `<think>` / `<thinking>` blocks leaking into notes** — closed reasoning envelopes are now stripped globally. Any residual marker is rejected before persistence, save, hydration, copy, or rendering.
- **Misleading generic no-content failure** — unterminated reasoning markers and genuinely empty post-sanitization content now return distinct errors. The sanitizer also logs the malformed-marker condition.
- **Competing Tailwind/PostCSS configurations** — removed the dead TypeScript Tailwind config and incomplete PostCSS module config. `tailwind.config.js` is authoritative and includes typography plus container-query plugins; `postcss.config.js` retains Tailwind and Autoprefixer.
- **Legacy reasoning-key redaction / error-body fallback / string cancellation classification** — summary save/read boundaries validate reasoning markers independently of a flag; HTTP errors retain body-read failures instead of reporting `Unknown error`; terminal cancellation uses `CancellationToken::is_cancelled()`.
- **Superseded polling lifecycle** — polling, cancellation, database terminal writes, and cleanup are now bound to the exact generation `start_time` token. Old responses cannot repaint, stop, cancel, or clean up a newer generation.

### Deliberate decisions / remaining follow-ups

- **Unclosed leading `<think>` with valid prose after it** — I did **not** salvage text after an unclosed marker. The old heading-based silent truncation is removed, but malformed reasoning envelopes now fail visibly and retryably. This is a deliberate privacy-first fail-closed rule: ambiguous content is never persisted.
- **`llm_client.rs` decomposition** — deferred. The change fixes provider parsing and lifecycle correctness without a broad provider/client refactor.
- **Failed chunk aborts the entire run** — retained intentionally. One retry is allowed; a failed chunk makes the summary fail rather than persisting a partial summary.
- **Broader `any` cleanup and `SummaryPanel` prop grouping** — summary process/data boundaries are now typed and runtime-validated. Pre-existing meeting/BlockNote `any` values and the remaining wide panel contract were not broadened into this fix.
- **Repeated Ollama compatibility probe and the low `aria-labelledby` note** — deferred as non-blocking follow-ups; no measured probe regression or concrete accessibility failure was demonstrated.

safvanatzack:
@sujithatzackriya Thanks for the consolidated review. I pushed `03dfb7b`, addressing the privacy, terminal-state, failure-surfacing, and empty-content findings.

### P0 findings

- **P0-1 — Non-leading reasoning leaks into saved notes/UI** — fixed. Closed `<think>` / `<thinking>` envelopes are stripped globally before persistence. Residual/unclosed markers fail validation, so the Notes UI cannot render a successful summary containing those markers.
- **P0-2 — First-generation failures have no persistent error** — fixed. `EmptyStateSummary` now receives the error, renders it with `role="alert"`, and exposes **Retry summary**. Preflight, empty-output, backend, and defensive completed-without-data failures all enter that visible error state.

### P1 findings

- **P1-1 — Completed corrupt/absent summary rows leave the UI generating forever** — fixed. A completed row whose data is missing, malformed, blank, or non-renderable is returned as `error`; the frontend also defensively treats `completed && !data` as a retryable failure.
- **P1-2 — Unclosed leading reasoning discards heading-less valid text** — deliberate policy decision: the old silent heading-based discard is removed, but malformed reasoning markers now fail closed rather than salvaging post-marker text. The error is visible and retryable; ambiguous text is not persisted.
- **P1-3 — “Reasoning filtered” success toast can be dishonest** — fixed. A successful result now has no residual reasoning marker; otherwise it is rejected rather than shown with a filtered-success message.
- **P1-4 — Started analytics lacks a terminal outcome** — fixed with the existing analytics contract. Each started attempt records exactly one terminal event using stable categories: `ok`, `fallback`, `generation_error`, `empty_result`, or `cancelled`. Raw provider/backend errors are not sent as analytics values.

### P2 findings

- **P2-1 — Expand Ollama retry beyond 400/422** — not changed intentionally. The compatibility retry remains one-shot and restricted to demonstrated 400/422 parameter rejection; retrying generic 404/500 responses could conceal endpoint misconfiguration or provider failures.
- **P2-2 — `content: null` response failure** — fixed via `Option<String>` normalization.
- **P2-3 — Title-only output persists empty Markdown** — fixed. Summary result creation rejects output with no visible content after stripping a leading H1.
- **P2-4 — Cancel/complete terminal-write race** — fixed. Terminal database updates use compare-and-set conditions on the exact `start_time` attempt token while the row is pending.
- **P2-5 — Meeting-only cancellation registry can lose overlapping generation** — fixed. Registry entries contain `{ started_at, token }`; cancellation and cleanup operate only on the matching generation.
- **P2-6 — Blank/legacy payload bypasses empty-summary checks** — fixed. Shared content validation rejects blank Markdown, blank BlockNote payloads, empty legacy sections, and residual reasoning markers before render or save.
- **P2-7 — Mobile tab override, separator text, Gemma model mismatch** — fixed. A manually selected mobile Transcript tab is not overridden by the first completed summary; the separator exposes `aria-valuetext`; guidance and auto-configuration consistently use `gemma3:1b`.

### Open decisions

- **Fenced reasoning blocks:** resolved by stripping closed envelopes everywhere, including fenced content, to preserve the privacy invariant.
- **Stale `generating` rows after an app crash:** still deferred. This fix targets active-attempt identity and terminal-write races, not crash-orphan recovery.

athulchandroth:
## Deep Review: reasoning isolation + sidebar logo

Reviewed `adapt/pr-665-followups` -> `devtest` (32 files, +2429/-1827) with focus on the reasoning-isolation path and the summary-pipeline refactor. Findings are re-prioritized by production risk.

**Verdict: no merge blockers.** The reasoning-isolation design is strong — four-layer, fail-closed, no ReDoS, concurrency verified. But its headline feature has one verified hole, and the core detection logic is triplicated and already drifting.

---

### P1 — Should fix

**1. Legacy-format summaries leak raw reasoning (the exact thing this PR exists to prevent)**
`summary-content.ts:12-63` · `commands.rs:282-295` + `redact_flagged_reasoning ~:328-350`

Verified by executing the guard. Both the TS and Rust reasoning detectors recurse only into a key allowlist (`summary_json`/`children`/`content`/`blocks`). Legacy summaries are keyed by section names (`Decisions`, `ActionItems`, ...), so neither detector ever reaches the inner `blocks[].content` — yet the sibling visibility checks understand that shape and mark it renderable.

- Failure mode: a pre-existing/migrated legacy-format summary containing `<think>...</think>` (e.g. an older row from a reasoning model like DeepSeek-R1/QwQ, reopened after upgrade) renders raw chain-of-thought to the user and is copyable. It passes the Rust save/read gate, is untouched by redaction, and passes the frontend gate. Narrow (only legacy rows with reasoning; new summaries are always markdown), and no test covers it.
- Fix: make both detectors structural — scan every string leaf, recurse every array element and object value — instead of key-whitelisting. One change; also closes P2 item 1. Add the failing legacy test.

**2. UI can wedge into a permanent fake "Generating..." state**
`commands.rs:516-522` (swallowed cleanup) + `commands.rs:529-544` (dropped `JoinHandle`)

- Failure mode: if `save_transcript_data` errors, `let _ = update_process_failed(...)` discards the result, leaving the row stuck `pending` with no log; likewise a panic in the fire-and-forget background task. There is no active poll to hit the 15-min timeout, so the spinner spins forever with no recovery. DB errors/panics are uncommon but plausible, and the consequence is unrecoverable-looking.
- Fix: inspect the result and log on `Err`/`Ok(false)`; wrap the spawned task body so a panic still marks the row failed.

**3. Auto-generation silently skips when transcripts load after mount** (pre-existing, in heavily-refactored file)
`page-content.tsx:157-178` — effect deps are only `[shouldAutoGenerate, meeting.id]`

- Failure mode: if `shouldAutoGenerate` is true before transcripts arrive (`length === 0`), the effect no-ops and never re-runs when they load, so the summary never auto-generates, and stale `handleGenerateSummary`/`modelConfig` are captured. Add `transcripts.length` (or a `hasTranscripts` gate) to deps.

---

### P2 — Worth noting

- **Reasoning logic triplicated across Rust/TS and already diverged** (`summary-content.ts` <-> `processor.rs` <-> `commands.rs:249-350`). Marker regex is copy-pasted; Rust does not recurse object-valued `content` (BlockNote table cells) or the `blocks` key, so the authoritative DB gate is weaker than the TS gate (the frontend still catches the render, so no live leak). The structural fix in P1 item 1 closes this; also define the regex once per language plus a shared fixture list both sides must classify identically.
- **Save-guard refuses valid edits with a misleading message** (`commands.rs:106-108` + `processor.rs:51-53`): one string conflates "empty" vs "reasoning marker", and the regex false-positives on prose like `"<think we should ship>"`. Fail-closed, but blocks legitimate manual edits.
- **Empty catch swallows stale-generation cancel** (`useSummaryGeneration.ts:213`), so an orphaned LLM job keeps burning tokens silently. Log it.
- **`meeting: any` across new hooks/props** (`useMeetingData.ts:10`, `useCopyOperations.ts:10,146`, `page-content.tsx:37`) defeats type safety on the clipboard/Tauri paths; `MeetingMetadata`/`Transcript` already exist.
- **Pre-existing scale/correctness in scope:** `token_threshold - 300` underflow for tiny-context models -> `saturating_sub` (`processor.rs:403`); `chunk_text` O(n^2) offset recompute (`processor.rs:299-334`); cloud path skips chunking -> hard-fails on long transcripts (`service.rs:474-476`); read path never re-sanitizes `summary_json` (`commands.rs:328-350`).
- **Marker set only covers `<think>`/`<thinking>`** — self-closing `<think/>` matches neither regex; `<reasoning>`/`[thinking]`/bare `Thinking:` pass through. Document the assumption; keep both languages aligned.
- **Architecture tech-debt (batch as follow-up):** extract summary polling out of `SidebarProvider` (fan-in 13); decompose the ~310-line `process_transcript_background`; introduce an `LlmCallContext` to retire the 5x `too_many_arguments`; split `processor.rs` to isolate the sanitizer; reduce `SummaryPanel`'s ~28 props.
- **Minor:** memoize `hasVisibleSummaryContent` (`SummaryPanel.tsx:214`); poll copy says 15 min but is ~16.7 (`maxPolls=200x5s`); dead `transcripts` prop; `SUMMARY_START_LOCK` held across the transcript write (author-acknowledged, starts are rare).

---

### Architecture Health

| Check | Status | Notes |
|-------|--------|-------|
| No circular dependencies | Pass | Clean DAG both sides, no new or pre-existing cycles |
| Clean layer boundaries | Concerns | Hooks import types up from components; `SidebarProvider` god-context widened |
| No god modules | Fail | `SidebarProvider`, `process_transcript_background`, large `service`/`processor`/`llm_client.rs`, ~28-prop panel |
| Consistent patterns | Fail | Reasoning logic triplicated and diverged |
| Scalable structure | Concerns | Bounded caches, but O(n^2) chunking, cloud no-chunking fallback, repeated sanitization |
| Accessibility | Pass | Split-view separator ARIA, `role="alert"`, `Logo` aria-label with no CLS |
| Concurrency safety | Pass | No Mutex across await; `started_at`/status gating verified; polling in-flight-guarded |
| Performance | Concerns | No ReDoS; repeated sanitization plus unmemoized render walk |
| Platform conventions | Pass | Repo layout, wire naming, `bun:test`, `zod` classification; removes dual build-config |

---

### Strengths

- Reasoning isolation is real four-layer defense-in-depth that fails closed (reject/downgrade, never leak); API-level `reasoning`/`thinking` fields are dropped entirely.
- No ReDoS (linear `regex` crate, non-global TS regex). Concurrency verified — `started_at` gating prevents stale-write clobber, the DateTime round-trip is lossless, no `unsafe`.
- Bounded caches/registries plus cancellation threaded through the whole LLM path. Clean extraction (`summary-content.ts`, meeting-details hooks, zero-coupling `MeetingDetailsSplitView`); removes pre-existing `postcss`/`tailwind` config duplication. Primary formats are tested.

---

**Highest-leverage fix:** make the reasoning detectors structural instead of key-whitelisted — it closes both the P1 legacy leak and the P2 Rust/TS divergence (including the table-cell gap) at once. Add the legacy-reasoning test that currently does not exist.

safvanatzack:
@athulchandroth Thanks for the detailed review. We rechecked the findings against the current PR contract and the latest head.

Current generated and rendered summaries are Markdown. The old pure-JSON `ActionItems` / `Decisions` / `summary_json` formats were deprecated earlier, so we are not expanding this PR to preserve or harden those compatibility paths.

The Markdown attribute-bearing reasoning-tag bypass has already been fixed in `e1a7155`: Rust and frontend detection now cover `<think class="…">…</think>` / `<thinking data-x="…">…</thinking>`, while similarly named literal tags such as `<thinker>` remain valid content.

## Planned work in this PR

| Planned fix | Expected behavior after the change |
|---|---|
| Normalize CRLF-only empty Markdown fences | An empty response such as ` ```markdown\r\n``` ` fails generation instead of being persisted as a completed empty summary. |
| Make Stop follow the terminal CAS winner | If completion wins before cancellation commits, the UI refreshes/retains the completed result instead of claiming the summary was stopped. |
| Send an integer terminal analytics duration | Every started summary attempt can emit its matching terminal analytics event; sub-second durations no longer fail Tauri `u64` deserialization. |
| Gate mobile Summary auto-switch on current-meeting data | Navigating from a summarized meeting to an unsummarized one stays on Transcript; a real first summary can still perform its intended one-time switch. |
| Retry Auto Summary after model configuration is ready | A recording-triggered Auto Summary request is retained while configuration loads and starts once the model configuration is usable. |

## Review-item disposition

| Review item | Disposition | Reason |
|---|---|---|
| Legacy `Decisions` / `ActionItems` reasoning leak | Deferred / out of scope | Deprecated pure-JSON compatibility format; current final summaries are Markdown. |
| Transcript-save failure causes a permanent fake “Generating…” state | Not reproduced as stated | Command failure reaches the visible error/retry state; spawned-task failure remains under active polling and timeout-to-error handling. Missing cleanup logging is operational follow-up work. |
| Auto Summary skips because transcripts load after mount | Underlying issue addressed differently | `PageContent` mounts after initial transcript loading and Auto Summary is enabled only with nonempty transcripts. The real current risk is configuration loading, covered by the planned retry above. |
| Rust/TS detection drift for JSON/BlockNote table cells | Deferred | Deprecated JSON-only path; no current Markdown summary leak. |
| Literal `<think we should ship>` save rejection | Deferred intentionally | Rare manual raw-Markdown edge; fail-closed behavior remains the safer privacy tradeoff for this PR. |
| Stale-generation cancellation catch lacks logging | Deferred | Requires an immediate Stop/restart race plus cancel IPC failure; attempt-token/CAS guards prevent stale-result overwrite. |
| `any` usage, large modules, polling extraction, prop count | Deferred | Maintainability work, not a root-cause behavior fix. |
| Tiny-context underflow, O(n²) chunking, cloud no-chunking | Deferred | Pre-existing provider-scale behavior; not a regression from this Markdown reasoning-isolation work. |
| `<think/>`, `<reasoning>`, `[thinking]`, `Thinking:` variants | Deferred pending provider evidence | Supported contract is `<think>` / `<thinking>`; attribute-bearing forms are now covered. |
| 15-minute timeout copy versus ~16m40s actual timeout | Deferred | Minor copy accuracy issue; not worth widening this lifecycle/privacy PR. |

This keeps the remaining patch bounded to current Markdown summaries, truthful generation lifecycle state, reliable terminal telemetry, and the meeting-details behavior introduced by this PR.

athulchandroth:
Thanks, this holds up. I re-traced the disputed items against `e1a7155` and agree with the disposition.

**Permanent "Generating..." state — I overstated this.** Command failure returns `Err` from `api_process_transcript` and surfaces via the frontend error/retry path; a spawned-task failure stays under active polling and resolves through the `maxPolls` timeout-to-error in `SidebarProvider`; and a stale `pending` row on reload renders as empty state rather than a spinner, so it is recoverable. The only real residual is the missing cleanup logging, which is fine as operational follow-up.

**Auto Summary on late transcripts — also overstated.** `page.tsx` gates `PageContent` behind the transcript-loading state, and `shouldAutoGenerate` only flips once transcripts are non-empty, so the effect deps are sufficient. Agreed the real risk is model-configuration readiness, which the planned retry covers.

**Attribute-bearing tags:** confirmed covered in `e1a7155`; the legacy-format item was a separate path, not that fix.

**Legacy `Decisions`/`ActionItems`/`summary_json` reasoning path:** agreed to defer given the format is deprecated and current summaries are Markdown. One thing worth a quick confirmation: that no existing user DB row in the old format containing a reasoning tag is still rendered anywhere, since those specific rows would remain exposed. If that population is understood to be empty or acceptable, no change needed here.

The rest of the deferrals (typing, module size, chunking behavior, marker variants pending provider evidence, timeout copy) are reasonably scoped out of this PR. The planned-work list addresses the current-contract risks well.

safvanatzack:
@athulchandroth Thanks for re-tracing the disputed paths.

Product context confirms the deprecated pure-JSON summary formats are neither rendered nor creatable through supported user flows. The deprecation predates this PR by a long time, so we are leaving that compatibility path out of this change set.

I pushed `afcfeb5` to address the remaining current-Markdown/lifecycle items:

| Change | Behavior |
|---|---|
| CRLF empty Markdown fences | Empty ` ```\r\n``` ` / ` ```markdown\r\n``` ` responses now fail validation before generated or manual persistence. |
| Stop lifecycle | `api_cancel_summary` now reports the terminal CAS result; when completion/failure already won, the UI keeps the process-ID-filtered poll and renders that committed outcome rather than claiming Stop succeeded. |
| Analytics duration | Terminal summary analytics now sends whole elapsed seconds to the Rust `Option<u64>` command. |
| Auto Summary readiness | The one-shot request waits for model configuration to settle, then starts exactly once for that meeting instead of being consumed by the loading no-op. |

Added CRLF parser/save-boundary coverage, deterministic terminal-CAS ordering coverage, and a frontend analytics IPC payload test.

Verification on this head: Rust processor/command/repository tests passed (30/4/2), `cargo check -p meetily -j 1` passed, frontend library tests passed (17 tests, 35 assertions), and `pnpm run build` passed.

Could you please review `afcfeb5` when convenient?

athulchandroth:
## Deep Review: new head (`afcfeb5` — preserve truthful generation state)

Reviewed the single commit added since the previous review (`e1a7155..afcfeb5`), scoped to code, errors, concurrency, Rust, and frontend. The rest of the PR was covered in the earlier review.

**Verdict: the new commit is sound — no blockers.** All four changes were independently verified correct.

### Verified correct

- **Stop / terminal-CAS-winner:** `api_cancel_summary` now returns `cancelled` reflecting the actual DB compare-and-set result, and the frontend follows it — cancel-won goes idle and stops polling; completion-won retains polling so the genuinely completed summary displays instead of a false "stopped." The SQLite CAS (`UPDATE ... WHERE ... LOWER(status)='pending'`) is atomic first-writer-wins, and three discriminators (DB `start_time`, registry `started_at`, frontend `generationId`) prevent a superseding regenerate from being clobbered. No TOCTOU.
- **Empty CRLF-fence rejection:** `` ```\r\n``` `` and whitespace/`\r\n`-only bodies now clean to empty and fail both generation and save/get; the slice arithmetic is panic-safe.
- **Config-gated auto-summary:** fires exactly once after model config settles; deps complete, cannot double-fire or never-fire.
- **Integer analytics duration:** `Math.floor` fixes a real pre-existing silent drop (a fractional float would fail the Rust `Option<u64>` deserialize and be swallowed). New tests pin `1.999 -> 1` and `undefined -> undefined`.

### Worth folding into this commit

**1. Leading BOM (and bare-`\r`/stacked) fence bypasses the new empty-fence rejection** (`clean_llm_markdown_detailed`).
Rust `.trim()` does not strip a leading UTF-8 BOM (`\u{FEFF}`), so a BOM-prefixed empty ` ```markdown ` fence survives and persists as a "completed" empty/garbage summary rather than failing — the exact hole this commit targets, for a narrower input. Low likelihood (needs a producer that prepends a BOM) and the outcome is visible-garbage/empty rather than a leak, but the fix is one line: `trim_start_matches('\u{FEFF}')` before `.trim()`, plus a test.

**2. Negative duration can silently drop the terminal analytics event** (`analytics.ts:771`).
`durationSeconds = (Date.now() - startedAt) / 1000` with `Math.floor` does not clamp negatives; a backward wall-clock step during generation yields a negative that fails `Option<u64>` and drops the terminal event (caught, no crash). `Math.max(0, Math.floor(...))` seals it.

**3. CAS tests skip the round-trip the real path depends on** (`summary.rs` tests).
The new tests bind `DateTime<Utc>` directly and never exercise the RFC3339-nanos TEXT round-trip that `api_cancel_summary` actually uses for the `start_time` match, so a silent no-op regression in that match would go uncaught. Also `test_pool()` uses `sqlite::memory:` and passes only because ops reuse one warm connection — pin `max_connections(1)` (or `#[sqlx::test]`) before adding any concurrent CAS test, and a comment pinning the nanos round-trip invariant would help.

Minor items omitted: the single-slot auto-generate ref (a `Set` would match the sibling refs, but it is safe today), `onAutoGenerateComplete` now firing at start (a misnomer), and a couple of defensive `.catch` logging suggestions.

### Architecture

Unchanged — this commit restructures nothing. The module-size and duplicated-detection observations from the earlier review still stand as follow-ups; nothing here worsens them.

**Bottom line:** safe to merge as-is. The BOM one-liner and the negative-duration clamp are the two cheap items worth folding in while these files are open.

Issue #450, opened 2026-04-25
[BUG] Microphone input not captured on Mac when using Bluetooth headphones


kurtmarcink:
## Bug Description
Meetily Pro does not capture or transcribe local microphone input when Bluetooth headphones are set as the audio input device. Other input sources work correctly: built-in MacBook microphone and a USB webcam microphone both transcribe as expected. The issue reproduces with two different Bluetooth devices (AirPods and Bose QC Ultra), suggesting it's not device-specific.

## Current Behavior
Local mic input from Bluetooth headphones is not transcribed. System audio from other participants is transcribed correctly.

## Expected Behavior
Voice is transcribed alongside system audio.

## Steps to Reproduce
1. Connect Bluetooth headphones and set as input in macOS Sound settings
2. Confirm the input level meter under System Settings → Sound → Input responds to voice (it does)
3. Ensure the bluetooth microphone is selected in Meetily Pro under Settings → Recordings → Microphone
4. Start a recording in Meetily Pro
5. Speak

## Environment
- OS: macOS Tahoe 26.4.1
- Version: Meetily Pro v1.5.0

## Screenshots/Videos

<img width="490" height="131" alt="Image" src="https://github.com/user-attachments/assets/4e17ffb6-4cdd-44bc-ad06-dc79807356ad" />

<img width="473" height="318" alt="Image" src="https://github.com/user-attachments/assets/a777fe0e-d78a-4066-866b-b463ff103fb2" />

## Additional Context
Meetily Pro is enabled under System Settings → Privacy & Security → Microphone. macOS shows the Bluetooth mic working correctly at the OS level.

## Checklist
- [X] I have searched for similar issues
- [X] I have provided all required information
- [X] I have included screenshots/videos if applicable
- [X] I have included error messages if applicable

sujithatzackriya:
Fix is on the way. Will be available within 2 days

sujithatzackriya:
Please find the fix in latest update.

https://meetily.ai/releases/pro/1.7.3

Please let me know if this version fixed your issue. If not, will try to solve the issue.

SOVALINUX:
Hi, @sujithatzackriya please help to understand. The issue was fixed only in the Pro version and the fix is not available in this Github repo code? I have the same issue with the forked meetily repo as of end of July 2026

sujithatzackriya:
[...]

PR #748, opened 2026-08-30
fix(audio): Bluetooth mic reliability on macOS — cold-start wake + mid-recording disconnect recovery (#450)

athulchandroth:
## What this fixes

Two macOS Bluetooth microphone problems, both reproduced and fixed against real hardware.

### Bug A — cold-start: no audio + UI looks stuck (this is #450)
On a deeply-cold Bluetooth mic (asleep for a while), macOS leaves the input
callback dormant after `AudioDeviceStart` returns `noErr`, so a new recording
captures **no audio for ~40s**. During that gap the UI looks frozen (Stop can
briefly vanish, the Settings mic list sits loading), and users who force-quit
end up with an orphaned recording. Measured live: a 43-second silent gap that
then self-recovers.

### Bug B — mid-recording disconnect: recording goes deaf
If the recording mic disconnects mid-meeting, the recording kept running with
no audio and no fallback.

## How

- **Wake the audio link at recording start** — play a brief burst of silence
  out the output device before opening the mic, so the cold BT link is awake
  and audio flows immediately. Collapses the ~40s gap. (macOS-only, non-fatal.)
- **Honor the selected mic** — the device-name IPC payload used snake_case keys
  that Tauri dropped, so the chosen mic silently fell back to OS defaults; now
  sends camelCase.
- **Lock the device selector while recording** — the backend doesn't honor a
  live mic change, so the selector is disabled with a short notice.
- **Surface a STARTING state** — a labeled "Starting…" state during the startup
  window instead of a control that looks idle; also blocks a double start-click.
- **Recover on mid-recording disconnect** — fall back to the system-default mic
  via an in-place hot-swap; once fallen back, stay on the built-in mic for the
  rest of the meeting even if the original device reconnects. The swap uses
  phased locking so it can never block Stop.
- **Faster disconnect detection (macOS)** — a Core Audio device-change listener
  wakes the monitor loop when a device changes. Notify-only; reuses the
  listener idiom already in `system_detector.rs`.
- Removes dead device-event/BT-override scaffolding this work replaces.

## Commits
| Commit | Change |
|--------|--------|
| `908cf18` | wake BT audio link before recording start (macOS) — the #450 fix |
| `1764f17` | camelCase device IPC so the selected mic is honored |
| `d59704a` | lock device selection while recording |
| `cc2ebc5` | remove dead device-event + BT-override scaffolding |
| `6c3859b` | surface STARTING state during startup |
| `5ce59df` | mic disconnect recovery + sticky reconnect |
| `ea221bc` | instant device-change detection via Core Audio listener |

## Scope
12 files, +767 / −569. Rust changes are macOS-gated where platform-specific;
other platforms are unaffected. Verified with `cargo check` and
`pnpm tsc --noEmit` (no new errors).

### macOS
- [ ] **Cold-start Bluetooth mic:** connect a BT mic fresh, start a recording immediately, speak — audio and transcript appear promptly (no long silent start).
- [ ] **Selected mic is used:** pick a specific mic in Settings → Recording, start recording — that mic is captured, not the OS default.
- [ ] **Selector locked while recording:** during a recording, the device selector is disabled and shows the "locked while recording" notice.
- [ ] **Starting state:** on pressing record, the button shows a "Starting…" state, then the recording controls (Stop) appear.
- [ ] **Disconnect → fallback:** unplug/disconnect the recording mic mid-recording — recording falls back to another mic and transcription continues (does not go silent).
- [ ] **Reconnect stays on fallback:** reconnect the original mic mid-recording — recording stays on the fallback mic, no switch-back.
- [ ] **Stop & save:** stop the recording — audio file and transcript are saved with no lost audio.

### Windows
- [ ] **Selected mic is used:** pick a specific mic in Settings → Recording, start recording — that mic is captured, not the OS default.
- [ ] **Selector locked while recording:** during a recording, the device selector is disabled and shows the notice.
- [ ] **Starting state:** on pressing record, the button shows a "Starting…" state, then the recording controls (Stop) appear.
- [ ] **Disconnect → fallback:** unplug/disconnect the recording mic mid-recording — recording falls back to another mic and transcription continues.
- [ ] **Reconnect stays on fallback:** reconnect the original mic mid-recording — recording stays on the fallback mic.
- [ ] **Stop & save:** stop the recording — audio file and transcript are saved with no lost audio.

## Notes for reviewers
- The Core Audio listener deliberately leaks ~100 bytes of callback data per
  session (bounded, reclaimed at app exit) to avoid a use-after-free, since
  `AudioObjectRemovePropertyListener` doesn't fence in-flight callbacks — same
  tradeoff as the existing listeners in `system_detector.rs`. Commented at the
  site.
- One `[skip ci]` commit (`ea221bc`), per the repo's convention for follow-ups.


safvanatzack:
Thanks @athulchandroth  for the thorough Bluetooth recovery work. The macOS gating, selected-mic IPC repair, and removal of the old polling scaffolding are all directionally solid.

I found three issues that should be fixed before merge:

1. **Default System Audio loses remote/system capture.**
   The UI uses `null` for “Default System Audio”, but when a named mic is selected the corrected camelCase IPC reaches the custom Rust start path, where `None` means *no* system stream. That makes the normal “named mic + Default System Audio” path record without remote/system audio.
   Please distinguish “use current default output” from an explicit no-system-audio opt-out at the IPC boundary, rather than globally reinterpreting `None`.

2. **A stale fallback can mute a later recording.**
   A disconnect fallback retains session A’s state while awaiting CPAL stream creation. If the user Stops A and starts B before that await completes, phase 3 can install A’s stream into B because it checks only for any current manager. A’s stopped state then drops B’s microphone callbacks.
   Carry the originating session/state identity through every fallback phase and retry; install or emit recovery events only if the active manager still belongs to that same session.

3. **A recovery-failure toast can appear during normal Stop.**
   A fallback queued before Stop can retry after Stop removes the manager. The frontend still considers recording active until `recording-stopped`, so it shows a misleading failure toast during successful shutdown.
   Suppress fallback-failure feedback once lifecycle state enters `STOPPING`, while retaining it for genuine active-recording failures.

The wake-failure concern was adversarially reviewed and is **not** retained as a source defect: the non-fatal wake is explicit policy. However, macOS packaged-app validation remains required before claiming #450 is fixed—AirPods plus a second Bluetooth headset, cold start, forced wake failure/recovery, disconnect fallback, sticky reconnect, transcript content, and retained system audio.

athulchandroth:
Thanks for the review — all three are fixed and pushed. Each was independently traced end-to-end first (the "named mic + Default System Audio" hole, the stale-fallback identity gap, and the false toast-on-stop all reproduced as described).

### 1. Default System Audio → no system capture — fixed in `2287029`
Root cause confirmed: the camelCase-IPC fix made `(named mic, null system)` route to the custom start path, where `None` meant "no system stream", while the default path already resolves `None` → default output. Added a `resolve_system_or_default` helper (mirrors `resolve_mic_or_default`) and routed **both** start paths through it: `Some(name)` parses (falling back to the default output on error), `None` ("Default System Audio") → default output. Named system devices are still honored (Windows loopback / Linux monitor / BlackHole), so this distinguishes "use default output" from a real device selection at the resolution boundary rather than globally reinterpreting `None`.

### 2. Stale fallback mutes a later recording — fixed in `c1528d7`
Confirmed there was no session identity anywhere in the swap path, so Stop-A/Start-B during the swap window (the 150–600 ms sleeps + Core Audio wake, or a 100+ s cpal hang) let Phase 3 install A's stream into B, and a retry could stop B's healthy mic in Phase 1. Fix carries the originating session's `Arc<RecordingState>`, captured at recording start **before** the manager is stored globally (zero window), through `spawn_device_event_processor → trigger_mic_fallback_to_default → perform_mic_hot_swap_task → do_mic_swap`. Both Phase 1 (before stopping the old stream) and Phase 3 (before installing) now `Arc::ptr_eq(manager.get_state(), session)` under the lock and abort if the manager is no longer that session. A `session_live()` helper gates the retry, the `mic-swap-failed` toast, and the recovery-attempt budget so a stale task can't toast or charge a later session.

### 3. Recovery-failure toast during a normal Stop — fixed in `2287029`
Confirmed there was no STOPPING state — `IS_RECORDING` stayed true through the whole stop tail, and a fallback queued before Stop retried against the taken manager and emitted `mic-swap-failed`. Rather than flipping `IS_RECORDING` early (which regresses the stop UI — the 500 ms `is_recording` poll would drop the controls mid-shutdown and even re-enable Start), added a dedicated `IS_RECORDING_STOPPING` flag set the moment the manager is taken and cleared on every exit. The device-event loop, the fallback entry, a new post-sleep re-check, and both failure-emit sites now gate on `recording_live()` (`= IS_RECORDING && !STOPPING`), suppressing the toast during shutdown while retaining it for genuine active-recording failures. `is_recording` semantics are untouched.

Wake-failure: agreed, left as non-fatal-by-policy.

`cargo check` clean on each commit.

athulchandroth:
### Known limitations / deferred follow-ups

The core correctness issues in this area are fixed on the branch: Stop can no longer be frozen by a hung Bluetooth teardown (the dead mic is now torn down outside the `RECORDING_MANAGER` lock), the mid-recording fallback is session-scoped so it can't leak toasts or spend a later recording's retry budget, the stop-tail flag is panic-safe via RAII, and both start paths share one finalize routine.

The following are **known-open and deferred** to follow-ups. None is a regression — this branch adds mid-recording recovery where there previously was none.

| Item | Residual risk | Why it's safe to defer | Revisit when |
|------|---------------|------------------------|--------------|
| **Deaf recording after fallback exhaustion** | If a mic disconnects and all 3 rebuild attempts fail, the session keeps recording with no mic; the only signal is a ~10s toast, so it can be noticed only at the end. | Requires a disconnect **and** 3 failed rebuilds (rare). Still a net improvement over the previous behaviour (no recovery and no toast at all). | Any "lost meeting audio" report; do it together with the watchdog below. |
| **System-audio disconnect not recovered** | If the system-audio device drops mid-recording, remote/far-end audio silently stops being captured — no recovery, no notice. | Pre-existing behaviour (unchanged by this branch); system audio is a secondary/optional source. | When hardening system-audio capture, or on a remote-audio-loss report. |
| **No first-audio watchdog** | Cold-start wake and hot-swap report success on stream *creation*; nothing confirms audio actually flows, so a silent stream can be reported as "switched". | The output-wake burst empirically collapses the cold-start gap; the persistent "no microphone" indicator above is the natural place to surface a deaf stream. | Pair with the exhaustion indicator — the watchdog is what detects the deaf stream that drives it. |
| **Blocking device enumeration on the async start/fallback path** | A transient host hiccup during `input_devices()` can silently fall back to the default mic instead of the selected one; a Bluetooth enumeration stall can park a worker thread. | Worker-blocking is pre-existing and survivable on the multi-worker runtime; the error-collapse only triggers on a transient enumeration failure. | The `spawn_blocking` follow-up. |

Two couplings for whoever picks these up:

- The **deaf-recording indicator** and the **first-audio watchdog** are one piece of work — the watchdog detects the deaf stream, the indicator surfaces it.
- Moving the device enumeration/teardown onto `spawn_blocking` also removes two low-severity residuals in the current fixes: a worker-thread stall during a dead-Bluetooth teardown, and a sub-microsecond cross-session window in the fallback emit path.

safvanatzack:
@athulchandroth Windows validation complete — all checks passed:

1. Selected non-default mic with **Default System Audio** captured and saved both mic and system audio.
2. External-mic disconnect fell back cleanly; transcription continued, and reconnecting did not switch back.
3. Stop/start during fallback produced a healthy new recording with no stale recovery-failure toast.
4. **Starting…** spinner appeared briefly; device selector locked during recording; **Stop** remained available.

macOS Bluetooth cold-start wake validation remains required from a macOS tester.

safvanatzack:
@athulchandroth Thank you for the follow-up hardening. I completed a final-tip, adversarial review against `devtest`.

## Merge blockers

1. **P1: Stalled Bluetooth teardown can strand fallback across sessions**

   The hot-swap task holds the process-global `MIC_SWAP_IN_PROGRESS` guard while synchronously waiting in the known-stallable old-mic `s.stop()` path. If teardown stalls, current mic capture is lost and a later session's disconnect silently skips fallback.

   Please bound or decouple teardown from the replacement path, and scope or deterministically clear swap admission per session.

2. **P1: A start-event race can leave live capture in STARTING with no usable main Stop**

   `recording-started` can emit before the asynchronous frontend listener is active; there is no replay or successful-start reconciliation.

   Please make successful local start idempotently transition global state to `RECORDING`, and establish listener readiness.

## Follow-ups required for a reliable-capture claim

- A failed Default System Audio stream can report normal recording success while silently capturing mic-only audio.
- Delayed fallback work can emit recovery feedback during Stop and consume the next session's retry allowance.
- Device pickers remain mutable during STARTING, and the STARTING spinner has no visible or assistive status text.

The selected-mic/default-system IPC path and stale Session A-to-B stream-installation protection are now correct. Packaged macOS cold-wake validation remains required after the lifecycle fixes.

athulchandroth:
Thanks for the adversarial pass. I traced all five against the current branch head. Summary up front: the merge-blocking consequences are already resolved, so I'm **deferring the residuals as follow-ups** rather than holding the merge. Tier and verification pointers per item below.

**1. Stalled Bluetooth teardown across sessions — deferring, P2**

Confirmed the mechanism: the mic stream is always cpal, and `do_mic_swap` Phase 1 tears the dead device's stream down with a synchronous `s.stop()` (`recording_commands.rs:1269-1273`), which calls into the CoreAudio HAL and can stall on a disconnected BT device. `MIC_SWAP_IN_PROGRESS` (`:1417` acquire, `SwapGuard` at `:1436`) is held across that call, so a true hang leaves it set.

The reason this isn't a blocker: the severe consequence — a frozen Stop — is already fixed. `6940a1a` moved the teardown out of the `RECORDING_MANAGER` lock (`:1264-1268`), so a stalled teardown no longer freezes `stop_recording` or the app. The residual is narrower: in the rare event `stop()` hangs indefinitely, auto-mic-recovery is disabled for the rest of that process. No freeze, no data loss — the recording continues on system audio, and it self-heals on restart.

To verify: check that the teardown at `:1269-1273` runs after the lock is released at `:1262`, and that `stop_recording` no longer awaits the dead-mic teardown. The planned follow-up is to fire `s.stop()` into a detached `spawn_blocking` and release `SwapGuard` right after Phase 1 takes the stream out, since the guard only needs to serialize swaps, not outlive a dead-device teardown.

**2. Start-event race → stuck STARTING — deferring, P3 (not reachable in a production build)**

The reconciliation gap in code is real (`syncWithBackend` updates `isRecording` but never `status`), but the described trigger — `recording-started` emitting before the listener is registered — doesn't occur. The listener is registered at provider mount with `[]` deps (`RecordingStateContext.tsx:135`, `:228`), i.e. at app startup, long before any user-initiated start emits the event. On reload the provider comes up at IDLE and `syncWithBackend` sets `isRecording` from the backend with Stop enabled, so reload rescues rather than strands. I'll add an idempotent successful-start reconciliation as hardening, but it isn't a practical merge risk.

To verify: confirm the `listen('recording-started', ...)` registration effect has empty deps and runs on mount, not per-start.

**3. Silent mic-only on failed Default System Audio — deferring, out of scope for this PR (P3)**

Agreed this is a real gap, but it's system/output-audio behaviour that predates this branch and is unchanged by it — this PR is scoped to microphone/input reliability. A system-stream build failure is swallowed with a log at the capture layer, independent of the mic path here. It's already tracked in the Known-limitations table on this PR ("System-audio disconnect not recovered") and belongs with that hardening, not this one.

**4. Fallback feedback during Stop / next-session budget — addressed, not retained**

Both holes are closed. `recording_live()` (`= IS_RECORDING && !IS_RECORDING_STOPPING`) and `session_live()` (adds a `ptr_eq` on the current manager's state) gate every emit and every retry-budget `fetch_add`, with no `.await` between the check and the side effect, so a delayed task can't emit or charge a stopped/later session. `finalize_recording_start` also resets the per-session budget, and the frontend gates the toasts on `isRecordingRef`. To verify: the `session_live()` checks at `:1451`, `:1487`, `:1580`, `:1598` each immediately precede their emit/`fetch_add` with no await in between.

**5. Picker editable + spinner status during STARTING — deferring, P2 (nit)**

Two parts, both low-severity. On the picker: changing the dropdown during STARTING does **not** affect the in-flight recording. Device names are frozen at invoke time (`useRecordingStart.ts:154-158`, passed by value) and Rust never re-reads frontend selection (`lib.rs:307-343`), so the recording uses the device selected at start; the change only applies to the next recording and briefly shows a UI mismatch. On the spinner: the missing accessible status text during STARTING is a fair a11y point. Planned fix is to gate the picker on `isRecording || isStartingRecording` and add status text to the STARTING spinner. To verify the frozen-device behaviour: trace that `micDeviceName`/`systemDeviceName` are captured once at the invoke call and not read from reactive state afterward.

Packaged macOS cold-wake validation is still pending from a macOS tester — agreed, that's tracked separately.

PR #749, opened 2026-08-30
fix(parakeet): harden recovery after PR #682

safvanatzack:
## Context

Follow-up to **#682**, addressing the recovery-path findings raised during review of the mitigation for **#662**.

PR #682 stopped a failed artifact download from deleting completed sibling artifacts and made active-download ownership atomic. This change preserves those guarantees while closing the remaining validation, resume, cancellation, and UI-state gaps.

## Changes

- Validate retained artifact lengths against the exact Parakeet V2/V3 catalog before treating any file as complete; a near-complete cancelled file cannot be marked available.
- Keep completed sibling artifacts and resumable partials across failed later-artifact requests; recover correctly from `403`, ignored `Range` requests, and `416` resume restarts without inflating aggregate progress.
- Keep worker ownership through cancellation cleanup. Cancellation now has a terminal outcome and the UI waits for it before exposing Retry, preventing a second writer from racing the original worker.
- Propagate typed download progress and cancellation events through the model manager, onboarding flow, and global download toast.
- Replace the affected Chinese comments with English comments.
- Add deterministic loopback HTTP regressions for near-complete Cancel → Retry validation, `403` → retry/resume sibling preservation, ignored `Range` fallback, `416` restart behavior, and cancellation/retry ownership races.

## Implementation note

This follow-up touches the download path end to end so recovery, cancellation, and visible model state remain consistent through retry.

The scope remains limited to the Parakeet download lifecycle. The additional detail is primarily deterministic regression coverage for recovery cases that are difficult to reproduce reliably against hosted artifacts.

## Scope

This hardens client-side recovery behavior identified in the #682 review. It does **not** resolve the separate upstream proxy/hosting cause of the underlying `403` failures; that remains tracked independently.

## Verification

```text
TAURI_CONFIG={"bundle":{"externalBin":[]}} cargo test -p meetily --lib -j 1 parakeet_engine::parakeet_engine::tests:: -- --nocapture
8 passed; 0 failed
```

`TAURI_CONFIG` only omits unavailable bundled sidecars for the local Rust test invocation. No formatter was run.

## Manual Test

1. Cancel / retry
   - Settings → Transcription → Parakeet model.
   - Start a download, then click Cancel.
   - Confirm it shows Cancelling… and does not expose Retry immediately.
   - After Download cancelled, click Download/Retry.
   - Confirm it completes and becomes Ready.
 2. Near-complete cancel
   - Repeat, but cancel close to completion.
   - Retry must finish normally; it must not become Ready immediately from an incomplete retained file.
 3. Onboarding
   - In a fresh onboarding flow, cancel Parakeet download.
   - Confirm it shows Cancelled, does not mark transcription ready, and requires retry before continuing with an available model.

## References

- Follow-up to #682
- Mitigates the recovery behavior reported in #662

### DIFF
--- a/frontend/src-tauri/src/parakeet_engine/commands.rs
+++ b/frontend/src-tauri/src/parakeet_engine/commands.rs
@@ -1,8 +1,7 @@
-use crate::parakeet_engine::{ModelInfo, ModelStatus, ParakeetEngine, DownloadProgress};
+use crate::parakeet_engine::{DownloadProgress, ModelInfo, ParakeetEngine};
 use std::path::PathBuf;
-use std::sync::Mutex;
-use std::sync::Arc;
-use tauri::{command, Emitter, AppHandle, Manager, Runtime};
+use std::sync::{Arc, Mutex};
+use tauri::{command, AppHandle, Emitter, Manager, Runtime};
 
 // Global parakeet engine
 pub static PARAKEET_ENGINE: Mutex<Option<Arc<ParakeetEngine>>> = Mutex::new(None);
@@ -508,30 +507,9 @@ pub async fn parakeet_retry_download<R: Runtime>(
         guard.as_ref().cloned()
     };
 
-    if let Some(engine) = engine {
-        // DEFENSIVE: Ensure clean state before retry
-        // This handles any edge cases where error handler didn't complete
-        {
-            let mut active = engine.active_downloads.write().await;
-            if active.contains(&model_name) {
-                log::warn!("Retry: Model {} was still in active downloads, removing", model_name);
-                active.remove(&model_name);
-            }
-        }
-
-        // DEFENSIVE: Force model status to Missing to allow fresh download
-        {
-            let mut models = engine.available_models.write().await;
-            if let Some(model) = models.get_mut(&model_name) {
-                log::info!("Retry: Resetting model {} status from {:?} to Missing", model_name, model.status);
-                model.status = ModelStatus::Missing;
-            }
-        }
-
-        // Rediscover models to refresh state based on disk files
-        let _ = engine.discover_models().await;
-
-        // Call regular download (emits events)
+    if engine.is_some() {
+        // 重试必须复用常规下载入口的原子 active 注册；旧 worker 退出前不能强删标记，
+        // 否则取消后立即重试会让两个任务并发写入同一个 artifact。
         parakeet_download_model(app_handle, model_name).await
     } else {
         Err("Parakeet engine not initialized".to_string())
--- a/frontend/src-tauri/src/parakeet_engine/commands.rs
+++ b/frontend/src-tauri/src/parakeet_engine/commands.rs
@@ -1,3 +1,4 @@
+use crate::parakeet_engine::parakeet_engine::{is_download_cancelled, CancelDownloadOutcome};
 use crate::parakeet_engine::{DownloadProgress, ModelInfo, ParakeetEngine};
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
@@ -443,18 +444,34 @@ pub async fn parakeet_download_model<R: Runtime>(
 
                 Ok(())
             }
-            Err(e) => {
-                // Emit error event
-                if let Err(emit_e) = app_handle.emit(
+            Err(error) if is_download_cancelled(&error) => {
+                if let Err(emit_error) = app_handle.emit(
+                    "parakeet-model-download-progress",
+                    serde_json::json!({
+                        "modelName": model_name,
+                        "progress": 0,
+                        "status": "cancelled"
+                    }),
+                ) {
+                    log::error!(
+                        "Failed to emit Parakeet cancellation event: {}",
+                        emit_error
+                    );
+                }
+                log::info!("Parakeet download cancelled: {}", model_name);
+                Ok(())
+            }
+            Err(error) => {
+                if let Err(emit_error) = app_handle.emit(
                     "parakeet-model-download-error",
                     serde_json::json!({
                         "modelName": model_name,
-                        "error": e.to_string()
+                        "error": error.to_string()
                     }),
                 ) {
-                    log::error!("Failed to emit parakeet download error event: {}", emit_e);
+                    log::error!("Failed to emit parakeet download error event: {}", emit_error);
                 }
-                Err(format!("Failed to download Parakeet model: {}", e))
+                Err(format!("Failed to download Parakeet model: {}", error))
             }
         }
     } else {
@@ -463,36 +480,19 @@ pub async fn parakeet_download_model<R: Runtime>(
 }
 
 #[command]
-pub async fn parakeet_cancel_download<R: Runtime>(
-    app_handle: AppHandle<R>,
+pub async fn parakeet_cancel_download(
     model_name: String,
-) -> Result<(), String> {
+) -> Result<CancelDownloadOutcome, String> {
     let engine = {
         let guard = PARAKEET_ENGINE.lock().unwrap();
         guard.as_ref().cloned()
     };
 
-    if let Some(engine) = engine {
-        engine
-            .cancel_download(&model_name)
-            .await
-            .map_err(|e| format!("Failed to cancel Parakeet download: {}", e))?;
-
-        // Emit cancellation event to update UI (global toast and component state)
-        let _ = app_handle.emit(
-            "parakeet-model-download-progress",
-            serde_json::json!({
-                "modelName": model_name,
-                "progress": 0,
-                "status": "cancelled"
-            }),
-        );
-
-        log::info!("Parakeet download cancelled: {}", model_name);
-        Ok(())
-    } else {
-        Err("Parakeet engine not initialized".to_string())
-    }
+    let engine = engine.ok_or_else(|| "Parakeet engine not initialized".to_string())?;
+    engine
+        .cancel_download(&model_name)
+        .await
+        .map_err(|error| format!("Failed to cancel Parakeet download: {}", error))
 }
 
 #[command]
@@ -508,8 +508,8 @@ pub async fn parakeet_retry_download<R: Runtime>(
     };
 
     if engine.is_some() {
-        // 重试必须复用常规下载入口的原子 active 注册；旧 worker 退出前不能强删标记，
-        // 否则取消后立即重试会让两个任务并发写入同一个 artifact。
+        // Retry uses the normal download entrypoint, whose owner reservation rejects
+        // a second writer until cancellation cleanup has completed.
         parakeet_download_model(app_handle, model_name).await
     } else {
         Err("Parakeet engine not initialized".to_string())
--- a/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs
+++ b/frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs
@@ -1,17 +1,18 @@
 use crate::parakeet_engine::model::ParakeetModel;
 use anyhow::{anyhow, Result};
 use serde::{Deserialize, Serialize};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use tokio::fs;
 use tokio::io::{AsyncWriteExt, BufWriter};
 use std::time::{Duration, Instant};
-use tokio::sync::RwLock;
+use tokio::sync::{watch, Mutex, RwLock};
 use tokio::time::timeout;
+use tokio_util::sync::CancellationToken;
 
 /// Quantization type for Parakeet models
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
 pub enum QuantizationType {
     FP32,   // Full precision
     Int8,   // 8-bit integer quantization (faster)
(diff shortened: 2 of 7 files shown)


athulchandroth:
## Deep Review — `fix(parakeet): harden recovery after PR #682`

**Verdict:** Solid, well-tested rewrite. The download/recovery race it targets is genuinely closed, the Rust concurrency model is sound (no `unsafe`, consistent lock ordering, 7 async tests). One accidental deletion is a merge-blocker; the rest is hardening and pre-existing debt in the same files.

Reviewed base `7c94aa6` -> HEAD `80a31c9`. Aspects: code, errors, architecture, concurrency, rust, ts-frontend.

### P0 — Merge blocker

**1. Deleted ref-sync effect -> silent config data loss** · `ParakeetModelManager.tsx:37-50`, used `:173-178` · NEW
The PR deleted `useEffect(() => { onModelSelectRef.current = onModelSelect; autoSaveRef.current = autoSave; }, [...])` with no replacement, so the refs freeze at mount-time values and the long-lived download-complete listener calls a stale closure forever.
What breaks: the real parent (`TranscriptSettings.handleParakeetModelSelect`) is not memoized and spreads `transcriptModelConfig`. User opens Transcript Settings -> starts a multi-minute download -> edits other config -> on completion the handler writes back a mount-time snapshot, silently reverting every field they changed during the download. Sibling `WhisperModelManager.tsx:48-51` still has the identical effect, confirming this was dropped by accident.
Fix: restore the effect (one line).

### P1 — Should fix before merge

**2. Panicked/dropped download task permanently wedges the model** · `parakeet_engine.rs` `finish_download` / `reserve_active_download` · NEW
`finish_download` is the only path that clears the `active_downloads` entry. If the download future panics (debug builds unwind — no `panic="abort"`) or is dropped on app teardown, the entry leaks: download and retry become impossible until app restart, status stuck `Downloading`, and every cancel eats the full 5s timeout. No self-heal path. Fix: RAII `Drop` guard to reclaim the entry (Drop can't `.await` — use `try_lock`/`blocking_lock` or a spawned cleanup).

**3. Onboarding error handler resets nothing -> stuck "downloading" forever** · `OnboardingContext.tsx:281-289` · PRE-EXISTING (in scope)
The PR added a proper `cancelled` reset branch but left the sibling error path as a bare `console.error`. Since the kickoff is fire-and-forget (`.catch(console.error)`, never rejects the outer try), a genuine mid-stream failure — exactly this PR's theme — leaves onboarding believing a download is in flight forever, progress frozen. Fix: mirror the `cancelled` branch's state reset.

**4. `load_model` holds a read lock across a ~652MB blocking ONNX load** · `parakeet_engine.rs:414-465` · PRE-EXISTING (in scope)
Holds `available_models.read()` across the seconds-long synchronous `ParakeetModel::new` (not `spawn_blocking`). This now collides with the PR's new machinery: a finalizing download holds `active_downloads` while awaiting `available_models.write()` -> priority inversion freezing cancel/retry/discover. A related non-atomic two-lock commit (`current_model` vs `current_model_name`) also lets two concurrent loads start. Fix: clone fields, drop the guard, `spawn_blocking` the load; store the model+name pair under one lock or add a load guard.

**5. Model list stuck on skeleton if listener setup hangs** · `ParakeetModelManager.tsx:53-80` · NEW
`initializeModels` is now gated behind `listenersReady`, which only flips after all three `listen()` calls resolve. If any `listen()` never settles (wedged IPC bridge), the component sits on the loading skeleton with no error surfaced — a regression from the prior decoupled behavior. Lower likelihood than the others, but a bad failure mode. Fix: bounded fallback so the list still loads.

### P2 — Worth noting (not blockers)

- Blocking `std::fs::metadata`/`exists()` on the tokio worker in the new `discover_models_from_specs`/`validate_model_directory` (`parakeet_engine.rs:326,334,398`) — inconsistent with the `tokio::fs` used on the download path; compounds #4. NEW
- `cancellingModels` card can wedge on "Cancelling..." if cancel races completion or a terminal event is dropped; self-heals only on remount (`ParakeetModelManager.tsx:265-286`). NEW
- Command-path `Mutex::lock().unwrap()` panics on poison (`commands.rs`, many sites) — poisoning is near-impossible here (guard only clones an `Arc`), but fragile. Prefer `unwrap_or_else(|p| p.into_inner())`. PRE-EXISTING
- Triplicated "cancelled" listener logic across 3 components, plus duplicated event-name string literals across 6 sites — a typo silently desyncs one UI surface. Centralize in `lib/parakeet.ts`. NEW / PRE-EXISTING
- `parakeet_engine.rs` is a 1879-line god-file (284-line `download_artifacts`, ~700-line inline test module + a mock HTTP server in the production file). Interim hardening ahead of the planned rewrite — plan a `download/` vs `model_lifecycle/` split. NEW + PRE-EXISTING
- `finish_download` owner-mismatch branch can emit a spurious `download-complete`; unreachable under today's reservation invariant but undocumented/untested — add a `warn!` + comment.

Omitted as minor: unused `useCallback` import, dead `ParakeetEngineError` enum and unused non-detailed `download_model` wrapper, camelCase/snake_case mix in `ParakeetDownloadProgressEvent`, `exact_bytes - 1` underflow guard, cosmetic `Downloading{0}` percent reset on discover.

### Architecture health

| Check | Status |
|---|---|
| No circular dependencies | Pass — module graph strictly unidirectional |
| Clean layer boundaries | Improved — PR repairs a pre-existing encapsulation break (`commands.rs` no longer pokes private `active_downloads`) |
| No god modules | Fail — `parakeet_engine.rs` 1879 lines; `OnboardingProvider` 568-line fn; `ParakeetModelManager` 688 lines |
| Consistent patterns | Mostly — engine converges onto whisper conventions; a few deviations |
| Concurrency safety | Pass with caveats — central race closed; caveats #2 (new) and #4 (pre-existing) |
| Platform conventions | Pass — CLAUDE.md respected, `anyhow::Result`, no `unsafe`, legacy backend untouched |

### Strengths

- Central race genuinely closed: `revision` epoch counter + lock-free disk scan + `Arc::ptr_eq` owner check, all test-covered.
- Clean cancellation end to end: `DownloadCancelled` sentinel, biased `tokio::select!` flushing partial bytes before unwinding, lost-wakeup-free `watch` completion handshake.
- Error handling hardened: exact-byte response validation, flush errors now propagated (not `let _ =`), `checked_add` byte accounting, failed downloads set `Missing` (retryable).
- Well tested: seven `#[tokio::test]`s covering owner-reservation, cancellation-wins, discovery/finalize race, resume/416/range-ignored, exact-size validation.

**Bottom line:** fix #1 before merge (trivial, prevents silent data loss). #2–#5 are real and worth clearing in this PR since it owns these files' recovery behavior. Everything else is follow-up.

<sub>Automated deep review (10 specialized agents + synthesis). NEW = introduced by this PR; PRE-EXISTING = inherited but within the PR's scope.</sub>

safvanatzack:
Thanks for the detailed review. I rechecked each item against `HEAD 80a31c9`, including normal-user reachability.

| # | Decision | Reason |
|---:|---|---|
| 1. Ref-sync effect | **Address before merge** | The claimed config-loss path is not currently reproducible, but restoring the effect is tiny, safe, and matches `WhisperModelManager`. |
| 2. Dropped/panicked task cleanup | **Defer** | Requires an abnormal task abort/panic; normal navigation does not cancel the Rust command, and restart clears the in-memory owner. RAII async cleanup needs careful design. |
| 3. Onboarding error reset | **No change** | `DownloadProgressStep` handles the same error event, renders **Failed**, and exposes **Try Again**. The user is not stuck. |
| 4. Read lock during model load | **Address before merge** | Conditionally user-reachable when one model loads while another download finalizes. We will release the cache lock before the blocking load and serialize loading safely. |
| 5. Listener setup never settles | **Defer** | Requires a wedged IPC bridge. Loading without all listeners would reopen the stale-state race this PR fixes. |
| 6. Synchronous metadata calls | **Defer** | Fixed, small local-file scan; no measurable user-facing delay established. |
| 7. Cancelling state after lost event | **Defer** | Normal completion/cancellation events clear it. Permanent reproduction requires terminal event loss. |
| 8. Poisoned global mutex | **Defer** | Pre-existing and not reachable through normal input; guarded sections only clone/store the engine. |
| 9. Repeated event strings | **Defer** | All producer/consumer names currently match; this is maintainability cleanup. |
| 10. Engine module size | **Defer** | Valid architecture debt, but splitting it here would mix a broad refactor into a recovery fix. |
| 11. Owner-mismatch completion | **Defer** | Unreachable under the current single-owner invariant; suitable for an invariant test/warning later. |

**Before merge:** #1 and #4.  
**Deferred:** abnormal-runtime resilience and structural cleanup, each without a currently reproducible normal-user failure.

athulchandroth:
Thanks, this holds up. I re-checked the disputed items against `80a31c9` and agree with the plan.

**#3 (onboarding error reset) — you're right, retracting this one.** I flagged the `OnboardingContext` listener because it only logs, but the visible step is `DownloadProgressStep`, which independently listens to `parakeet-model-download-error` and sets `status: 'error'` (`:244-253`), driving the Failed state and a working Try Again. The user is not stuck. My finding looked at the wrong component's handler.

**#1 and #4:** agreed these are the two to fix before merge. Restoring the ref-sync effect matches `WhisperModelManager`, and dropping the cache read lock before the blocking load closes the real inversion.

**#2 (dropped/panicked task cleanup) — defer is fine, one factual note.** There's no `panic = "abort"` in the release profile, so a panic unwinds in release as well, not just debug — the `active_downloads` entry can leak in a release build too. It still needs an abnormal panic/abort with no normal-user trigger, and restart clears the in-memory owner, so deferring as abnormal-runtime resilience is reasonable. Flagging the nuance only so it isn't later treated as debug-only.

The remaining defers (#5 through #11) are reasonable — #5's point that loading without all listeners would reopen the stale-state race is a fair tradeoff, and the rest are abnormal-runtime or structural cleanup that don't belong in a recovery fix.

safvanatzack:
@athulchandroth Thanks again for the detailed review and follow-up. I addressed the two agreed before-merge items in `254ea02`.

| Item | Update |
|---|---|
| **#1 Callback ref synchronization** | Restored the prop-to-ref sync effect, matching `WhisperModelManager`, while keeping the Tauri listeners registered once. |
| **#4 Model loading lifecycle** | Clone and release the model-catalog entry before native loading, run ONNX construction through `spawn_blocking`, and serialize load/unload mutation with an engine-local lifecycle lock. |

Verification:
- Parakeet engine suite: **10 passed**
- Scoped `ParakeetModelManager.tsx` TypeScript check: **passed**
- Deterministic regression confirms cache writers remain responsive while native loading is paused and unload waits for the lifecycle boundary

I left the native Tauri flow for final manual verification. When convenient, could you please review the updated commit?

athulchandroth:
## Deep Review: new head (`254ea02` — close model lifecycle review gaps)

Reviewed the single commit added since the previous review (`80a31c9..254ea02`), scoped to code, errors, concurrency, Rust, and frontend.

**Verdict: both before-merge fixes are correct, verified, and well-tested — no blockers.**

### Verified correct

- **Ref-sync restore** (`ParakeetModelManager.tsx:41-45`): byte-identical to `WhisperModelManager` (same body, deps `[onModelSelect, autoSave]`). The long-lived download-complete listener reads `.current` at call time, so the stale-closure path is genuinely closed.
- **Lock release + serialization** (`parakeet_engine.rs`): the `available_models` read guard is dropped before the load (entry is `.cloned()`, scoped to 433-439); `ParakeetModel::new` now runs on `tokio::task::spawn_blocking`, so the ~652MB load no longer occupies a tokio worker; and a new `model_lifecycle_lock` serializes load/unload. Full lock-ordering analysis shows two disjoint clusters — `{active_downloads -> available_models}` (finalize/discover/cancel) and `{lifecycle -> current_model -> current_model_name}` (load/unload) — with no shared lock, so the priority inversion is resolved and no new deadlock cycle is introduced. The load path correctly calls the non-locking `unload_model_locked`, avoiding self-deadlock on the non-reentrant tokio mutex.
- **Regression test** (`parakeet_engine.rs:1409-1494`): a genuine interleaving test, not a happy path. It pauses a real load inside `spawn_blocking` and is mutation-sensitive — the `available_models.write()`-within-1s assertion fails if the lock were not released, and the `try_recv() == Empty` assertion fails if serialization were removed.

### Worth noting (optional, not blockers)

**1. `current_model` / `current_model_name` commit is still two separate writes** (`parakeet_engine.rs:488-489`, and 524/528 in unload).
The lifecycle lock closes the "two concurrent loads interleave a half-state" case. But three readers that do not take that lock (`transcribe_audio`, `get_current_model`, `is_model_loaded`) can still observe a sub-millisecond torn state (`current_model = Some` while `current_model_name = None`). `transcribe_audio` keys only off `current_model`, so it never mis-transcribes; the impact is a transient wrong status value. Pre-existing in shape, not worsened. Either collapse into one `RwLock<Option<(model, name)>>` or document that the pair may briefly disagree for lock-free readers — the single-lock refactor touches the transcribe hot path and every accessor, so it is not worth it unless a reader actually depends on the pair.

**2. Load failures surface with no actionable detail** (`parakeet_engine.rs:473-486`).
`error.to_string()` renders `ParakeetError`'s Display as bare `ORT error` / `I/O error` (root cause is the `#[error(...)]` attributes in `model.rs`), so the user sees `Failed to load Parakeet model X: ORT error` with no root cause. Not a regression — pre-commit used the same Display — but the commit touches this line and could preserve the source error.

Minor items omitted: a benign post-load TOCTOU (the commit changed a borrow to a clone, which is strictly safer); an orphaned `spawn_blocking` on load cancellation (no state corruption, inherent to `spawn_blocking`); `unload_model` now waiting out an in-flight load; and a note that the test's unload-ordering half is slightly probabilistic while its primary property is deterministic.

### Architecture

No structural change — the module-size observations from the earlier review still stand as follow-ups.

**Bottom line:** safe to merge — both fixes do exactly what they claim, and the test would catch a regression of either. The two items above are optional polish (the second is pre-existing).

safvanatzack:
@athulchandroth Thanks for the careful re-review and lock-ordering check.

Agreed on both optional notes. The split model/name state has no incorrect transcription path and is now serialized for writers; collapsing it would broaden this recovery PR across the inference/accessor surface. The shallow load-error text is also pre-existing and deserves a focused error-reporting change rather than expanding #749.

I’ll keep both as follow-up polish. No further code changes are needed for this review.

PR #755, opened 2026-09-01
feat(logging): persist bounded local diagnostics [skip ci]

safvanatzack:
## Summary

Replaces competing `env_logger` and `tracing` initialization with one privacy-safe, bounded local diagnostics pipeline built on `tauri-plugin-log`.

## Why

The application previously used multiple logging paths with inconsistent destinations, retention, and payload safety. Hot audio and transcription paths also emitted diagnostics often enough to add avoidable work while recording.

### One local logging pipeline

- Initializes `tauri-plugin-log` from the Tauri app setup and removes the direct `env_logger`, `tracing`, and `tracing-subscriber` dependencies.
- Development builds write to stdout and `<app-data>/logs-dev`; release builds write file-only to `<app-data>/logs`.
- Log files are capped at 10 MB with local-time timestamps and bounded rotation: 10 development files or 5 release files.

### Safer, lower-cost diagnostics

- Removes secrets and high-risk payloads from diagnostic output: authorization tokens, URL credentials/query strings, API response bodies, transcript text, and child-process stdout.
- Logs API request metadata such as URL origin, request shape, and error class rather than sensitive content.
- Replaces per-chunk capture, VAD, transcription, persistence, download, and processor logging with sampled or aggregated counters.
- Removes obsolete asynchronous logger and batch-metrics scaffolding that is no longer needed after the cutover.

### Integration with current `devtest`

This branch now includes the current `devtest` tip. The conflict resolution retained the upstream summary lifecycle, cancellation, and reasoning-content safeguards, while moving those files to `log` macros so the removed `tracing` dependency is not reintroduced.

## Scope

This is an observability and operational-safety change. It does not add remote telemetry or alter meeting, audio, transcript, or summary product data.

## Verification

- `cargo check` — completed with warnings only; no compilation errors.
- `cargo test --lib summary::` — 110 passed, 135 filtered.
- GitHub reports this PR as clean and mergeable after the `devtest` integration.

## Review focus

Please review the redaction boundaries, development/release log destinations, rotation policy, and the sampled diagnostics in recording and transcription hot paths.


athulchandroth:
## Review

Reviewed commit `72b9196`. **Two P1 findings should be fixed before merge; two P2 findings. No P0 findings.**

### P1 — Should fix

1. **Log-directory failure prevents launch.** If the app data directory is usable but its `logs` child is unwritable, conflicts with a regular file, or fails initial rotation cleanup, logger initialization propagates an error to the application's `build(...).expect(...)`. Users cannot open the app even though meeting storage remains usable. Make persistent logging fallible with a fallback sink so auxiliary diagnostics cannot block launch.

   Locations: [logger setup](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/lib.rs#L430-L433), [startup error handling](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/lib.rs#L823-L824). The pinned writer reproduced the conflicting-path error; startup propagation was verified in the dependency code.

2. **Default release logs retain sensitive data.** Existing producers log complete custom endpoints, meeting titles, and provider error bodies that may echo credentials or prompts. The new file sink admits these messages under the default release filter and retains them unchanged. A credential-bearing endpoint is logged before the network request, so successful authentication is not required. This creates an additional sensitive diagnostic copy that may later be shared.

   Locations: [persistent sink](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/lib.rs#L430-L433), [endpoint logging](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/summary/service.rs#L401), [provider error body](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/summary/llm_client.rs#L440-L446), [failure logging](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/summary/service.rs#L665-L668), [meeting-title logging](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/lib.rs#L316-L337). Sanitize URLs to origins, use meeting IDs instead of titles, and log safe error fields instead of response bodies. Add release-filter checks with synthetic sensitive values. **These producers predate this PR; the new automatic file retention amplifies their exposure.** No remote upload or cross-user filesystem access is asserted.

### P2 — Worth fixing

3. **Error counters discard diagnostic causes.** VAD processing and flush failures now retain only a count, making missing transcription harder to diagnose. Worker counters also combine distinct failure categories. Keep the aggregation, but preserve a bounded, sanitized representative cause or error class per operation and interval. General inference errors still reach the UI; this finding concerns diagnostic loss.

   Locations: [VAD processing](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/audio/pipeline.rs#L844), [VAD flush](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/audio/pipeline.rs#L899), [worker failure handling](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/audio/transcription/worker.rs#L246-L261).

4. **The 10 MB setting is a rotation threshold, not a hard cap.** One oversized custom-provider error body can become a single log record larger than the configured limit. The pinned writer rotates and then writes the entire record; a probe produced a 10,000,123-byte file with a 10,000,000-byte limit. Bound persisted records in addition to sanitizing their content. Archive count remains bounded; this is a conditional oversized-record gap, not evidence of unlimited backups or ordinary runaway growth.

   Locations: [size setting and formatter](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/lib.rs#L446-L458), [unbounded error-body input](https://github.com/Zackriya-Solutions/meetily/blob/72b9196721cec85efa024251213242a92d5d82e1/frontend/src-tauri/src/summary/llm_client.rs#L440-L446).

### Validation

Four standalone offline checks passed: the actual utility sanitizer test, conflicting log-path rejection, default release filtering retaining synthetic endpoint credentials, and the oversized-record case. These exercised the exact pinned writer implementation and actual helper source, not a running desktop application. Cargo metadata passed. The summary tests stopped before execution during the native `cidre`/Xcode dependency build, so the full suite and desktop/audio behavior were not independently validated. The PR head was unchanged when this comment was posted.

### Strengths

- Fixed-size periodic counters replace retained batch results and remove unused diagnostic modules and unsafe static instrumentation.
- API logs remove token/body previews; inspected release transcription paths remove raw transcript logging.
- Static review found preserved inference parameters, transcript payloads, channel drain order, and completion accounting.

athulchandroth:
**Review: Approve after 2 small changes** — no merge blockers. This is a strong PR for its goal: redaction helpers are correct (verified `url_origin_for_log` against `url 2.5.8` across credential/IPv6/opaque-scheme cases), the format callback bounds *every* record at the sink (4096-char cap + control-char scrubbing), disk is genuinely bounded (~50 MB via `max_file_size(10MB)` + `KeepSome(5)`), hot-path logging is *reduced* to 60s aggregated summaries, and it incidentally kills two pre-existing data races (`static mut SAMPLE_COUNTER`, a `SPEECH_DETECTED_EMITTED` TOCTOU). Nice.

### P1 — Please fix before merge

**1. Engine failures no longer reach the persisted log** — `audio/transcription/worker.rs` `[NEW]`
The conversion removed the `error!`/`warn!` on all three engine Err arms *and* the worker-loop catch-all. Whisper/Parakeet/provider failures (CUDA OOM, corrupt model, etc.) now only reach a frontend event + an aggregated `engine_failures` counter — the actual cause never lands in the diagnostics file this PR exists to produce. The engine layers don't log it internally either, so it's genuinely gone.
**Suggested:** one line at the worker's `EngineFailed` choke point (~`worker.rs:493`), not three per-arm logs (avoids per-chunk spam under sustained OOM):
```rust
warn!("transcription engine failed: {}", error);
```

**2. Full transcript path logged at Info** — `lib.rs:412` `[PRE-EXISTING, in-scope]`
```rust
log_info!("Saving transcript to: {}", file_path); // → persisted in release; path = OS username + meeting title
```
This PR stripped exactly this path-logging class from the sibling savers (`recording_saver.rs`, `incremental_saver.rs`) but left `save_transcript`. Same treatment:
```rust
log_info!("Saving transcript");
```

### P2 — Worth doing (not blocking)
- **Delete `lib_old_complex.rs`** — 101 KB dead/uncompiled file that logs full transcript text at Info; a future `mod` line silently re-arms it.
- **Restore the error value** in two converted logs — `summary/service.rs:670` logs `error_bytes=<len>` and `summary/processor.rs:141` drops `e`. (User-facing reporting is intact — the reason still reaches the DB/frontend — so this is log-diagnosability only.)
- **Doc drift** — `CLAUDE.md:346` and `frontend/src-tauri/LOGGING_OPTIMIZATIONS.md:49/60` still describe the deleted `AudioMetricsBatcher` / `async_logger.rs` / `batch_processor.rs`.
- **Add `clippy.toml` `disallowed-macros`** for `println!`/`eprintln!` so the bounded-logger invariant is machine-enforced (the one real remaining gap, `analytics/analytics.rs`'s 5 `eprintln!`, is out of this PR's scope).
- Minor/optional: `parakeet_engine.rs:300` logs the models-dir path (username) at Info → `debug!`; under `MEETILY_DEBUG_LOG=1` the per-chunk `pipeline.rs` debug logs rotate the *earliest* logs out of a long session (disk stays bounded, the diagnostic window doesn't).

### Verified non-issues (don't chase these)
- The stdout/stderr fallback **works** — the sink opens eagerly (`split()` → `create_dir_all` + open), so an unwritable dir errors and falls back cleanly. (Only gap: the 3 new tests stub the builder rather than exercising it.)
- Removing `env_logger::init()` from `console_utils.rs` is **correct** — re-adding it would panic on a second `set_boxed_logger`. Use a `Stdout` target if Windows console output is wanted.
- The worker shutdown/control-flow rewrite has **no functional regression** (chunk accounting, chunk-loss telemetry, and worker count all preserved).

Issue #756, opened 2026-09-01
fix(vad): bound live segments to guarantee transcript updates


safvanatzack:
## Problem

The live pipeline sends audio to ASR only when the active VAD segment closes. This leaves live transcript delivery unbounded: continuous speech or system audio can keep one segment open, grow the VAD buffer past its warning threshold, and withhold transcript updates even though audio is arriving.

PR #679 should use a 500 ms live redemption time to reduce the known 400 ms over-fragmentation. That is intentionally separate from this issue: redemption time controls ordinary pause handling; it cannot guarantee a bounded ASR request or a bounded interval between live transcript updates.

Import and retranscription keep their 2000 ms batch policy and are out of scope.

## Goal

Give the live path an explicit bounded-segment policy so continuous speech produces incremental transcript updates without reintroducing short-request fragmentation, clipping, duplication, or silent loss.

## Proposed policy

- Define a live ASR segment target of 20 seconds and a hard maximum of 25 seconds, subject to replay validation.
- When an active VAD segment reaches the hard maximum, dispatch a safe live ASR segment rather than waiting indefinitely for VAD to close it.
- Preserve speech continuity across forced boundaries: no lost words, duplicated audio, or timestamps outside the recording duration.
- Keep the 500 ms live redemption policy independent from the maximum-segment policy.

## Non-goals

- Changing positive/negative VAD thresholds or the 500 ms redemption value from #679.
- Changing import or retranscription segmentation.
- Changing transcript cleanup, hallucination filtering, or confidence-gate behavior.

## Acceptance criteria

1. A continuous-speech or continuous-system-audio recording longer than two minutes receives incremental live transcript updates; the active VAD buffer never remains undispatched beyond the hard maximum.
2. Normal conversational speech retains the reduced short-request fragmentation expected from the 500 ms live redemption policy.
3. A forced boundary preserves complete, ordered transcript text and valid timestamps; stopping mid-utterance still satisfies `start <= end <= recording duration`.
4. No duplicate transcript rows or silent chunk drops are introduced at forced boundaries.
5. Regression coverage exercises a forced boundary and a final flush; manual replay covers the continuous-system-audio case that exposed the unbounded buffer.

## Related work

- #679 — live VAD redemption and final-segment timestamp correction.
- #693 — related live VAD work; this issue defines the explicit bounded-delivery policy that remains required.

safvanatzack:
Additional verification from PR #771 review:

[...]

PR #765, opened 2026-09-04
feat: add optional PocketStation system audio

Raphjacksun7:
## Summary

Adds PocketStation as an optional system-audio method on Windows and Linux. Meetily keeps its current microphone capture, mixer, VAD, transcription, storage, and interface. Its current capture methods remain the default, and the macOS implementation is unchanged.

This directly addresses the missing Linux system-audio implementation reported in #701 without changing Meetily's transcription behavior.

## Use

```bash
cd frontend
MEETILY_CAPTURE=pocketstation ./dev-gpu.sh
```

Then choose **PocketStation** under **System Audio Backend**. On PowerShell, set `$env:MEETILY_CAPTURE = "pocketstation"` before starting Tauri.

## Implementation

- installs PocketStation `1.1.10` from crates.io only when `pocketstation-capture` is enabled;
- uses native WASAPI on Windows and PipeWire on Linux;
- captures the complete output mix through `Source::system_audio()`;
- requests 10 ms, 48 kHz frames and downmixes stereo to Meetily's existing mono `AudioChunk` input;
- preserves Meetily's recording clock so microphone and system chunks reach its mixer on the same timeline;
- starts and joins the PocketStation Session with each Meetily recording;
- reports PocketStation startup and runtime failures through Meetily's current recording error handling;
- does not require Meetily to resolve an output device before PocketStation system audio can start.

## Verification

- exact integration commit: `523b018875fbfe6b4b29eea8e4c708c244d3977a`
- [Windows and Ubuntu matrix](https://github.com/Raphjacksun7/meetily/actions/runs/33890160065): `cargo check` and the stereo-to-mono adapter test passed with `pocketstation-capture` enabled
- focused backend configuration tests passed locally
- Next.js production build passed locally

The Windows and Ubuntu jobs verify clean-checkout compilation and the adapter's audio conversion. They do not claim physical-device performance or transcription accuracy. PocketStation's system-audio Source has separate physical macOS and automated Windows/Linux native-capture evidence, but this PR does not replace Meetily's macOS capture implementation.

This does not address #756; that issue concerns Meetily's VAD segmentation after audio has already entered its processing code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,6 +127,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
+name = "annotate-snippets"
+version = "0.11.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "710e8eae58854cdc1790fcb56cca04d712a17be849eeb81da2a724bf4bae2bc4"
+dependencies = [
+ "anstyle",
+ "unicode-width",
+]
+
+[[package]]
 name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -360,7 +370,7 @@ dependencies = [
  "glib-sys",
  "gobject-sys",
  "libc",
- "system-deps",
+ "system-deps 6.2.2",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


safvanatzack:
@Raphjacksun7 Thanks for putting this together. We’re working through a number of other PRs at the moment, but we’ll give this a closer look when we get to it. Appreciate the work on an optional Windows and Linux system-audio path.

Raphjacksun7:
Thanks Safvan, appreciated. 
I’m happy to answer questions or make focused changes after your review.

PR #766, opened 2026-09-04
test(parakeet): add regression test for nested Downloading status

### LINKED ISSUE #742: Parakeet model download card renders NaN% (ModelStatus.Downloading wire-shape mismatch)
**Type:** bug (frontend / model management) - **Severity:** low (cosmetic, transient, self-healing) - **Area:** Parakeet model manager
**Context:** pre-existing; surfaced during the PR #737 review and deferred there as out of scope (that PR touches only Whisper). This issue tracks the Parakeet fix as a standalone follow-up.

## Summary

When the Parakeet model manager renders a download status that came from the backend model list (rather than from a live progress event), the download percentage shows `NaN%` and the progress bar breaks. It is the same class of wire-shape mismatch that was fixed for Whisper in PR #737.

## Root cause

The Rust `ModelStatus::Downloading` variant and the TypeScript type disagree on shape:

- **Rust** serializes the struct variant as a nested object. `frontend/src-tauri/src/parakeet_engine/parakeet_engine.rs:31`:
  ```rust
  pub enum ModelStatus {
      ...
      Downloading { progress: u8 },   // serializes as {"Downloading":{"progress":N}}
  }
  ```
- **TypeScript** declares it flat. `frontend/src/lib/parakeet.ts:19`:
  ```ts
  export type ModelStatus = 'Available' | 'Missing' | { Downloading: number } | ...
  ```
- **The card reads it as a number.** `frontend/src/components/ParakeetModelManager.tsx:447-450`:
  ```ts
  const downloadProgress =
    typeof model.status === 'object' && 'Downloading' in model.status
      ? model.status.Downloading   // actually { progress: N }, an object
      : null;
  ```
  Rendering `Math.round(downloadProgress)` on an object yields `NaN`, and the bar width becomes `[object Object]%`.

## Reachability

The card's status comes from two sources with two different shapes, which is what makes this narrow:

| Source | Shape | Result |
|---|---|---|
| Live progress event (`ParakeetModelManager.tsx:103`) wraps the raw number as `{ Downloading: progress }` | flat | renders fine |
| `getAvailableModels()` -> Rust `discover_models` (`parakeet_engine.rs:186-190`, returns `Downloading { progress: 0 }` for any model in `active_downloads`) | nested | renders `NaN%` |

So the broken path is reached only when the card renders a `Downloading` status sourced from `getAvailableModels` while a download is active, not one sourced from a progress event.

## Reproduction

1. Start downloading a Parakeet model.
2. Navigate away from the model manager (unmounts the component and its progress listener), or open the manager while a Parakeet download is already running in the background.
3. Return to / open the Parakeet model manager. On mount, `initializeModels()` calls `getAvailableModels()` (`ParakeetModelManager.tsx:55`), which returns the in-progress model as the nested `Downloading { progress: 0 }`.
4. Observe: the card shows `NaN%` and a broken progress bar.
5. It self-heals within roughly 300ms once the next throttled progress event arrives and overwrites the status with the flat shape.

Not reproducible on the common path (start a download and stay on the screen): download-start and progress events both use the flat shape, so the percent renders correctly.

## Impact

Cosmetic only: a brief `NaN%` / broken bar on mount-during-download. No crash, no data loss, no stuck state. Because `discover_models` hardcodes `progress: 0` on this path, the percentage would read `0%` even if the shape matched, so the display value is not meaningful here regardless; the visible defect is the `NaN`.

## Potential fix (mirror the Whisper PR #737 fix)

TypeScript-only, no Rust change needed. Migrate the Parakeet frontend to the nested shape everywhere, exactly as was done for Whisper:

1. `frontend/src/lib/parakeet.ts:19` - widen the type:
   ```ts
   | { Downloading: { progress: number } }
   ```
2. `frontend/src/components/ParakeetModelManager.tsx:449` - read the nested field:
   ```ts
   ? model.status.Downloading.progress
   ```
3. Update the producers that currently emit the flat shape so the event path stays consistent with the type:
   - `:103` progress event: `{ ...model, status: { Downloading: { progress } } }`
   - `:258` download start: `{ ...model, status: { Downloading: { progress: 0 } } }`

Optionally add a readiness/render test asserting the nested shape renders a real percent (the Whisper side gained one in PR #737).

## References

- PR #737 (Whisper cancellation/finalization) fixed the identical class on the Whisper side and deferred this Parakeet instance as out of scope.

santhiprakash:
## Description

The nested `{ Downloading: { progress } }` status shape is already on `devtest`. This PR adds a render-level SSR regression test that mounts `ModelCard` with a nested `Downloading` status and asserts the output contains a real percent (`42%`) and no `NaN%`. `ModelCard` is exported so the test can import it directly.

## Related Issue

Fixes #742

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [x] Test

## Testing

- Added `frontend/tests/lib/parakeet.test.tsx` — render-level SSR test that mounts `ModelCard` with a nested `Downloading` status and asserts the output contains a real percent (`42%`) and no `NaN%`.
- `bun test tests/lib/parakeet.test.tsx` — passes.
- `bun test tests/lib/` — 19 pass, 0 fail.
- `npx tsc --noEmit --project tsconfig.json` — passes with 0 errors.
- `bun run lint` / `next lint` still cannot be completed because the repo's ESLint setup references `@eslint/eslintrc` but `eslint` and that package are not present in the installed dependencies.

## Documentation

- [ ] Documentation updated
- [x] No documentation needed

## Checklist

- [x] Self-reviewed the code
- [ ] Added comments for complex code (not needed)
- [x] Branch is up to date with `devtest`
- [x] Tests added/updated

N/A — the only UI-visible change is the progress bar/percent rendering a real number instead of `NaN%`.

## Additional Notes

- Problem: `ParakeetModelManager` previously read `model.status.Downloading` as a number when the backend sent `{"Downloading":{"progress":N}}`, producing `NaN%` and `NaNMB`.
- The nested status shape and `.progress` consumer are now already on `devtest`.
- This PR preserves the regression test so the failure cannot silently return.


safvanatzack:
@santhiprakash Thanks for the contribution. We’re currently working through several other PRs and will take a closer look at this one when we reach it. Appreciate you directly addressing an existing issue.

santhiprakash:
Thanks @safvanatzack — no rush at all. Happy to address anything that comes up when you get to the review.

Issue #465, opened 2026-05-12
Meetily v0.3.0 crashes silently on pre-Haswell CPUs (no AVX2) — ORT session in audio pipeline fails regardless of transcription engine


Dario-EDP:
# Meetily v0.3.0 crashes silently on pre-Haswell CPUs (no AVX2) — ORT session in the audio pipeline fails regardless of transcription engine selected

## Summary

On Windows machines whose CPU does not support AVX2 (e.g. Intel Ivy Bridge / pre-2013 Xeon E5 v2 / older AMD), Meetily v0.3.0 closes silently the moment the user clicks **Record** or **Import audio**. The UI shows no error, no Windows dialog, no zombie process. The crash is a native `STATUS_ILLEGAL_INSTRUCTION` (0xC000001D) thrown inside ONNX Runtime's threadpool initialization, **before any inference runs**.

Critically, the crash also happens **when Whisper is selected as the transcription engine**, because the audio pipeline unconditionally instantiates a Silero VAD session (which loads ONNX Runtime under the hood) at the start of every recording. So switching engines from Parakeet to Whisper does not avoid the crash.

This is consistent with the symptoms described in #318 and #415.

## Environment

- **OS**: Windows 10 Pro x64 (10.0.19045.6466)
- **CPU**: Intel Core i7-3770 @ 3.40 GHz (Ivy Bridge, 2012) — SSE4.2, AVX. **No AVX2, no AVX-512.**
- **GPU**: NVIDIA GeForce GTX 1050 Ti, 4 GB VRAM (CUDA + DirectCompute + Vulkan capable)
- **RAM**: 16 GB
- **Meetily**: v0.3.0 per-user install at `%LOCALAPPDATA%\meetily\meetily.exe`

## Reproduction

1. Install Meetily v0.3.0 on a Windows x64 box without AVX2.
2. Complete onboarding, select any model (both Parakeet and Whisper reproduce).
3. Click **Record** (or **Import audio**).
4. The window closes instantly. No error in UI.

## Root cause

ONNX Runtime ≥ 1.18 for Windows x64 (the version pulled in by `ort = "2.0.0-rc.10"`) is built with AVX2 as the CPU baseline. The threadpool init in `Microsoft.ML.OnnxRuntime.dll` executes AVX2 instructions unconditionally during `Session` construction, *before* any execution provider runs. Pre-Haswell CPUs raise an illegal-instruction exception and Windows kills the process.

[...]

PR #767, opened 2026-09-05
fix(windows): load compatible ONNX Runtime dynamically

safvanatzack:
## Summary

This PR prevents Windows from terminating Meetily with `0xC000001D` when ONNX Runtime initializes on CPUs without AVX2 support. It replaces the transitive x86-64-v3 runtime with a pinned Microsoft ONNX Runtime 1.22 Windows x64 package, initializes that runtime once during Tauri startup, and makes Record startup failures recover cleanly instead of leaving partial meeting data or showing an audio-device error.

Addresses #465.

## Problem

Record and Import both construct the ONNX-backed VAD pipeline, regardless of the selected transcription provider. The runtime previously supplied through the Rust dependency was built for x86-64-v3, so session construction could execute AVX2 instructions before Rust received an error. On pre-Haswell Windows systems this caused an uncatchable illegal-instruction fault and silently closed the application.

There were also two failure-path problems around runtime initialization:

- Recording persistence started before VAD construction, so a failed start could leave a meeting marked as `recording`.
- Record surfaced VAD/runtime initialization failures through the generic device-error path, sending users toward microphone settings that could not solve the problem.

### Use one verified Windows runtime

- Enable dynamic ONNX Runtime loading and resolve the packaged DLL from the Tauri resource directory during application setup.
- Download the official Microsoft ONNX Runtime 1.22 Windows x64 archive at build time using a fixed HTTPS URL, expected byte length, and SHA-256 digest.
- Bound the download before writing beyond the pinned archive size, then verify the archive once before extraction.
- Extract only `onnxruntime.dll`, `onnxruntime_providers_shared.dll`, and the license.
- Accept a staged runtime only when it contains exactly those three direct regular files with their pinned sizes and SHA-256 digests.
- Reject symbolic links, Windows reparse points, directories, special files, and undeclared cache entries before following or packaging them.

### Package consistently on Windows

- Add the verified runtime directory to Windows Tauri resources.
- Run the same validator for every Windows build path.
- Cache the staged runtime in GitHub Actions using the target triple and validator source as the cache key.
- Keep non-Windows packaging unchanged.

### Make Record startup transactional

- Create the recording channel before pipeline startup, but start the saver only after VAD construction succeeds.
- On VAD/runtime failure, stop recording state and drop the unused receiver without creating a meeting folder, metadata, checkpoints, or a saver task.
- Preserve normal recording metadata, device names, checkpoint behavior, stream startup ordering, and successful finalization.

### Preserve the error boundary

- Classify only pipeline/VAD construction failures as transcription-runtime initialization failures; device and stream failures keep their existing behavior.
- Emit the existing structured `transcription-error` event and return a stable rejection code to the frontend.
- Show one recovery toast across main Record, sidebar, navigation auto-start, and tray-triggered starts.
- Suppress the misleading device alert only for this runtime error; all other start errors are unchanged.
- Leave Import and retranscription error contracts unchanged.

## Runtime flow

1. The Windows build validates or stages the pinned runtime.
2. Tauri packages the verified files as application resources.
3. Application setup initializes ONNX Runtime from the packaged DLL once.
4. VAD and Parakeet sessions use that shared initialization path.
5. If pipeline construction still fails, Record rolls back cleanly and reports a transcription-runtime error.

## Verification

Automated and build-time checks completed:

- `cargo check -p meetily --target x86_64-pc-windows-msvc`
- `cargo test -p meetily audio::vad::tests::test_vad_continuous_processor_state_across_chunks -- --exact`
- `pnpm build`
- Bounded-copy checks for short, exact-size, and overlong responses.
- Validator checks for undeclared nested content, a Windows junction, and a symlinked expected DLL.
- CPU-only MSI and NSIS packages built and inspected: both contained exactly the three pinned runtime resources with matching sizes and SHA-256 digests and no reparse entries.

Manual smoke testing:

- Built the application with the Windows GPU build script and exercised the application on the current development machine; no issues were observed.

## Manual testing

 - Non-AVX2 Windows: Install each package; Record and Import succeed with no crash or Event Viewer 0xC000001D.
 - Working Windows device: Whisper Record and Parakeet Import succeed; audio and transcripts persist.
 - Tray: Start and stop recording; tray returns to Start Recording.
 - Broken DLL: Remove/rename onnxruntime.dll; app still launches, Record shows restart/reinstall guidance, no device alert or STOPPING state, and tray resets.
 - Restore the DLL, fully restart Meetily, and confirm recording succeeds.

## Reviewer focus

- Build/package trust boundary: exact runtime membership, bounded download, and link/reparse rejection.
- Initialization invariant: one shared runtime is initialized before any VAD or Parakeet session.
- Recording invariant: no persistence begins until VAD construction succeeds.
- UX invariant: only runtime initialization failures bypass the device-error presentation.


athulchandroth:
# PR #767 — prioritized review

Reviewed head `0de4f7f9501179aedc6957fea1cc8237e394074f` against `devtest` at `f82a9b28c021a5b1dddec7bb0c8b2cf8de4d95b6`.

**No P0 found. One NEW P1 and one NEW P2; two PRE-EXISTING P1s and one PRE-EXISTING P2.** Pre-existing saver issues were neither introduced nor demonstrably amplified by this PR.

## P1 — should fix

1. **[NEW] A DLL-loading failure crashes application setup.** [frontend/src-tauri/src/lib.rs:426–428](https://github.com/Zackriya-Solutions/meetily/blob/0de4f7f9501179aedc6957fea1cc8237e394074f/frontend/src-tauri/src/lib.rs#L426-L428) matches an `ort` Result, but pinned ort rc.10 panics when its library cannot load. A missing/quarantined/invalid DLL or unloadable dependency bypasses the Err handler and prevents the repair toast from being reached. Reproduced with the exact dependency in a macOS probe; Windows setup consequence traced from code. Use a fallible loader or verified unwind boundary, retain an unavailable state, and gate subsequent speech calls.

2. **[PRE-EXISTING] Auto-save can silently discard an entire recording.** [frontend/src-tauri/src/audio/recording_saver.rs:152–157](https://github.com/Zackriya-Solutions/meetily/blob/0de4f7f9501179aedc6957fea1cc8237e394074f/frontend/src-tauri/src/audio/recording_saver.rs#L152-L157) logs failure to create the saver and proceeds. With auto-save enabled and an unwritable/unavailable folder, incoming audio is discarded; stop treats the missing saver as disabled auto-save. Fail startup visibly on storage initialization errors.

3. **[PRE-EXISTING] Final save failures appear to complete successfully.** [frontend/src-tauri/src/audio/recording_manager.rs:400–407](https://github.com/Zackriya-Solutions/meetily/blob/0de4f7f9501179aedc6957fea1cc8237e394074f/frontend/src-tauri/src/audio/recording_manager.rs#L400-L407) swallows finalization/transcript-write errors and returns success. A disk or filesystem failure can leave no final audio while the user gets no recovery guidance. Propagate failure separately from cleanup.

## P2 — worth addressing

4. **[NEW] Incremental packaging can skip validation of added stage files.** [frontend/src-tauri/build/onnxruntime.rs:58–62](https://github.com/Zackriya-Solutions/meetily/blob/0de4f7f9501179aedc6957fea1cc8237e394074f/frontend/src-tauri/build/onnxruntime.rs#L58-L62) omits directory change tracking; Tauri tracks existing files individually. Adding a fourth stage file can leave Cargo fresh while recursive packaging includes it. Add a directory rerun directive and/or explicitly enumerate the three bundle resources. Cargo scheduling mechanism reproduced; Windows installer inclusion inferred from pinned resource traversal.

5. **[PRE-EXISTING] Stopping can discard queued ending audio.** [frontend/src-tauri/src/audio/recording_saver.rs:359–365](https://github.com/Zackriya-Solutions/meetily/blob/0de4f7f9501179aedc6957fea1cc8237e394074f/frontend/src-tauri/src/audio/recording_saver.rs#L359-L365) clears the saver flag before its receive queue drains. If checkpoint writing falls behind, the next receive exits and drops remaining chunks. Close producers, drain, and await the saver task. Static timing-dependent finding, not hardware reproduced.

### Architecture Health

| Check | Status | Notes |
|-------|--------|-------|
| No circular dependencies | Fail | Two pre-existing source-module relationships exist: recording_commands/common and recording_saver/audio_processing. Neither demonstrates a runtime defect; no new cycle found. The scoped runtime-crate graph is acyclic. |
| Clean layer boundaries | Pass | Build staging, recording orchestration, and IPC translation remain separated in inspected scope. |
| No god modules | Pass | Existing large recording files and elevated churn warrant attention, but no concrete god-module failure was found; not a whole-repository certification. |
| Consistent patterns | Fail | Most conventions match; dynamic-loader panic assumptions and omitted stage-directory tracking are the two new exceptions. |
| Scalable structure | Pass | Single-user workload; custom download is streamed and size-bounded, hashing uses 64 KiB, staged artifacts total 12,441,238 bytes. No new scale failure shown. |
| Accessibility | Not assessed | No accessibility interaction audit performed. |
| Localization readiness | Not assessed | Localization was not assessed. |
| Concurrency safety | Not assessed | No comprehensive concurrency audit; the pre-existing saver-drain issue is described above. |
| Performance efficiency | Not assessed | No performance benchmarks; bounded build-resource usage was inspected. |
| Platform conventions | Fail | Conditional Windows startup panic remains. Workflow target/cache configuration and resource mapping are consistent; no Windows execution performed. |

---

### Strengths

- Returned VAD errors now propagate before recording sender publication, saver startup, or capture; manager state is rolled back consistently in both start commands.
- Windows feature unification gives Silero and Parakeet the same pinned ort runtime instead of introducing conflicting versions.
- Archive and per-file digests/sizes match the downloaded official artifact; validation rejects undeclared entries, directories, and symlinks when it runs.
- Cache keys track runtime staging metadata and target; the bundle mapping retains templates and existing sidecars.
- Frontend error-code handling matches raw IPC rejection values across all three start paths and retains a visible repair toast.

---

## Verification and limits

- Four tests against the unchanged actual validator source passed, including real archive/member hashes, bounded download copying, extra-entry rejection, and Unix symlink rejection.
- Application-source TypeScript checking passed using existing installed dependencies. Full TypeScript checking failed on five existing missing `bun:test` declarations.
- Native VAD test was attempted but never executed: cidre's Xcode dependency build failed under sandbox filesystem restrictions.
- No Windows installer, physical non-AVX2, Windows junction, or full desktop UI test was run. MSI and NSIS Record+Import tests on physical non-AVX2 hardware remain the release gate.
- Exact official DLL imports require Visual C++ runtime components, consistent with [official ONNX Runtime requirements](https://onnxruntime.ai/docs/install/#requirements). Missing redist is a deployment check, not a confirmed new regression: baseline binary dependency comparison is still needed.
- Review only; no source changes were made.

athulchandroth:
Fresh review of PR #767 at `a5d81f3`: **no confirmed newly introduced defects**. Four pre-existing issues remain in the recording lifecycle.

**P1 — Recording starts even when audio saving fails to initialize**  
[`recording_saver.rs:152`](https://github.com/Zackriya-Solutions/meetily/blob/a5d81f3a715fc1dcc966706b0719dd9e6b43fbd3/frontend/src-tauri/src/audio/recording_saver.rs#L152)  
If the recordings directory is unavailable or unwritable, initialization only logs the failure. Capture continues, incoming audio is discarded, and stopping treats the missing saver as disabled auto-save. Propagate initialization failure before capture starts, or explicitly offer transcript-only recording.

**P1 — Stopping can discard queued audio**  
[`recording_saver.rs:360`](https://github.com/Zackriya-Solutions/meetily/blob/a5d81f3a715fc1dcc966706b0719dd9e6b43fbd3/frontend/src-tauri/src/audio/recording_saver.rs#L360)  
When checkpoint encoding is slow, audio can remain queued after the pipeline stops. Clearing `is_saving` makes the consumer exit before draining that queue, so finalization can succeed with the recording tail missing. Retain and await the saver task before finalizing.

**P1 — Final save failures are hidden**  
[`recording_manager.rs:400`](https://github.com/Zackriya-Solutions/meetily/blob/a5d81f3a715fc1dcc966706b0719dd9e6b43fbd3/frontend/src-tauri/src/audio/recording_manager.rs#L400)  
Audio, transcript, or metadata write failures are logged, but `save_recording_only` returns success. Shutdown then reports successful saving without explaining that the export failed. Preserve cleanup while surfacing the save error and available checkpoint recovery.

**P2 — Fatal transcription errors leave capture running**  
[`RecordingControls.tsx:292`](https://github.com/Zackriya-Solutions/meetily/blob/a5d81f3a715fc1dcc966706b0719dd9e6b43fbd3/frontend/src/components/RecordingControls.tsx#L292), [`useRecordingStop.ts:144`](https://github.com/Zackriya-Solutions/meetily/blob/a5d81f3a715fc1dcc966706b0719dd9e6b43fbd3/frontend/src/hooks/useRecordingStop.ts#L144)  
A transcription worker initialization failure invokes post-stop processing without calling the native stop command. Capture continues without transcription. Stop controls return, so manual recovery remains possible. Route fatal errors through native shutdown before post-stop processing.

**Validation:** 17 frontend tests, 7 build-helper checks, and app-source TypeScript checking passed. The official runtime archive and packaged files matched all pinned sizes and checksums. Windows installer execution, affected CPU testing, and a full Rust application build remain unverified.

safvanatzack:
Thanks @athulchandroth for the fresh review. We verified all four findings against `a5d81f3` and the `devtest` history; the attribution is correct: they are pre-existing and were not introduced or amplified by PR #767.

We have captured the follow-up work with full reproduction context and acceptance criteria:

- #781 — recording saver initialization failure, queued-audio drain, and hidden final-save failures
- #782 — fatal transcription worker initialization can enter frontend post-stop processing without first stopping native capture

We’ll keep PR #767 scoped to the ONNX runtime compatibility and recovery changes and address these recording-lifecycle issues separately.

PR #771, opened 2026-09-07
fix(vad): stabilize live segmentation and final timestamps

### LINKED ISSUE #399: [BUG]Imported meeting causing out of sync transcript segment with audio(only the last segment, edge case)

safvanatzack:
## Summary

- Carries forward @Hus-Mek's focused VAD contribution from #679, preserving the original commits and attribution.
- Uses the established 500 ms live redemption policy while import and retranscription remain at 2000 ms.
- Keeps VAD start timestamps session-absolute and preserves the real post-resampling endpoint before padding the final Silero inference frame.
- Strengthens regression coverage so a forced final segment satisfies `start <= end <= audio duration`.
- Clarifies that 500 ms does not guarantee bounded delivery during uninterrupted speech; that remains tracked in #756.

## Context

This draft supersedes #679 after incorporating the remaining correction requested in [the final review comment](https://github.com/Zackriya-Solutions/meetily/pull/679#issuecomment-5492691211). It addresses the final-segment timestamp failure reported in #399 without including transcript cleanup or the bounded-segment work from #756.

Closes #399.

## Verification

- `cargo test --lib audio::vad::tests::test_flush_segment_timestamps_stay_within_audio_duration` — 1 passed
- `cargo test --lib audio::vad` — 7 passed
- `cargo test --lib audio::pipeline` — 1 passed
- `cargo check --lib` — passed with pre-existing warnings only

### DIFF
--- a/frontend/src-tauri/src/audio/import.rs
+++ b/frontend/src-tauri/src/audio/import.rs
@@ -52,9 +52,10 @@ impl Drop for ImportGuard {
 }
 
 /// VAD redemption time in milliseconds - bridges natural pauses in speech
-/// Batch processing needs longer redemption (2000ms) than live pipeline (400ms)
-/// because the entire file is processed at once by VAD, and 400ms fragments
-/// speech at every natural sentence/topic pause (500ms-2s)
+/// Batch processing needs longer redemption (2000ms) than the live pipeline
+/// (500ms) because the entire file is processed at once by VAD with no
+/// latency requirement, and short redemption fragments speech at every
+/// natural sentence/topic pause (500ms-2s)
 const VAD_REDEMPTION_TIME_MS: u32 = 2000;
 
 /// Maximum file size: 20GB (prevents OOM and excessive processing time)
--- a/frontend/src-tauri/src/audio/pipeline.rs
+++ b/frontend/src-tauri/src/audio/pipeline.rs
@@ -13,6 +13,15 @@ use super::recording_state::{AudioChunk, AudioError, RecordingState, DeviceType}
 use super::audio_processing::{audio_to_mono, LoudnessNormalizer, NoiseSuppressionProcessor, HighPassFilter};
 use super::vad::{ContinuousVadProcessor};
 
+/// How long a silence must last before the VAD closes a speech segment, and
+/// therefore how long the audio clips handed to the ASR engine are.
+///
+/// Kept equal to `import.rs::VAD_REDEMPTION_TIME_MS` and
+/// `retranscription.rs::VAD_REDEMPTION_TIME_MS` so that live recording and
+/// offline re-transcription of the same audio segment identically. Raising this
+/// trades live-transcript latency for longer, more accurate ASR requests.
+const VAD_REDEMPTION_TIME_MS: u32 = 2000;
+
 /// Ring buffer for synchronized audio mixing
 /// Accumulates samples from mic and system streams until we have aligned windows
 struct AudioMixerRingBuffer {
@@ -719,16 +728,34 @@ impl AudioPipeline {
         // For now, we log it for monitoring and potential optimization
         let _ = (mic_device_name, mic_device_kind, system_device_name, system_device_kind);
 
-        // Create VAD processor with balanced redemption time for speech accumulation
-        // The VAD processor now handles 48kHz->16kHz resampling internally
-        // This bridges natural pauses without excessive fragmentation
-        // For mac os core audio, 900ms, for windows 400ms seems good
-
-        let redemption_time = if cfg!(target_os = "macos") { 400 } else { 400 };
-
-        let vad_processor = match ContinuousVadProcessor::new(sample_rate, redemption_time) {
+        // Create VAD processor. The VAD processor handles 48kHz->16kHz resampling
+        // internally.
+        //
+        // Redemption time is how long a silence must last before the VAD closes a
+        // speech segment, so it decides how long the audio clips handed to the ASR
+        // engine are. Conversational speech pauses constantly for breath and
+        // mid-sentence thought, and every pause longer than this becomes a segment
+        // boundary and therefore a separate transcription request.
+        //
+        // This was 400ms, which fragmented a 26-minute meeting into 322 requests with
+        // a median length of 3.5s. Whisper is a fixed 30-second-window model: below
+        // that it zero-pads the window and leans on its language-model prior, which
+        // was trained on web subtitles, so short clips come back as memorised
+        // boilerplate ("subscribe to the channel", "thank you") instead of speech.
+        // Measured on a real recording, 47% of segment boundaries sat in the
+        // 0.42-0.75s range that a longer redemption simply bridges.
+        //
+        // 2000ms matches what the offline paths already use (see
+        // `import.rs::VAD_REDEMPTION_TIME_MS` and
+        // `retranscription.rs::VAD_REDEMPTION_TIME_MS`), so live and batch
+        // transcription now segment identically. It costs up to ~1.6s of extra
+        // live-transcript latency at the end of each utterance.
+        let vad_processor = match ContinuousVadProcessor::new(sample_rate, VAD_REDEMPTION_TIME_MS) {
             Ok(processor) => {
-                info!("VAD-driven pipeline: VAD segments will be sent directly to Whisper (no time-based accumulation)");
+                info!(
+                    "VAD-driven pipeline: segments dispatched per speech burst (redemption_time={}ms)",
+                    VAD_REDEMPTION_TIME_MS
+                );
                 processor
             }
             Err(e) => {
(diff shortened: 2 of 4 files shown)


athulchandroth:
Reviewed PR #771 at `cda2ff0`. **No new actionable regressions found. Two P1 findings, both pre-existing:**

1. **Final flush drops the beginning of the last utterance** — [`vad.rs:198`](https://github.com/Zackriya-Solutions/meetily/blob/cda2ff0d457ef5f67d9997e73013e1b41af32106/frontend/src-tauri/src/audio/vad.rs#L198).

   Stopping during speech emits `current_speech`, which only begins accumulating after Silero's delayed `SpeechStart` notification. The emitted timestamp includes the earlier start, but the audio omits the prefix. A runtime probe using the PR's fixture confirmed **570 ms of missing audio**, including pre-speech padding. This can truncate the opening of the final utterance; a short final word can fall below the minimum segment length and disappear.

   Use Silero's full active speech buffer for forced completion, trim synthetic final padding to the real endpoint, and add a regression that checks sample content as well as timestamp bounds.

2. **Failed final transcription reports a successful flush** — [`pipeline.rs:956`](https://github.com/Zackriya-Solutions/meetily/blob/cda2ff0d457ef5f67d9997e73013e1b41af32106/frontend/src-tauri/src/audio/pipeline.rs#L956).

   If the transcription consumer has exited, sending the final chunk fails after VAD has already drained the speech segment. The pipeline logs a warning and returns success, hiding an incomplete transcript. VAD errors during flush are similarly suppressed. The recording manager also logs and discards propagated flush errors.

   Preserve cleanup while surfacing the failed transcription outcome through the recording workflow. Retain the failed segment or mark the transcript incomplete and recoverable from saved audio.

**Verification:** All 7 tests in the unchanged VAD module passed through an isolated harness. An instrumented copy of the fixture confirmed the missing prefix. Full-application testing was blocked by the missing `binaries/llama-helper-aarch64-apple-darwin` sidecar. The error-suppression finding was verified through source inspection, without fault injection.

The real-endpoint capture before padding is correct, and no new dependency or layering problems were identified. The existing uninterrupted-speech limitation remains tracked separately in #756.

safvanatzack:
Thanks @athulchandroth for the review. We independently rechecked the reported findings, their runtime reachability, and their commit-level origin.

For attribution, the original #679 contribution consists of `4d626bb` and `0ceee52`. Our follow-up is `cda2ff0`.

| Finding | Commit-level origin | Verification | Proposed disposition |
|---|---|---|---|
| **Forced final segment loses the beginning of the utterance** | The defective `current_speech` accumulation and forced clone predate #679 and trace to `1e81fdb3`. Neither #679 nor `cda2ff0` introduced it. | Confirmed independently with the existing 23-second fixture. The segment declared 39,200 real interval samples but carried 30,240 samples, including 160 synthetic padding samples. This means 570 ms of real prefix audio was missing, and the live sample-derived endpoint was 560 ms early. | **Fix in this PR.** Although pre-existing, it is inside the exact `flush()` path that #679 and our follow-up modify and certify. Use Silero's backfilled active-speech buffer and trim it to the real interval. |
| **Failed final VAD or transcription send still reports a successful flush** | Warning-only flush and sender handling also predate #679 and trace to the older pipeline implementation. Neither #679 nor `cda2ff0` changed these branches. | The mechanism is confirmed by source inspection. However, normal model-initialization and provider failures already emit `transcription-error` and reach the frontend. Silent loss requires a VAD runtime failure or unexpected consumer exit after initialization, and we do not yet have evidence that this is a common production path. | **Track separately.** This is a valid reliability issue, but it is independent of #679's VAD policy and our endpoint correction. It should not expand this focused PR. |
| **Unbounded active VAD segment under continuous speech** | The unbounded accumulator predates #679. Commit `0ceee52` changes live redemption from 400 ms to 500 ms, so any proven incremental reachability belongs to the original #679 policy change. | The underlying continuous-speech limitation is real and remains #756. A local 400-versus-500 probe using 390, 420, 450, 480, 510, and 600 ms acoustic gaps did not reproduce a behavioral difference, so the specific added user risk from 500 ms remains unproven. | **Keep 500 ms and keep #756 separate.** Do not add an ad hoc maximum segment here. Add representative real-audio replay cases to #756. |
| **Final mixer or non-48 kHz resampler tails are not drained** | Both shutdown-buffer behaviors predate #679 and are upstream of every changed VAD line. | Mixer behavior was reproduced: 28,799 samples per source remained buffered, `can_mix()` was false, and terminal handling had no mixer drain. | **Separate audio-shutdown work.** Correct order is capture-resampler drain, mixer drain, then VAD flush. |
| **All-empty retranscription can replace the existing transcript with nothing** | Predates #679. The PR changes only retranscription policy comments. | Source path is reachable: empty ASR results produce an empty replacement, `DELETE` commits with zero inserts, and completion is emitted. | **Separate urgent data-integrity issue.** Guard before the transaction and preserve existing rows. |
| **Import outcome, sidecar persistence, and related test gaps** | These paths and tests predate #679, except for #679's new direct VAD tests and policy comments. | Valid lower-priority UX, persistence, or assurance concerns, but none is caused by the current runtime diff. | **Do not expand this PR.** Track separately or address when the relevant subsystem is changed. |

safvanatzack:
Deferred findings from the verification above are now tracked separately:

- #775: surface final VAD and transcription-send failures
- #773: drain capture-resampler and mixer tails before final VAD flush
- #778: preserve the existing transcript when retranscription produces no usable ASR text
- #774: explain imports where speech is detected but ASR produces no usable text
- #777: report and repair partial transcript sidecar saves
- #776: make VAD chunk equivalence exercise the actual large-file path
- #756: updated with the 400-versus-500 probe result, forced-boundary payload constraint, and additional replay cases

Each issue links back to the reviewer comment and/or the independent reachability and commit-origin analysis in this thread. PR #771 remains scoped to the forced-final-payload correction and its consumer-visible regression.

athulchandroth:
Follow-up review of **`bc00314db1dc8ae6e638ca0674694b2c48b22b63`**: **no new actionable regressions found.**

**Resolved: final-utterance prefix loss.** The forced flush now uses Silero's complete active speech buffer and trims terminal padding. The strengthened regression verifies that the emitted samples exactly match the interval described by the timestamps, and that a second flush does not emit duplicate audio.

**P1 [PRE-EXISTING] remains: final-flush failures are reported as success** — [`pipeline.rs:956`](https://github.com/Zackriya-Solutions/meetily/blob/bc00314db1dc8ae6e638ca0674694b2c48b22b63/frontend/src-tauri/src/audio/pipeline.rs#L956).

If the transcription consumer has exited, sending the final chunk fails after VAD has already drained the segment. The pipeline logs a warning and returns success, hiding an incomplete transcript. VAD errors during flush are similarly suppressed at line 967. The recording-manager caller also discards propagated flush errors, so changing only the pipeline return value would not fully resolve this.

Preserve cleanup while surfacing the failed transcription outcome through the recording workflow. Retain the failed chunk or mark the transcript incomplete and recoverable from the saved recording. This handling is unchanged by the PR; no valid-input failure of the new interval guards was established.

**Verification:**

- All **7 tests in the current VAD module passed** in a separately compiled standalone harness.
- An additional probe passed **four 16 kHz scenarios** after a normally completed segment: 500/2000 ms silence thresholds, each with and without terminal padding. It verified exact audio content, endpoints, and empty repeated flushes.
- The exploratory 48 kHz fixture did not reach the required active-speech state, so that scenario remains unverified.
- Full-application testing was not rerun for this head. The previous attempt was blocked by the missing `llama-helper` sidecar.
- Error propagation was inspected statically; no runtime fault injection was performed.

No new dependency or layering problems were identified. Uninterrupted-speech bounds remain tracked separately in #756.

PR #779, opened 2026-09-07
Fix microphone checks, summary progress, and Start button layout [skip ci]

athulchandroth:
## Description

Fix three recording and summary issues:

- Add a microphone callback check to both macOS recording-start paths. A failed probe prevents recording from starting and returns an error to the existing UI.
- Restore summary generation status and reconnect polling when returning to a meeting. Preserve regeneration and cancellation behavior, prevent stale responses from affecting another meeting or a newly resumed poll, and avoid replaying historical completion notifications.
- Remove the homepage entrance animation whose transform changed the fixed recording controls' positioning reference and caused the Start button to jump.

## Related Issue

No linked issue.

## Type of Change

- [x] Bug fix

## Testing

- [x] All tests pass

`bun test tests`: 29 passed, 0 failed, including 12 React lifecycle regression tests with mocked IPC. Application-source TypeScript checking and scoped diff checks pass.

`cargo check --locked --offline` is blocked by a missing generated Tauri permissions file (`app_hide.toml`) referenced through a stale build-cache path from an old checkout. Rust compilation and native desktop behavior are not verified. Full TypeScript checking also encounters missing `bun:test` declarations in the existing test setup; the passing check covers application sources.

### Manual test cases

| Case | Steps | Expected result |
| --- | --- | --- |
| Fresh microphone permission (macOS) | Quit Meetily. Run `tccutil reset Microphone com.meetily.ai` to remove its microphone permission entry completely; disabling the toggle is insufficient. Reopen the app, start recording, and allow access when prompted. | macOS requests microphone access, and recording starts with microphone audio after access is allowed. |
| Summary progress after navigation | Generate a summary, navigate away, and return while generation is active; wait for completion. Repeat when regenerating an existing summary. | Loading remains visible after returning, the summary updates on completion, and existing content is preserved during regeneration. |
| Stop summary after navigation | Start summary generation, navigate away and back, then select Stop. | The active generation is cancelled and the loading state clears. |
| Home recording controls | Repeatedly navigate to Home with the sidebar expanded and collapsed, then start and stop a recording. | The Start button does not jump when Home opens, and recording controls remain usable. |

## Documentation

- [x] No documentation needed

## Checklist

- [x] Self-reviewed the code
- [x] No merge conflicts

## Additional Notes

The microphone check intentionally uses the default input device and accepts the first callback without inspecting sample content. System-audio silence detection is outside this PR. The summary navigation issue and homepage animation issue predate the microphone change.

The test renderer and its types are development-only dependencies; existing dependency resolutions are unchanged. No screenshots were captured.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -103,9 +103,11 @@
         "@types/node": "^20",
         "@types/react": "^18.3.18",
         "@types/react-dom": "^18.3.5",
+        "@types/react-test-renderer": "18.3.1",
         "autoprefixer": "^10.0.1",
         "concurrently": "^8.2.2",
         "postcss": "^8",
+        "react-test-renderer": "18.3.1",
         "tailwindcss": "^3.4.1",
         "typescript": "^5.7.2",
         "wait-on": "^7.2.0"
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -198,6 +198,9 @@ importers:
       '@types/react-dom':
         specifier: ^18.3.5
         version: 18.3.7(@types/react@18.3.28)
+      '@types/react-test-renderer':
+        specifier: 18.3.1
+        version: 18.3.1
       autoprefixer:
         specifier: ^10.0.1
         version: 10.5.0(postcss@8.5.14)
@@ -207,6 +210,9 @@ importers:
       postcss:
         specifier: ^8
         version: 8.5.14
+      react-test-renderer:
+        specifier: 18.3.1
+        version: 18.3.1(react@18.3.1)
       tailwindcss:
         specifier: ^3.4.1
         version: 3.4.19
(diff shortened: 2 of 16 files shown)


safvanatzack:
Thanks @athulchandroth for putting this together. I did a fresh review after updating the comparison base to the current `devtest` tip at `82d9cb3`.

The summary restoration and request-ownership work looks solid, the Home animation removal is narrowly scoped, and the updated VAD changes from `devtest` remain intact in the projected merge.

I found two items to address:

1. **Bluetooth wake ordering**
   - Both new macOS microphone probes run before `RecordingManager::start_recording` at `recording_commands.rs:290-301` and `477-486`.
   - The probe fails after five seconds without a callback, but the existing manager wake at `recording_manager.rs:275-300` is specifically intended to restore callbacks from a cold Bluetooth route that may otherwise remain silent for 60 to 90 seconds.
   - This can report a false permission failure before the wake has a chance to run, affecting both recording-start routes.
   - Please move the probe after the existing non-fatal output wake and before `start_streams`, while preserving selected-output resolution and the no-microphone system-audio-only fallback.

2. **Sidebar permission guidance**
   - The native preflight returns actionable System Settings guidance, but both sidebar catches at `useRecordingStart.ts:261-269` and `356-363` replace it with `Failed to start recording. Check console for details.`
   - The stored status message is not rendered, so users starting from the sidebar do not receive the recovery instructions shown by the manual control.
   - Please surface `errorMsg` in both non-race sidebar failures or reuse the existing device-error presentation. The `already in progress` handling should remain unchanged.

I also adversarially checked the proposed F32 callback concern and did not retain it: CoreAudio accepts the requested F32 client stream and converts the physical PCM format.

Could you address the Bluetooth ordering issue before merge and include the sidebar guidance fix in this PR? Thanks again for the careful lifecycle work.

Issue #775, opened 2026-09-07
fix(audio): surface final VAD and transcription send failures


safvanatzack:
## Problem

The live shutdown path can lose the final transcript segment while still reporting a successful pipeline flush.

`AudioPipeline::flush_remaining_audio` logs and suppresses both:

- `ContinuousVadProcessor::flush()` failures;
- failure to send the final `AudioChunk` after the transcription consumer has exited.

`AudioPipelineManager::stop` and `RecordingManager::stop_streams_and_force_flush` also convert or log terminal pipeline failures without returning a transcription-incomplete outcome.

Normal model-initialization and provider failures already emit `transcription-error`, so this issue is specifically about VAD runtime failures and unexpected consumer exits after initialization.

## User impact

The WAV recording may complete successfully while the live transcript is incomplete. The user is not told that final transcription degraded and receives no direct recovery instruction.

## Reachability and evidence

- Final VAD/send handling: `frontend/src-tauri/src/audio/pipeline.rs:932-970`
- Pipeline task result suppression: `frontend/src-tauri/src/audio/pipeline.rs:1042-1057`
- Recording-manager suppression: `frontend/src-tauri/src/audio/recording_manager.rs:355-374`
- Existing transcription error UI: `frontend/src/components/RecordingControls.tsx:275-309`

The mechanism was confirmed by source inspection. Production frequency has not been established through fault injection, so this should preserve recording rather than treating every transcription degradation as fatal.

## Acceptance criteria

1. A final VAD failure or failed final transcription send produces a caller-visible `transcription incomplete` or `transcription degraded` outcome.
2. WAV recording cleanup and persistence still complete when recoverable.
3. The meeting records enough state to tell the user that retranscription from saved audio is available.
4. A closed consumer does not cause repeated futile sends or duplicate user notifications.
5. Fault-injection coverage exercises both final VAD failure and final channel-send failure.
6. Existing model-initialization and provider error behavior remains unchanged.

## References

- Reviewer finding: https://github.com/Zackriya-Solutions/meetily/pull/771#issuecomment-5568522226
- Independent reachability and attribution follow-up: https://github.com/Zackriya-Solutions/meetily/pull/771#issuecomment-5569666733
- Related PR: https://github.com/Zackriya-Solutions/meetily/pull/771

PR #780, opened 2026-09-07
fix(vad): surface final VAD flush and transcription-send failures (#775)

Hus-Mek:
## Summary

Fixes the reliability gap identified in the #679 VAD follow-up review: the forced final flush silently reports success when the VAD processor errors, or when the transcription consumer has exited and the final segment send fails. This hides an incomplete transcript and marks a truncated meeting as clean. Relates to #775.

## Change

In `flush_remaining_audio`:

- A failed final-segment `transcription_sender.send(...)` now propagates as an error instead of logging a warning and continuing.
- A failed VAD `flush()` now returns `Err` instead of logging and returning `Ok`.

Both failures now reach the recording workflow so it can flag the meeting as incomplete and recoverable from the saved audio.

## Test evidence

`cargo check --lib` passes on the branch (no warnings from this change). Full application runtime was not exercised; the two error paths are verified by source inspection as the review thread described.

## Notes

The forced-final-segment truncation (the timestamp/prefix part of the original review) is being handled separately in #771/#773; this PR is scoped to surfacing flush and send failures only (#775).

### DIFF
--- a/frontend/src-tauri/src/audio/pipeline.rs
+++ b/frontend/src-tauri/src/audio/pipeline.rs
@@ -919,11 +919,11 @@ impl AudioPipeline {
                             device_type: DeviceType::Microphone,
                         };
 
-                        if let Err(e) = self.transcription_sender.send(transcription_chunk) {
-                            warn!("Failed to send final VAD segment: {}", e);
-                        } else {
-                            self.chunk_id_counter += 1;
-                        }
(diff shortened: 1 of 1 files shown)


safvanatzack:
Thank you for tackling the final VAD flush and transcription-send failure paths. This is useful work, and we’ll take a closer look at the approach and share any questions or feedback during the review.

PR #783, opened 2026-09-08
test(vad): exercise the large-file path in chunk equivalence

### LINKED ISSUE #776: test(vad): make chunk equivalence exercise the large-file path
## Problem

`test_vad_chunked_vs_single_processing` does not currently exercise the chunked production branch it names.

The implementation chunks only when `total_samples > 960_000`, while the test supplies exactly 60 seconds at 16 kHz, exactly 960,000 samples. Both calls therefore take the small-file branch. The test also compares only segment counts, not timestamp equivalence.

No current runtime timestamp drift was found. This is regression-assurance debt.

## Evidence

- Large-file threshold and chunk loop: `frontend/src-tauri/src/audio/vad.rs:348-421`
- Exact-boundary equivalence test: `frontend/src-tauri/src/audio/vad.rs:462-485`
- Existing 120-second progress test reaches the branch but does not compare whole-input timestamps: `frontend/src-tauri/src/audio/vad.rs:487-529`

The threshold and named test predate PR #679. PR #679 added adjacent timestamp tests but did not modify this branch.

## Acceptance criteria

1. The equivalence fixture contains strictly more than 960,000 samples.
2. One processor receives the whole input in one call.
3. A second processor receives the identical input in 160,000-sample calls.
4. Each processor is flushed exactly once.
5. Assertions compare segment order, start timestamps, end timestamps, and duration bounds, not only segment count.
6. The test fails if session timestamp state resets or drifts across `process_audio` calls.

## References

- Consolidated verification and attribution: https://github.com/Zackriya-Solutions/meetily/pull/771#issuecomment-5569666733
- Related PR: https://github.com/Zackriya-Solutions/meetily/pull/771

7487:
## Description

`test_vad_chunked_vs_single_processing` never ran the chunked branch it is named
after. `generate_test_audio_with_speech(60.0, 16000)` produces exactly 960,000
samples, and `get_speech_chunks_with_progress` only chunks when
`total_samples > LARGE_FILE_THRESHOLD` (960,000), so both calls took the
small-file branch. The only assertion compared segment counts with a `diff <= 1`
tolerance, so a timestamp reset or drift across `process_audio` calls would not
have failed it either.

The test now drives `ContinuousVadProcessor` directly, which is the only way to
compare a real single-shot run against a real chunked run: with a fixture above
the threshold both public helpers take the chunked path, since
`get_speech_chunks` delegates to `get_speech_chunks_with_progress`.

- Fixture is 29 seconds of silence followed by 40 seconds of speech-like audio
  (1,104,000 samples), asserted to exceed `LARGE_FILE_THRESHOLD`. The leading
  silence puts the detected segment past the first chunk and across a chunk
  boundary — it lands at 29580–32950 ms, spanning the 30 s boundary.
- One processor receives the whole input in a single `process_audio` call; a
  second receives the identical input in `CHUNK_SIZE` (160,000-sample) calls.
  Each is flushed exactly once, matching the two production branches.
- Assertions compare segment count, per-segment start and end timestamps, and
  sample counts, then check ordering and duration bounds on the chunked side.
- A final assertion requires the first segment to start beyond one chunk's worth
  of audio, so a session clock that restarted per call fails instead of quietly
  producing plausible-looking segments.

`LARGE_FILE_THRESHOLD` and `CHUNK_SIZE` moved from function-local constants to
module scope so the test asserts against the same values production uses instead
of re-hardcoding them — that duplication is what let the fixture sit exactly on
the boundary unnoticed. Values are unchanged and no production behaviour changes.

Checked that the new assertions are load-bearing by mutating the test locally:
recreating the processor per chunk drops the segment entirely (count assertion
fires), and rebasing each call's timestamps to a per-call origin reports the
segment at -420 ms instead of 29580 ms (timestamp assertion fires).

## Related Issue

Fixes #776

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [x] Other: test coverage

## Testing
- [x] All tests pass

`cargo test --lib` on macOS aarch64: 244 passed, 0 failed, 3 ignored.

## Documentation
- [ ] Documentation updated
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] No merge conflicts

### DIFF
--- a/frontend/src-tauri/src/audio/vad.rs
+++ b/frontend/src-tauri/src/audio/vad.rs
@@ -8,6 +8,13 @@ use std::time::Duration;
 /// sample count and timestamp inside this module is expressed in it.
 const VAD_SAMPLE_RATE: u32 = 16000;
 
+/// Inputs longer than one minute at 16kHz take the chunked path in
+/// `get_speech_chunks_with_progress` so progress can be reported and cancelled.
+const LARGE_FILE_THRESHOLD: usize = 960_000;
+
+/// Size of a single chunk on the large-file path: 10 seconds at 16kHz.
+const CHUNK_SIZE: usize = 160_000;
+
 /// Represents a complete speech segment detected by VAD
 #[derive(Debug, Clone)]
 pub struct SpeechSegment {
@@ -380,10 +387,6 @@ where
 
     let total_samples = samples_mono_16k.len();
 
-    // For large files (>1 minute at 16kHz = 960,000 samples), process in chunks with progress logging
-    const LARGE_FILE_THRESHOLD: usize = 960_000;
-    const CHUNK_SIZE: usize = 160_000; // 10 seconds at 16kHz
-
     let mut all_segments = Vec::new();
 
     if total_samples > LARGE_FILE_THRESHOLD {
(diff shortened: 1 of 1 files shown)


safvanatzack:
Thanks for picking up [issue 776](https://github.com/Zackriya-Solutions/meetily/issues/776). Another contributor has since opened [PR 785](https://github.com/Zackriya-Solutions/meetily/pull/785) for the same issue. We’ll consider both implementations together and follow up with the next steps.

PR #786, opened 2026-09-09
fix(retranscription): preserve transcript on empty ASR result

### LINKED ISSUE #778: fix(retranscription): preserve existing transcript when ASR returns no usable text
## Problem

Retranscription validates that VAD returned at least one segment, but it does not validate that ASR produced at least one usable transcript before destructive replacement.

If every processable segment returns empty or whitespace text, `all_transcripts` remains empty. The code then deletes all existing transcript rows, performs zero replacement inserts, commits the transaction, writes an empty sidecar, and emits successful completion.

## User impact

A model miss can remove a meeting's existing transcript while the UI reports `Retranscription complete! 0 segments created.` The source audio remains available, but recovery requires rerunning transcription and the previous transcript is no longer preserved.

## Reachability and evidence

Whisper can legitimately return an empty successful result after output cleanup, and Parakeet can return an empty model result. Individual empty segments are therefore expected, but an all-empty replacement must not be treated as success.

- Empty-result handling: `frontend/src-tauri/src/audio/retranscription.rs:365-413`
- Empty segment conversion: `frontend/src-tauri/src/audio/retranscription.rs:420-423`
- Delete and zero-insert commit: `frontend/src-tauri/src/audio/retranscription.rs:430-461`
- Successful completion: `frontend/src-tauri/src/audio/retranscription.rs:492-499`
- Completion UI: `frontend/src/components/MeetingDetails/RetranscribeDialog.tsx:145-160`

This behavior predates PR #679 and is unrelated to PR #771's timestamp correction.

## Acceptance criteria

1. After ASR processing and before opening the replacement transaction, an entirely empty usable result returns a caller-visible `No transcribable speech` error.
2. Existing transcript rows and `transcripts.json` remain untouched in that case.
3. Individual empty or false-positive VAD segments remain skippable when other usable segments exist.
4. The frontend receives `retranscription-error`, not `retranscription-complete`.
5. Intentional transcript clearing, if supported, remains a separate explicit operation.
6. Regression coverage begins with existing rows, supplies nonzero VAD output plus all-empty ASR results, and proves the rows survive.

## References

- Consolidated verification and attribution: https://github.com/Zackriya-Solutions/meetily/pull/771#issuecomment-5569666733
- Related PR: https://github.com/Zackriya-Solutions/meetily/pull/771

dajiaohuang:
## Summary

Retranscription currently deletes the existing SQLite transcript before inserting the new result. When VAD finds speech but every ASR call returns empty text, the operation commits zero replacement rows, writes an empty sidecar, and reports successful completion.

This change rejects an all-empty retranscription result before the destructive transaction, preserving the existing transcript and returning a stable, actionable error. Individual empty segments remain skippable when another segment produces usable text.

## Related issue

Fixes #778

## Validation

- `cargo test -p meetily --lib audio::retranscription` — 17 passed.
- `cargo test -p meetily --lib audio::retranscription::tests::test_empty_retranscription_result_is_rejected_before_replacement -- --exact` — 1 passed.
- `git diff --check` — passed.

The focused tests used the repository's temporary ignored native helper placeholder because the sidecar is not checked in; it was removed after validation.

### DIFF
--- a/frontend/src-tauri/src/audio/retranscription.rs
+++ b/frontend/src-tauri/src/audio/retranscription.rs
@@ -417,6 +417,11 @@ async fn run_retranscription<R: Runtime>(
         return Err(anyhow!("Retranscription cancelled"));
     }
 
+    // Never replace an existing transcript with an empty result. Individual
+    // empty segments are valid and remain skippable, but an all-empty result
+    // must stop before the destructive delete+insert transaction below.
+    ensure_usable_retranscription_result(&all_transcripts)?;
+
     emit_progress(&app, &meeting_id, "saving", 80, "Saving transcripts...");
 
     // Create transcript segments with proper timestamps from VAD
@@ -497,6 +502,18 @@ async fn run_retranscription<R: Runtime>(
         duration_seconds,
         language,
     })
+}
+
+fn ensure_usable_retranscription_result(
+    transcripts: &[(String, f64, f64)],
+) -> Result<()> {
+    if transcripts.is_empty() {
+        return Err(anyhow!(
+            "No transcribable speech was produced; existing transcript was preserved"
+        ));
+    }
+
+    Ok(())
 }
 
 /// Emit progress event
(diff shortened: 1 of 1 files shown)


safvanatzack:
Thanks for raising this fix. We’ll look into it.

PR #792, opened 2026-09-09
fix(vad): bound live segments during continuous speech

### LINKED ISSUE #756: fix(vad): bound live segments to guarantee transcript updates

dajiaohuang:
Closes #756.

Bound uninterrupted live VAD delivery with a 20 second target and 25 second hard ceiling. Forced boundaries take a contiguous prefix from the canonical VAD session using absolute timestamps and advance the session buffer, preserving ordering and avoiding overlap or gaps; the existing 500ms live redemption policy remains unchanged.

Validation: full local Rust library tests pass with Visual Studio libclang and the branch's bundled ONNX Runtime (247 passed, 2 ignored), including deterministic bounded and contiguous live-segment coverage. A real microphone/system-audio replay remains unavailable; hosted checks are currently absent. The touched-file rustfmt parse reaches existing repository formatting drift without syntax errors.

### DIFF
--- a/frontend/src-tauri/src/audio/pipeline.rs
+++ b/frontend/src-tauri/src/audio/pipeline.rs
@@ -26,6 +26,11 @@ use super::vad::{ContinuousVadProcessor};
 /// during continuous speech is tracked separately in #756.
 const VAD_REDEMPTION_TIME_MS: u32 = 500;
 
+/// Live transcription is delivered incrementally during uninterrupted speech.
+/// The target leaves room for a frame to arrive before the hard ceiling.
+const LIVE_SEGMENT_TARGET_MS: u32 = 20_000;
+const LIVE_SEGMENT_HARD_MAX_MS: u32 = 25_000;
+
 /// Ring buffer for synchronized audio mixing
 /// Accumulates samples from mic and system streams until we have aligned windows
 struct AudioMixerRingBuffer {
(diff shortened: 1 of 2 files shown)


safvanatzack:
Thanks for tackling this. We’ll take a look.

PR #806, opened 2026-09-16
fix(linux): capture real system audio via PulseAudio/PipeWire monitor sources

hgschmidtmw:
## Description
On Linux, "System Audio" recording always silently fell back to the ALSA `default` capture device — which is the microphone-like default input, not the meeting/call playback. Recordings never actually contained system audio.

Root cause: cpal has no PulseAudio host on Linux (ALSA only), so the existing monitor-source lookup — scanning cpal's ALSA host for device names containing `"monitor"` — could never find anything. PipeWire/PulseAudio virtual monitor sources are never exposed as plain ALSA PCMs on a PipeWire-only system (confirmed via `aplay -L`/`arecord -L`). The System Audio picker was also showing raw ALSA hardware playback devices (HDMI outs, surround-mode variants, etc.) that can never function as capture sources on this backend.

Fixes:
- `audio/devices/platform/linux.rs`: enumerate real PulseAudio/PipeWire monitor sources via `pactl list short sources` instead of the broken cpal-based scan.
- `audio/devices/discovery.rs`: stop listing dead-end ALSA hw playback devices as "System Audio" options on Linux.
- `audio/capture/pulse_monitor.rs` (new) + `audio/stream.rs`: capture a selected `*.monitor` source directly via `parec` (the PulseAudio/PipeWire-compatible recorder), bypassing cpal entirely since it cannot open a virtual PulseAudio source by name. Requested at 48kHz/stereo float32 to match the pipeline's expected format, avoiding extra resampling.

Bonus audio-quality fix: the old fallback path captured the "default" ALSA device at 44.1kHz and ran it through the pipeline's sinc resampler to 48kHz on every chunk, which produced audible periodic clicking/ticking in recordings. Since `parec` is asked for native 48kHz/stereo directly, system audio now skips resampling entirely — the clicking is gone.

Also includes two small `build-gpu.sh` fixes hit while building/testing this on Fedora + AMD ROCm:
- `llama-helper` has no `hipblas` Cargo feature (only `metal`/`cuda`/`vulkan`) — map `hipblas → vulkan` for that sidecar, same as the existing `coreml → metal` mapping.
- Added a `cpu`/`none` override value to explicitly skip GPU feature autodetection (useful when a GPU backend is broken/unavailable, as ROCm 7.1's hipBLAS is with the bundled whisper.cpp right now).

## Related Issue
None filed yet — found while trying to record a Zoom meeting on a Fedora/PipeWire workstation.

## Type of Change
- [x] Bug fix

## Testing
- [x] All tests pass

Manually verified end-to-end: built and ran the app on Fedora 44 + PipeWire, selected the real monitor source for "System Audio", recorded a live Zoom meeting, and got a full transcript of the meeting audio, with the periodic clicking/ticking artifacts from the old resampled path gone. Selecting "Default System Audio" correctly behaves as a plain default input (no longer silently misroutes).

## Documentation
- [x] No documentation needed

## Checklist
- [x] Self-reviewed the code
- [x] Branch is up to date with main
- [x] No merge conflicts

## Additional Notes
Linux-only change (`#[cfg(target_os = "linux")]` throughout); no behavior change on macOS/Windows.


hgschmidtmw:
Closing for now — still reviewing the change myself. Will reopen (or open a fresh PR from the same branch) once review is done.

